Issue #21, opened 2014-10-13
Bug: Task-Related Operator Overloads Introduce Concurrency


mattpodwysocki:
Copied from https://rx.codeplex.com/workitem/48

Hi,

I just ran into a problem related to the unexpected introduction of concurrency when using Task-based asynchrony in conjunction with Rx in a Windows 8 app.  I've had this problem before and thought that I was doing something wrong or that it's the correct behavior of some kind of new multi-threaded UI model in WinRT, yet now I realize that it seems to be a bug in Rx.

If you search the entire Rx codebase for ".ContinueWith(" you'll see that every usage doesn't specify the [ExecuteSynchronously](http://msdn.microsoft.com/en-us/library/system.threading.tasks.taskcontinuationoptions.aspx) flag, thus allowing the TPL to schedule the continuation on the thread pool, and it always does.

According to the [documentation](http://msdn.microsoft.com/en-us/library/dd270696.aspx): 

> Creates a continuation that executes asynchronously when the target Task<TResult> completes.

But we don't want asynchronous execution, we want synchronous execution.  I'd imagine this is true for all uses of __ContinueWith__ throughout Rx, though I haven't looked at the finer details yet.

Executing __ContinueWith__ asynchronously is a problem in general because the concurrency being introduced isn't related to the specified IScheduler, if one is supplied by the user, and it's entirely unexpected when using particular operators that should never introduce concurrency by default, as shown below.  As a result, the threading context of notifications is incorrect and we have little control over it.

There are two situations in which I've run into this problem in my code: 
 1. The __ToObservable__ extension for __Task<T>__ always introduces concurrency.  For example, the following code illustrates how the UI threading context is automatically captured by _await_, since I'm not specifying `ConfigureAwait(false)`, yet it's lost with __ToObservable__.  My expectation has always been that the observable returned by __ToObservable__ simply generates a notification within the completion context of the converted __Task__, without introducing concurrency unnecessarily.  (NOTE: I'm not actually sure whether forcing Rx's usage of __ContinueWith__ to execute synchronously would solve this problem, though I expect that it will.)
```c#
private async void _Loaded(object sender, RoutedEventArgs e)
{
    Contract.Assert(Dispatcher.HasThreadAccess);  // true
    await DoSomethingAsync();
    Contract.Assert(Dispatcher.HasThreadAccess);  // true
    DoSomethingAsync().ToObservable().Subscribe(_ =>
    {
[...]

PR #65, opened 2014-11-16
Overloaded both Task.ToObservable methods with IScheduler parameters.

RxDave:
Partial fix for #21 

I've also included an Rx.vssettings file in my commits. I actually intended on making a separate pull request for this, but it seems that GitHub won't let me choose the commits that I want to include/exclude from my pull request.


bartdesmet:
Thanks! The change / addition makes sense. We'll review it for the next release of Rx.

bartdesmet:
We've added an overload for `ToObservable` which takes in an `IScheduler`.

Issue #73, opened 2014-12-21
Consistent CompositeDisposable construction semantics


LeeCampbell:
The `CompositeDisposable` type does not allow for null values to be added to it via the `Add` method, however this check is not performed on the constructor that takes a params array or an `IEnumerable` of `IDisposable` values.

i.e. if these tests are added to the Rx.NET test suite, it would fail.

        [TestMethod, ExpectedException(typeof(ArgumentNullException))]
        public void CompositeDisposable_AddNull_via_params_ctor()
        {
            IDisposable d1 = null;
            new CompositeDisposable(d1);
        }
        [TestMethod, ExpectedException(typeof(ArgumentNullException))]
        public void CompositeDisposable_AddNull_via_IEnum_ctor()
        {
            IEnumerable<IDisposable> values = new IDisposable[] {null};
            new CompositeDisposable(values);
        }

This then has the effect that this code would run

            IDisposable d1 = null;
            new CompositeDisposable(d1);

however this code would throw

            IDisposable d1 = null;
            new CompositeDisposable{d1};

which seems unintuitive.

My suggestion is that `null` values not be allowed at all into the `CompositeDisposable`. It should be noted, that even with this suggested null checking in place at `ctor` and on the `Add` method, null checking will still need to be in-place in the `CompostieDisposable.Remove`, `CompostieDisposable.Dispose` & `CompostieDisposable.Clear` methods due to the manual re-sizing algorithm of the internal list.

bartdesmet:
Merged your pull request. Thanks!

PR #74, opened 2014-12-23
Ensuring consistent CompositeDisposable invariant

LeeCampbell:
This pull request enforces that null values can not be added to the CompositeDisposable type via the constructors. Closes issue #73


bartdesmet:
Cool. We should go through other collections as well and add a "ContractUtils" style helper for checking these (cf. the approach in System.Linq.Expressions). We'll get this into a future release.

PR #84, opened 2015-02-24
Bug in async Observable.Create fixes #83

### LINKED ISSUE #83: IObservable from Observable.Create unable to complete
It appears that the following methods [IObservable<TResult> Create<TResult>(Func<IObserver<TResult>, CancellationToken, Task<IDisposable>> subscribeAsync)](https://github.com/Reactive-Extensions/Rx.NET/blob/fcf00a1d0c95672934609b6a2dbb4b93fca55428/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs#L62) and [IObservable<TResult> Create<TResult>(Func<IObserver<TResult>, CancellationToken, Task<Action>> subscribeAsync)](https://github.com/Reactive-Extensions/Rx.NET/blob/fcf00a1d0c95672934609b6a2dbb4b93fca55428/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs#L87) do not correctly mark the resulting `IObservable<TResult>` as complete when the asynchronous delegate completes.

This can be seen by the use of `Stubs.Nop` for the completion delegate [here](https://github.com/Reactive-Extensions/Rx.NET/blob/fcf00a1d0c95672934609b6a2dbb4b93fca55428/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs#L70):

```
var taskCompletionObserver = new AnonymousObserver<IDisposable>(d => subscription.Disposable = d ?? Disposable.Empty, observer.OnError, Stubs.Nop);
```

and [here](https://github.com/Reactive-Extensions/Rx.NET/blob/fcf00a1d0c95672934609b6a2dbb4b93fca55428/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs#L95):

```
var taskCompletionObserver = new AnonymousObserver<Action>(a => subscription.Disposable = a != null ? Disposable.Create(a) : Disposable.Empty, observer.OnError, Stubs.Nop);
```

I believe both of these `Stubs.Nop` calls should be replaced with `observer.OnComplete`.

james-world:
This PR fixes a problem where the completion of an asynchronous subscription that does not return a disposable or action incorrectly triggers a call to `OnCompleted` on the observer. See the following repro:

```C#
var normalObs = Observable.Create<int>(async o => {
    await Task.FromResult(0);
    o.OnNext(1);
    o.OnCompleted();
});
        
normalObs.Subscribe(Console.WriteLine, () => Console.WriteLine("Done"));
```
This outputs an erroneous "Done" due to this problem:

```
Done
1
Done
```

Thanks to @Lukazoid for spotting the inconsistency. See [issue 83](https://github.com/Reactive-Extensions/Rx.NET/issues/83) for more.

### DIFF
--- a/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs
+++ b/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/QueryLanguage.Creation.cs
@@ -47,7 +47,7 @@ namespace System.Reactive.Linq
                 var cancellable = new CancellationDisposable();
 
                 var taskObservable = subscribeAsync(observer, cancellable.Token).ToObservable();
-                var taskCompletionObserver = new AnonymousObserver<Unit>(Stubs<Unit>.Ignore, observer.OnError, observer.OnCompleted);
+                var taskCompletionObserver = new AnonymousObserver<Unit>(Stubs<Unit>.Ignore, observer.OnError, Stubs.Nop);
                 var subscription = taskObservable.Subscribe(taskCompletionObserver);
 
                 return new CompositeDisposable(cancellable, subscription);


james-world:
Oops, didn't mean to close it.

james-world:
Although I'm confident this fixes incorrect behaviour, this is one of those fixes that should probably be considered a breaking change.

james-world:
This is one of those bug fixes that should be considered a breaking change, as it's possible clients are depending on the incorrect behaviour.

RxDave:
Are you sure this is actually a bug? I think it was probably intentional. An async computation implicitly completes when the computation completes. The other overloads aren't handled automatically by the compiler (namely, Task<IDisposable>, Task<Action>) so it's the dev's responsibility to call OnCompleted themselves.

I rely on the auto-completion behavior of async iterators, so it would be nice to keep it in Rx :)

james-world:
No, I'm not sure... :) Thankfully GitHub has given us these PR discussion threads to appease my uncertainty...

I am sure that *something* is wrong here though. At the very least, the intended behaviour isn't clear. If it was intended this way, it's quite different from the non-async versions, and the other overloads.

I have a problem with assuming completion of the subscription process = completion of the whole stream. How would you write `Never` with this overload?

RxDave:
I agree that calling OnCompleted should prevent the auto-completion - no sense of having two "Done's" :)
But my point is that it doesn't typically make sense to call OnCompleted yourself in an async iterator. There's a big semantic different between the overloads that return just a Task and the ones that return an IDisposable or an Action. Returning in the former means that you're complete. That's just the semantics of an async method. However, returning an IDisposable or an Action doesn't. It implies that some other async work may be continuing; otherwise, there's no point in returning a cancellation method.
So actually I've never used the latter overloads, that I can remember anyway.
I'm sure you're aware, but just in case you're interested or somebody else reading this is interested, here's a nice little example of an async iterator state machine: 

https://gist.github.com/RxDave/9adc25455917ff12b9b8

IMHO, in this GIST it would be counter-intuitive to require a call to OnCompleted at the end of the async method passed to `Create`. It's just an async method - it completes when the method returns for the last time. What would it mean to "complete" and then continue processing, while the Task that was returned offers no way to cancel it? That's my understanding anyway, I could be in the minority :)

james-world:
Hmmm. I've mulled over this and I can't see it your way. My issue is that this is not overtly a call to the TPL - it's a call to `Observable.Create`, where we create observables of our own according to the rx grammar. I naturally expect to see the grammar honoured explicitly.

Indeed, the only "official" example I can find where these methods are used is in @bartdesmet 's post on the 2.0 beta where he gives an example (admittedly of the `IDisposable` returning form) with an explicit `OnCompleted` in a Range implementation about half way down (about page 8090 of a typically lengthed Bart "note" :))

Maybe it's just me, but it looks weird to me that your gist doesn't have an explicit call, because I just think that's how `Observable.Create` should read - even though that sample distances itself from the `Create` call by way of being in a separate method, it *still* looks weird. Most invocations wouldn't be doing that I imagine, making it far less "obvious" that TPL is involved. No explicit termination in an `Observable.Create` call that is going to terminate just feels wrong.

Finally, don't you think this line (however necessary) is a little ironic? Just a teensy bit? A smidgeon of a bit? :)

```C#
// Each observable test must never complete to avoid an exception
// thrown by Rx when awaiting an empty observable.
RunTurnstile(transitions.ToObservable()
    .Repeat(10)
    .Concat(Observable.Never<TurnstileTransition>()));  // <- Smile.
```

I do think that the inconsistency between the overloads is a real problem too. The simple repro in my opening post is just *not right*!

RxDave:
All good points. I guess I'm on the fence then. I can see it my way, but I can see it your way too. Although, I can't agree with your comment about the irony of having to concatenate to `Never`, even a little bit :) If I were to define an empty `async` method, then C# would return a completed Task, which is more to my point. The fact that `await` semantics on an observable requires at least one value seems orthogonal to me.

DanHarman:
I can see this both ways, to me the real defect is that none of the docs on msdn seem to even mention the async overloads, never mind explain that obs.OnCompleted() is not required for the Task overload.

Lukazoid:
I have found another good reason not to have this automatic call to `.OnCompleted`: https://github.com/Reactive-Extensions/Rx.NET/issues/83#issuecomment-381659478.

Lukazoid:
It seems this automatic call to `.OnCompleted` doesn't play very nicely with schedulers: https://github.com/Reactive-Extensions/Rx.NET/issues/83#issuecomment-381660735

Lukazoid:
@akarnokd But the overload which returns `Task<IDisposable>` doesn't complete when the task completes? Seems inconsistent to me.

This completes:

     Observable.Create<int>(async obs => {})

This never does:

     Observable.Create<int>(async obs => Disposable.Empty)

danielcweber:
I wouldn't expect the current behaviour, but I can see the rationale behind it and once you subscribe to that view, it feels as natural to me as if it would not call OnCompleted by itself.

At least, the sequence terminates in any case. We have a bug in Rx (I think it was SkipUntil) where a sequence would not terminate at all. In both cases, changing the behaviour could be seen as a breaking change. So I would say it is what it is, that is the way async Create works. Is everybody ok if I close this then?

PR #94, opened 2015-05-13
Fix #93

### LINKED ISSUE #93: IX-Async: SelectMany: Dispose called twice.
See repro below:

```c#
[TestMethod]
public void SelectMany_Dispose_invoked_only_once()
{
	var disposeCounter = new DisposeCounter();

	var result = AsyncEnumerable.Return(1).SelectMany(i => disposeCounter).Select(i => i).ToList().Result;

	Assert.AreEqual(0, result.Count);
	Assert.AreEqual(1, disposeCounter.DisposeCount); // DisposeCount == 2 here
}

private class DisposeCounter : IAsyncEnumerable<object>
{
	public int DisposeCount { get; private set; }

	public IAsyncEnumerator<object> GetEnumerator()
	{
		return new Enumerator(this);
	}

	private class Enumerator : IAsyncEnumerator<object>
	{
		private readonly DisposeCounter _disposeCounter;

		public Enumerator(DisposeCounter disposeCounter)
		{
			_disposeCounter = disposeCounter;
		}

		public void Dispose()
		{
			_disposeCounter.DisposeCount++;
		}

		public Task<bool> MoveNext(CancellationToken _)
		{
			return Task.FromResult(false);
		}

		public object Current { get; }
	}
}
```
cc @bartdesmet

danielcweber:
This fixes AsyncEnumerable.SelectMany. The sequence of inner IAsyncEnumerators is represented through an AssignableDisposable (equivalent to SerialDisposable from Rx). This takes care that the previous inner IAsyncEnumerator is properly disposed when a new one is assigned, and each inner IAsyncEnumerator is disposed only once.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
@@ -244,27 +244,12 @@ namespace System.Linq
 
             return Create(() =>
             {
-                // A lock seems inevitable. Disposal of the outer enumerator and completion of
-                // MoveNext of the inner enumerator can happen concurrently.
-                var syncRoot = new object();
                 var e = source.GetEnumerator();
                 var ie = default(IAsyncEnumerator<TResult>);
-
-                var disposeIe = new Action(() =>
-                {
-                    var localIe = default(IAsyncEnumerator<TResult>);
-
-                    lock (syncRoot)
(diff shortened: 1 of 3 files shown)


danielcweber:
I don't mind you copy-pasting these changes instead of pulling the commits, however, I'd be very happy if you could just acknowledge the effort e.g. in your commit message.

PR #96, opened 2015-05-18
Add overloads for AsyncEnumerable.Scan that allow async accumulators

danielcweber:
As noted in the commit messages themselves, it is subject to discussion whether these overloads should be renamed (eg. ScanAsync) since they might induce ambiguities between the different overloads (see modified unit tests).

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
@@ -2344,6 +2344,140 @@ namespace System.Linq
             });
         }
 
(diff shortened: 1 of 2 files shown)


bartdesmet:
We won't take this change right now, given it'd introduce a severe asymmetry in the API surface; all other functions passed to operators would need async overloads too. There's nothing that makes Scan special in this regard.

bartdesmet:
See https://github.com/Reactive-Extensions/Rx.NET/issues/194.

onovotny:
Closing due to age and incoming C# 8 features. Lets revisit if still an issue there.

PR #98, opened 2015-05-19
Installed Rx-Main on EventCorrelationSample project

andrewchaa:
I was reading the code in the sample project and compiled it. Rx-Main library was missing, as the referenced path was going to "Program Files (x86)\Microsoft SDKs..." It may be right if you install reactive extensions as sdk, but these days, many people prefer package installation via nuget.

So, I've added package.config and changed the path to "\packages\Rx-Core..."


LeeCampbell:
It appears that you have also added binaries to your commit. i.e. _Rx.NET/Samples/EventCorrelationSample/packages/_ folder has been committed.


**This should not be merged in its current state.**

shiftkey:
@LeeCampbell thanks for spotting this!

@andrewchaa we've published up a bunch of changes as part of the new 3.0 release - can you see if this  issue still needs to be resolved, and open a new PR against `master` with just the source changes? Thanks!

andrewchaa:
Of course. I'll have a look.

PR #100, opened 2015-05-22
Add Disposable.Compose()

bordoley:
I've added this static method to several projects i've worked on. I prefer it over using CompositeDisposable as its more functional in nature, and I've never needed the ability to add/remove disposables. Thoughts?


LeeCampbell:
See `StableCompositeDisposable` [https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/StableCompositeDisposable.cs]

Usage : 

    var disposables = StableCompositeDisposable.Create(disp1, disp2, disp3);

bartdesmet:
`StableCompositeDisposable` should be a good substitute.

PR #104, opened 2015-06-10
ToObservable: Do not call MoveNext with an already canceled CancellationToken

danielcweber:
This would happen whenever an IObserver<T>.OnNext would dispose its subscription within a call to OnNext. The built in AnonymousAsyncEnumerator<T> checks whether it's already disposed, however, not every user code might handle canceled tokens that gracefully. Also, the continuation of MoveNext will not be scheduled anyway.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/ToObservable.cs
+++ b/Ix.NET/Source/System.Interactive.Async/ToObservable.cs
@@ -202,7 +202,13 @@ namespace System.Linq
                                                  if (t.Result)
                                                  {
                                                      observer.OnNext(e.Current);
-                                                     f();
+
+                                                     if (!ctd.Token.IsCancellationRequested)
+                                                         f();
+                                                     
+                                                     //In case cancellation is requested, this could only have happened
+                                                     //by disposing the returned composite disposable (see below).
+                                                     //In that case, e will be disposed too, so there is no need to dispose e here.
                                                  }
                                                  else
                                                  {
--- a/Ix.NET/Source/Tests/AsyncTests.Conversions.cs
+++ b/Ix.NET/Source/Tests/AsyncTests.Conversions.cs
@@ -316,6 +316,124 @@ namespace Tests
             Assert.Equal(ex1, ((AggregateException)ex_).InnerExceptions.Single());
         }
 
+        [Fact]
+        public void ToObservable_disposes_enumerator_on_completion()
+        {
+            var fail = false;
+            var evt = new ManualResetEvent(false);
+
+            var ae = AsyncEnumerable.CreateEnumerable(
+                () => AsyncEnumerable.CreateEnumerator<int>(
+                    async ct => false,
+                    () => { throw new InvalidOperationException(); },
+                    () => { evt.Set(); }));
+
+            ae
+                .ToObservable()
+                .Subscribe(new MyObserver<int>(
+                    x =>
+                    {
+                        fail = true;
+                    },
+                    ex =>
+                    {
+                        fail = true;
+                    },
+                    () =>
+                    {
+                    }
+                ));
+
+            evt.WaitOne();
+            Assert.False(fail);
+        }
+
+        [Fact]
+        public void ToObservable_disposes_enumerator_when_subscription_is_disposed()
+        {
+            var fail = false;
+            var evt = new ManualResetEvent(false);
+            var subscription = default(IDisposable);
+            var subscriptionAssignedTcs = new TaskCompletionSource<object>();
+
+            var ae = AsyncEnumerable.CreateEnumerable(
+                () => AsyncEnumerable.CreateEnumerator(
+                    async ct =>
+                    {
+                        await subscriptionAssignedTcs.Task;
+                        return true;
+                    },
+                    () => 1,
+                    () => { evt.Set(); }));
+
+            subscription = ae
+                .ToObservable()
+                .Subscribe(new MyObserver<int>(
+                    x =>
+                    {
+                        subscription.Dispose();
+                    },
+                    ex =>
+                    {
+                        fail = true;
+                    },
+                    () =>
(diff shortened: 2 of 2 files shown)


Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs
```diff
@@ -287,7 +287,9 @@ public IDisposable Subscribe(IObserver<T> observer)
                         if (t.Result)
                         {
                             observer.OnNext(e.Current);
-                            f();
+
+                            if (!ctd.Token.IsCancellationRequested)
+                                f();
```
bartdesmet: This code path won't trigger a dispose of the enumerator, unlike the case where `MoveNext` observes the cancellation and causes the returned `Task` to be in a canceled state.

martinwoodward:
Hi @danielcweber - while the team are looking into this one, would you mind signing the [.NET Foundation Contribution License agreement](http://cla2.dotnetfoundation.org)? Shouldn't take long and is hopefully pretty painless. Many thanks!

martinwoodward:
Sorry - ignore me. I see you signed it already on June 8 2015

danielcweber:
I fixed the enumerator-dispose-issue. Should be good now.

onovotny:
Hi @danielcweber: can you please verify the original issue against the https://github.com/Reactive-Extensions/Rx.NET/tree/ix-optimizations branch?

I've been making a lot of improvements there based on the CoreFX designs and fixed a number of issues related to this.

Thanks!

danielcweber:
I rebased it on ix-optimizations but cannot edit the target branch, thats why we have 9 commits here. I could open a new pull request though.

onovotny:
Before rebasing it / including this fix, can we confirm that it's still needed? I made changes to the underlying AsyncIterator that has a wide effect and may mean it's not needed...so that's what I was looking to confirm?

danielcweber:
I would say the change still has a valid point, unless you check the state of the AsyncIterator after [this line](https://github.com/Reactive-Extensions/Rx.NET/pull/104/commits/e42909bebff93c49929361b58ad4f9331cc7ee2b#diff-4705a953f4a3884f7447d132bcb65a2aL58). An already cancelled token will immediately trigger Dispose, but MoveNextCore will still be executed right? I don't see anything throwing.

onovotny:
That code looks like it's missing a few commits to optimization branch from this morning.

So the scenario here is that we are not calling/checking the cancellation status before calling MoveNextCore and thus we're potentially doing more work than necessary. Fair point and I can add a throwIfCancelled before the MoveNextCore call.

Would that solve this?

danielcweber:
The point of my pull request is to avoid any throwing. The CancellationToken passed to MoveNext will be cancelled if the subscription on the observable (from `ToObservable`) is disposed. The consumer will not expect any calls to OnNext or OnError. However, if MoveNext fails, OnError will be called [here](https://github.com/Reactive-Extensions/Rx.NET/pull/104/commits/d9168aed25805ae950054afec5fba5c1c88069e3#diff-29f7a1a4434a2ed9fbd11bcd1768ce5fR193).

danielcweber:
That last statement of mine is probably false, since the continuation will not be scheduled.

onovotny:
Okay, so this fix is indeed highly specific to the IObservable impl? That's ok, thanks -- can you please rebase this back to master and we can get this included?

onovotny:
One other thing -- please include a test that verifies the desired behavior so we can ensure we don't break again.

Thanks

danielcweber:
If a canceled token makes MoveNext return false instead of throwing a TaskCanceledException (as in your [latest change](https://github.com/Reactive-Extensions/Rx.NET/commit/8908b5e8e4ce705a595e1da48b2a23cbf773638d#diff-4705a953f4a3884f7447d132bcb65a2aR91)) then we don't need the fix in this pull request at all. I'm just a little confused here - is returning false really the expected behaviour ?

onovotny:
I believe it is expected behavior as an iterator's MoveNext() method returns false if there's nothing more to do. It would be the same as if cancel was called while awaiting a sequence's MoveNext because once canceled, the iterator is in an indeterminite state.

https://msdn.microsoft.com/en-us/library/78dfe2yb(v=vs.110).aspx

Basically, once it's invalidated, then it's not possible to use that enumerator any further, which is what a `false` return indicates. 

Like calling `Cancel` "in-flight", it also destroys the instance of the interator via Dispose and ensures sequences are cleaned up. 

I don't belive a `TaskCanceledException` is warranted as that's generally used to escape out of a deep call stack. In the case where it can be safely detected and short-cirucuted, it should be ok.  

If you look at the MSDN example of using it, it's just a detection mechanism:
https://msdn.microsoft.com/en-us/library/system.threading.cancellationtoken.iscancellationrequested(v=vs.110).aspx

No gaurantees that it throws anything -- there's a different `ThrowIfCancellationRequested()` method that does that.

danielcweber:
For ToObservable, getting false leads definitely to the right behavior, and there's no uncaught exception. Still there's the unnecessary call to MoveNext but that's ok I guess. Let's leave it as it is at the moment and see if there's other opinions. I'll rebase it back on master now.

onovotny:
@danielcweber thanks! Please just add a unit test for this behavior and I'll merge it in.

onovotny:
I see a test failure in the CI build:

`Tests.AsyncTests.ToObservable_does_not_call_MoveNext_again_when_subscription_is_disposed()`

danielcweber:
I will investigate this. Test passed locally (of course, right?). Sorry for inconvenience.

danielcweber:
Tests are passing now, the failing build is probably due to the master also failing to build at the moment.

onovotny:
Yup, thanks, fixed the build script in master.

PR #121, opened 2015-08-02
Update README.md

Aleks-d:
Added a section covering books on reactive extensions.


shiftkey:
Due to the recent changes for the upcoming 3.0 release, this now has a merge conflict.

I do like this idea to track some resources for understanding Rx and Ix, perhaps as part of the documentation refresh we've been kicking around. Thoughts @onovotny @bartdesmet?

onovotny:
I think that some of this belongs on the wiki section. A wiki that includes learning resources, including books, online learning and more would be great.

PR #133, opened 2015-09-14
 Add lazy RefCount operator for IObservables.

danielcweber:
LazyRefCount connects like RefCount but may delay disconnection. This is useful whenever a lot of connect/disconnect cycles are expected within a short timespan but with a significant overhead in connecting/disconnecting.

---

This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/133)

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/IQueryLanguage.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/IQueryLanguage.cs
@@ -307,6 +307,8 @@ namespace System.Reactive.Linq
         IConnectableObservable<TSource> PublishLast<TSource>(IObservable<TSource> source);
         IObservable<TResult> PublishLast<TSource, TResult>(IObservable<TSource> source, Func<IObservable<TSource>, IObservable<TResult>> selector);
         IObservable<TSource> RefCount<TSource>(IConnectableObservable<TSource> source);
+        IObservable<TSource> RefCount<TSource>(IConnectableObservable<TSource> source, TimeSpan disconnectDelay);
+        IObservable<TSource> RefCount<TSource>(IConnectableObservable<TSource> source, TimeSpan disconnectDelay, IScheduler schedulder);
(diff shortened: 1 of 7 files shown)


RxDave:
How about just adding a `TimeSpan delayDisconnect` overload to `RefCount`, rather than introducing a new operator? (Coming from a keep-the-public-surface-area-simple and cohesive point of view.)

danielcweber:
I'm ok with that. Having an overload instead of a new operator would reflect the sanity of the idea even more and show that this is indeed useful.

shiftkey:
:thumbsup: to an overload on `RefCount`, especially if the delay is the only difference in behaviour here...

RxDave:
Would it also make sense to reuse the implementation? The classic operator would then just call the new overload passing `TimeSpan.Zero`. This would eliminate some code bloat.

danielcweber:
LazyRefCount is now only an overload for RefCount. The classic operator reuses the new implementation. Note that there's possibly room for some performance optimization since disconnection for the classic RefCount is now scheduled on the Immediate scheduler without any waiting. We might bypass this based on the delay being positive. However, this might introduce unexpected behaviour, e.g. a user could pass a scheduler and TimeSpan.Zero and expect disconnection to be scheduled (but it isn't). This is to be discussed.

Rx.NET/Source/src/System.Reactive/Linq/Observable/LazyRefCount.cs
```diff
@@ -0,0 +1,104 @@
...
+                var subscription = parent._source.SubscribeSafe(this);
+
+                lock (parent._gate)
+                {
+                    if (++parent._count == 1)
+                    {
+                        if (parent._connectableSubscription == null)
+                            parent._connectableSubscription = parent._source.Connect();
+
+                        parent._serial.Disposable = new SingleAssignmentDisposable();
```
akarnokd: I'd move this earlier so that the timeout task is cancelled as soon as possible.

PR #139, opened 2015-10-10
Fix unit tests for Ix.Net build.

sagarsumant:
Ix.Net build unit tests are failing for various reasons (mostly for ReleasePL|AnyCPU flavor). Fixed all the issues in multiple commits.


shiftkey:
@sagarsumant as part of getting Ix.NET building and testing on .NET Core in time for the upcoming 3.0 release, we should have resolved these failures. Please open a new PR if you find anything else that needs to be addressed.

PR #150, opened 2015-11-30
Make PriorityQueue<T> public. Strengthen it to implement IReadOnlyCollection<T>

hickford:
Many people would find the priority queue data structure useful 
1. https://stackoverflow.com/questions/102398/priority-queue-in-net
2. https://github.com/dotnet/corefx/issues/163

---

This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/150)

### DIFF
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/Internal/PriorityQueue.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/Internal/PriorityQueue.cs
@@ -2,10 +2,14 @@
 
 using System.Threading;
 using System.Collections.Generic;
+using System.Collections;
 
 namespace System.Reactive
 {
-    internal class PriorityQueue<T> where T : IComparable<T>
+    /// <summary>
+    /// Priority queue data structure implemented using a min-heap. 
+    /// </summary>
+    public class PriorityQueue<T> : IReadOnlyCollection<T> where T : IComparable<T>
     {
 #if !NO_INTERLOCKED_64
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/System.Reactive.Core/Reactive/Internal/PriorityQueue.cs
```diff
@@ -133,6 +158,20 @@ public bool Remove(T item)
             return false;
         }
 
+        /// <summary>
+        /// List all the elements in the priority queue.
+        /// </summary>
+        public IEnumerator<T> GetEnumerator()
+        {
+            foreach (var item in _items)
```
bartdesmet: The `_items` field has no guarantee about ordering, cf. `Enqueue` performing an append and `Percolate`. It also has a size beyond the number of elements stored (cf. `_size`). Even if it would have such guarantees, we'd need a versioning strategy here to detect changes during enumeration, which will add cost in order to support this functionality.

martinwoodward:
Hi @hickford  - while the team are looking into this one, would you mind signing the [.NET Foundation Contribution License agreement](http://cla2.dotnetfoundation.org)? Shouldn't take long and is hopefully pretty painless. Many thanks!

onovotny:
Closing due to age. I don't think we want to be focusing in on exposing collections here, but feel free to take the code and incorporate it into another collections library.

rxteam:
[Test comment to see what GitHub account is generating tons of e-mail notifications]
________________________________

PR #151, opened 2015-11-30
[WIP] Ported core projects to project.json

### LINKED ISSUE #148: CoreCLR
Please consider creating an implementation that supports CoreCLR target platforms.

khellang:
I'm opening this up early to see if anyone has some feedback... :eyes:

Currently I have the 4 main projects building for `net40`, `net45` and `dotnet5.4` (`netstandard1.3`), without *any* code changes :smile:
This should cover a minimum of `net40`, `net45`, `net46`, `uap10`, `netcore50` and `dnxcore50`. I *think* this includes the Xamarin products as well :sparkles:

I tried to aim for `dotnet5.2` (`netstandard1.1`), which would include minimum `win8` and `wpa8` as well, but could only get **Interfaces** to compile. Will give it another go soon. :watch:
I currently don't have the tooling for sl5, wp7 and wp8, so I punted on those for now :dancers:

All the preprocessor directives are directly from [Common.targets](https://github.com/Reactive-Extensions/Rx.NET/blob/2252cb4edbb25aca12005b9a912311edd2f095f3/Rx.NET/Source/Common.targets), so hopefully they should be good. Anyway, they're a horrible mess and should :fire: ASAP, once the older targets are deprecated.

There are still a few features that are not supported for project.json yet, like .tt-files etc., so this'll probably have to stay side-by-side for a while :disappointed: 

For reference, see the [standard platform document](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/standard-platform.md).

Should close #148

### DIFF
--- a/Rx.NET/Source/System.Reactive.Linq/project.json
+++ b/Rx.NET/Source/System.Reactive.Linq/project.json
@@ -1,0 +1,72 @@
+﻿{
+    "version": "3.0.0-*",
+    "title": "Reactive Extensions - Query Library",
+    "description": "Reactive Extensions Query Library used to express complex event processing queries over observable sequences.",
+    "authors": [ "Microsoft" ],
+    "copyright": "Copyright (C) Microsoft Corporation",
+    "tags": [ "Rx", "Reactive", "Extensions", "Observable", "LINQ", "Events" ],
+    "iconUrl": "http://go.microsoft.com/fwlink/?LinkId=261274",
+    "projectUrl": "http://go.microsoft.com/fwlink/?LinkId=261273",
+    "licenseUrl": "http://go.microsoft.com/fwlink/?LinkID=261272",
+    "requireLicenseAcceptance": true,
+
+    "exclude": [ "*/**/ImmutableList.cs" ],
+
+    "dependencies": {
+        "System.Reactive.Core": { "target": "project" }
+    },
+
+    "frameworks": {
+        "net40": {
+            "compilationOptions": {
+                "define": [
+                    "NO_TASK_DELAY",
+                    "HAS_APTCA",
+                    "HAS_WINFORMS",
+                    "USE_TIMER_SELF_ROOT",
+                    "NO_WEAKREFOFT"
+                ]
+            }
+        },
+        "net45": {
+            "compilationOptions": {
+                "define": [
+                    "NO_EVENTARGS_CONSTRAINT",
+                    "HAS_EDI",
+                    "HAS_WINRT",
+                    "HAS_PROGRESS",
+                    "PREFER_ASYNC",
+                    "HAS_AWAIT",
+                    "HAS_APTCA",
+                    "HAS_DISPATCHER_PRIORITY",
+                    "HAS_WINFORMS",
+                    "USE_TIMER_SELF_ROOT"
+                ]
+            }
+        },
+        "dotnet5.4": {
+            "compilationOptions": {
+                "define": [
+                    "NO_EVENTARGS_CONSTRAINT",
+                    "HAS_EDI",
+                    "HAS_WINRT",
+                    "HAS_PROGRESS",
+                    "PREFER_ASYNC",
+                    "HAS_AWAIT",
+                    "HAS_APTCA",
+                    "NO_REMOTING",
+                    "NO_SERIALIZABLE",
+                    "NO_THREAD",
+                    "CRIPPLED_REFLECTION",
+                    "PLIB",
+                    "USE_TIMER_SELF_ROOT"
+                ]
+            },
+            "dependencies": {
+                "System.Collections.Concurrent": "4.0.11-beta-23516",
+                "System.Reflection.TypeExtensions": "4.1.0-beta-23516",
+                "System.Runtime.InteropServices.WindowsRuntime": "4.0.1-beta-23516"
+            }
+        }
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,9 @@ ClientBin
 stylecop.*
 ~$*
 *.dbmdl
+.vs/
+project.lock.json
+artifacts/
 Generated_Code #added for RIA/Silverlight projects
 
 # Backup & report files from converting an old project file to a newer
--- a/Rx.NET/Source/Rx-New.sln
+++ b/Rx.NET/Source/Rx-New.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.23107.0
+VisualStudioVersion = 14.0.25008.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "System.Reactive.Core", "System.Reactive.Core\System.Reactive.Core.xproj", "{FDA62C35-4174-44B0-BEBE-61E80B2515BA}"
 EndProject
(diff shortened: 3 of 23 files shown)


Rx.NET/Source/System.Reactive.Linq/project.json
```diff
@@ -0,0 +1,72 @@
...
+    "description": "Reactive Extensions Query Library used to express complex event processing queries over observable sequences.",
+    "authors": [ "Microsoft" ],
+    "copyright": "Copyright (C) Microsoft Corporation",
+    "tags": [ "Rx", "Reactive", "Extensions", "Observable", "LINQ", "Events" ],
+    "iconUrl": "http://go.microsoft.com/fwlink/?LinkId=261274",
+    "projectUrl": "http://go.microsoft.com/fwlink/?LinkId=261273",
+    "licenseUrl": "http://go.microsoft.com/fwlink/?LinkID=261272",
+    "requireLicenseAcceptance": true,
+
+    "exclude": [ "*/**/ImmutableList.cs" ],
```
khellang: For some reason this file caused trouble because of a duplicate declaration in the **Core** project. This one was picked over the other, but it lacked the `ImmutableList<T>.Empty` property, so it wouldn't compile. The implementations looks pretty similar - is there a reason why this exists?

shiftkey:
@khellang is there anything in particular I can help with here? (feedback, testing etc)

khellang:
I don't know. There are still quite a few projects that needs porting, but it would be nice with some feedback on the target platforms. What do we want to support going forward?

shiftkey:
> What do we want to support going forward?

For simplicity's sake, I'd target whatever can be built OOTB in VS2015. But I'm feeling a bit radical this evening and really want to simplify contributing to the codebase.

onovotny:
Is there any reason to support Net4? Microsoft has killed it support-wise too. Only .NET 4.5.2 and higher will be getting security updates.

I don't think there's any point in supporting WP7 or SL5. WP8 is still a stretch, but maybe leave that one for another version/year and kill it then.

CADbloke:
> Is there any reason to support Net4?
I know of weird edge-cases like older versions of AutoCAD that fall over when they see modern things. Most of this was fixed in their service packs but some things break obscurely. Honestly, if a client wanted me to support an older version I'd either drag them kicking and screaming into the present day or make them pay for the custom tweakage. Or walk away.

tldr; Don't listen to the laggards. Modern is good.

shiftkey:
@onovotny @CADbloke excellent points

@onovotny 

> WP8 is still a stretch, but maybe leave that one for another version/year and kill it then.

We're talking about the `wp` [Target Platform Name](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#mapping-the-net-platform-standard-to-platforms) here and not `wpa`, right?

onovotny:
Yes

onovotny:
Otoh, rxui 6.5 still works for wp8. Maybe housecleaning is in order. :fire: Wp8 now?

shiftkey:
@khellang I sat down and tried to get the test project `Tests.System.Reactive` building - unfortunately that's all using MSTest so it's not really ideal. But here's some more hacking I got to today:

https://github.com/khellang/Rx.NET/compare/core-port...shiftkey:some-build-script-improvements?expand=1

There's a few other projects that are needed by the tests:

 - `System.Reactive.Providers`
 - `System.Reactive.Observable.Aliases`

So I'm just wrapping my head around those and any portability issues right now...

M-Zuber:
I'd like to have a shot at converting the tests to xunit. I hope to start over the weekend and will update after mondayish if it looks like something I can pull through to the finish. Sounds okay?

shiftkey:
@M-Zuber :thumbsup: I'm curious what you find out about the migration!

khellang:
@M-Zuber Before you start doing anything manually, you should know that https://github.com/dotnet/codeformatter should be able to translate MSTest to xUnit automatically :smile:

M-Zuber:
:thumbs-up: that's what I was planning on using 
:grinning:

Let's hope that all we need to do is run that

shiftkey:
Just a heads up - if we're going to opt-in for the new way of creating packages, the parent folders must be renamed to match the existing package, e.g `System.Reactive.Interfaces` becomes `Rx-Interfaces`

onovotny:
Aren't Xamarin specific impls still needed for some of the light-up? If so, then those can't yet be built with xproj, afaik.

shiftkey:
@onovotny I have no idea what the support looks like over on that side of the fence...

shiftkey:
@onovotny it's been a while since I've been looked at WinRT (is it now what we call UAP?), and I'm kinda stumped as to how to resolve types under the `Windows.*` namespace now that all the dependencies are distributed via NuGet, e.g [`TypedEventHandler<TSender, TEventArgs>`](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.WindowsRuntime/EventPatternSource.cs#L15)

> C:\Users\shiftkey\Documents\GitHub\Rx.NET\Rx.NET\SOurce\System.Reactive.WindowsRuntime\EventPatternSource.cs(15,15): .NETCore,Version=v5.0 error CS0246: The type or namespace name 'TypedEventHandler<TSender, TEventArgs>' could not be found (are you missing a using directive or an assembly reference?)

Some other errors talk about `Windows, Version=255.255.255.255` but that's clearly not right:

> The type 'IAsyncActionWithProgress<>' is defined in an assembly that is not referenced. You must add a reference to assembly 'Windows, Version=255.255.255.255, Culture=neutral, PublicKeyToken=null, ContentType=WindowsRuntime'.

Even referencing [`Microsoft.NETCore.UniversalWindowsPlatform`](https://www.nuget.org/packages/Microsoft.NETCore.UniversalWindowsPlatform/) didn't seem to resolve this, so I'm kinda puzzled what I'm missing here. Thoughts?

onovotny:
I believe those types come from Windows.winmd? That's not currently on NuGet; that's part of the SDK. It also may not be possible to build with xproj.

shiftkey:
@onovotny basically

>  C:\Program Files (x86)\Windows Kits\10\References\Windows.Foundation.FoundationContract\1.0.0.0\Windows.Foundation.FoundationContract.winmd

Oh well, back to the drawing board on that project...

onovotny:
That's not quite the reference you'd want, it'd be 
> C:\Program Files (x86)\Windows Kits\10\UnionMetadata\Windows.winmd

However, all is not lost, there might still be a way, but it might require "borrowing" some of the private targeting packages that the corefx team is using

https://github.com/dotnet/corefx/tree/8737f673b55ccd3eee7fbb8f914b5bcbb7c44d7b/src/System.Runtime.WindowsRuntime/src

Needs some spelunking, but I see `Microsoft.TargetingPack.Private.WinRT": "1.0.1"`

onovotny:
And here it is:

`https://dotnet.myget.org/F/dotnet-core/` or 
`https://dotnet.myget.org/F/dotnet-core/api/v3/index.json`

Then look for `Microsoft.TargetingPack.Private.WinRT` 1.0.1

It has a `lib\dotnet5.1\windows.winmd` in it.

I know it'd be better to use the main nuget feeds, but perhaps later once stuff is baked?

M-Zuber:
@khellang the [MSFT too](https://github.com/dotnet/codeformatter/tree/master/src/XUnitConverter) didn't seem to work for me.
In addition we have to use xUnit 1.9.2 until `4.0` is dropped.
Now converting by hand :muscle: and will make a PR onto this when I am done

onovotny:
Don't use xUnit 1.9.2. It's really old and won't work on devices. Just make sure your unit test project is .NET 4.5.

onovotny:
Or better yet, put the unit tests into a shared code project so they can be compiled on different platforms for per-platform testing.

M-Zuber:
> Don't use xUnit 1.9.2. It's really old and won't work on devices. Just make sure your unit test project is .NET 4.5.

I get this, which I do not know where to find
image

>Or better yet, put the unit tests into a shared code project so they can be compiled on different platforms for per-platform testing.

Currently I am working off master. Do I need to be switched to be based off this PR in order to use a shared code project? (either way a pointer or two in how to do would be appreciated as I am not that sure exactly what you mean)

LeeCampbell:
+1 for dropping .NET 4.0 support. It seems at odds with itself to expect users who don't upgrade (or are on WinXP) to jump to the latest build of Rx :-)

khellang:
I'll just close this. There's work, based on this, that will bring Rx to the new world of .NET core in the pipeline. Watch #148

xperiandri:
How to target UWP APIs in .NET Core library project with `uap10.0` profile?
@onovotny solution for RT and `dotnet5.1` works, but what about `uap10.0`?

onovotny:
you can use an `"imports": ["dotnet5.4"]` in the `uap10.0` framework section.

xperiandri:
And just use the `Microsoft.TargetingPack.Private.WinRT`?

xperiandri:
Like this?
``` JSON
{
  "version": "1.0.0-*",

  "frameworks": {
    ".NETPortable,Version=v4.6,Profile=Profile32": {
      "imports": "dotnet5.1",
      "frameworkAssemblies": {
        "mscorlib": "",
        "System": "",
        "System.Core": "",
        "System.Collections": "",
        "System.Globalization": "",
        "System.Linq": "",
        "System.Linq.Expressions": "",
        "System.Reflection": "",
        "System.Reflection.Extensions": "",
        "System.Runtime": "",
        "System.Runtime.Extensions": "",
        "System.Runtime.InteropServices": "",
        "System.Windows": ""
      },
      "dependencies": {
        "Microsoft.TargetingPack.Private.WinRT": "1.0.1"
      }
    },
    "uap10.0": {
      "buildOptions": {
        "define": [ "WINDOWS_UWP" ]
      },
      "imports": "dotnet5.4",
      "dependencies": {
        "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.0",
        "Microsoft.TargetingPack.Private.WinRT": "1.0.1"
      }
    }
  },
  "buildOptions": {
    "xmlDoc": true
  }
}
```

PR #152, opened 2015-12-02
Reverse subscription order in WithLatestFrom. In case both observable…

danielcweber:
…s produce values immediately on subscription, I claim that most of the time, the expected behavior is to already have the value from the context-observable when source produces a value.

---

This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/152)

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
@@ -1516,6 +1516,8 @@ namespace System.Reactive.Linq
 
         /// <summary>
         /// Merges two observable sequences into one observable sequence by combining each element from the first source with the latest element from the second source, if any.
+        /// Starting from Rx.NET 4.0, this will subscribe to <paramref name="second"/> before subscribing to <paramref name="first" /> to have a latest element readily available
+        /// in case <paramref name="first" /> emits an element right away.
         /// </summary>
         /// <typeparam name="TFirst">The type of the elements in the first source sequence.</typeparam>
         /// <typeparam name="TSecond">The type of the elements in the second source sequence.</typeparam>
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/WithLatestFrom.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/WithLatestFrom.cs
@@ -49,8 +49,8 @@ namespace System.Reactive.Linq.ObservableImpl
                 var fstO = new FirstObserver(this);
                 var sndO = new SecondObserver(this, sndSubscription);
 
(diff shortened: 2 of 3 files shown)


sramshaw:
OK
this improves the operator, as changing subscription order did help TakeUntil

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/152#-:-KXHtFQQ6JRIybY-jRJW:b5lrfjq)*

danielcweber:
@onovotny @shiftkey Can we please get an opinion on this issue here, it's of interest to at least some other community member (#300) and has bitten me recently again as well... thanks a lot.

akarnokd:
For reference, RxJava's `withLatestFrom` also subscribes to the other source first then to the main source. 

Not sure about Rx.NET 3.x but in RxJava 2.x, the general pattern is to subscribe to all the secondary sources before  subscribing to the main source in case the main source is an eager one and doesn't return control after `subscribe()` (such as a very long `range()` for example).

khellang:
It's been over a year since the last comment. Is this something that could go into 4.0?

danielcweber:
@onovotny @ghuntley @akarnokd This has been rebased to resolve merge conflicts. It's a breaking change and still very reasonable IMO, so we might want to discuss it a bit before the stable release.

danielcweber:
@onovotny @akarnokd Added some doc. Good to go then?

Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
```diff
@@ -1516,6 +1516,8 @@ public static IObservable<TSource> Switch<TSource>(this IObservable<Task<TSource
 
         /// <summary>
         /// Merges two observable sequences into one observable sequence by combining each element from the first source with the latest element from the second source, if any.
+        /// Starting from Rx.net 4.0, this will subscribe to <paramref name="second"/> before subscribing to <paramref name="first" /> to have a latest element readily available
```
akarnokd: I would have woven this into the first sentence bot okay. Also not sure how to refer to 4.0: `System.Reactive 4.0`, `Rx.NET 4.0`, `dotnet/reactive 4.0`, or just saying `Note that version 4.0 changed the order in which the given observable sequences are subscribed to from first->second to second->first.` in the remarks?

akarnokd: Capitalization: `Rx.net 4.0` -> `Rx.NET 4.0`

danielcweber: yep.

PR #163, opened 2016-02-16
Added SafeDisposable

Tuatan:
Some classes doesn't implement [IDisposable](https://msdn.microsoft.com/en-us/library/system.idisposable(v=vs.110).aspx) correctly and their Dispose implementation might throw.

### DIFF
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SafeDisposable.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SafeDisposable.cs
@@ -11,8 +11,6 @@ namespace System.Reactive.Disposables
     {
         private IDisposable disposable;
 
-        private bool disposed;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="SafeDisposable"/> class.
         /// </summary>
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SafeDisposable.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SafeDisposable.cs
@@ -7,9 +7,11 @@ namespace System.Reactive.Disposables
     /// <summary>
     /// Represents a resource that could be safely disposed in thread safe manner if that resource is a disposable one.
     /// </summary>
-    public sealed class SafeDisposable : IDisposable
+    public sealed class SafeDisposable : ICancelable
     {
         private IDisposable disposable;
+
+        private bool disposed;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="SafeDisposable"/> class.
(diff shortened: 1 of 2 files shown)


LeeCampbell:
This appears to be very similar to `AnonymousDisposable`, but public. Instead of having this type, users can use `Disposable.Create()` or alternatively `SingleAssignmentDisposable`.

No test coverage.

Tuatan:
All disposable implementations could be created through `Disposable.Create()`, the point is to have a reusable fragment that allows to user code less. And this one cannot be replaced by  `SingleAssignmentDisposable `.

Test coverage could be added.

LeeCampbell:
My point was that we could consider making `AnonDisposable` public (as it already does exactly what this new type does). Alternatives to

    var safeDisposable = new SafeDisposable(other);

could be 

    var safeDisposable = new SingleAssignmentDisposable { Disposable = other};

or 

    var safeDisposable = Disposable.Create(()=>other.Dispose());

With regards to using `Disposable.Create()` for all disposable implementations, that not true. Any of the mutable instances like `SingleAssignmentDisposable`, `SerialDisposable`, `MultiAssignmentDisposable`, `CompositeDisposable` could not be recreated with `Disposable.Create` without a serious amount of hackery.

I am not saying that we should not include `SafeDisposable`, but should first consider what other options there are before increasing the public surface area of the API.

LeeCampbell:
And apologies, I missed the Try/Catch/Swallow part. I would not want this in the code base. Compensating behaviors for poorly designed APIs is a slippery slope in my opinion.

Tuatan:
Yes, you are obviously right here, and according to the guidelines the implementation of Dispose should not throw, but even Task from TPL/FCL does throw on Dispose.

RxDave:
It's true that `Task` throws on `Dispose` only in the case where disposing is an invalid operation, but it's not a great example of "acceptable" behavior given that, unlike most `IDisposable` implementations, the recommendation is to **not** dispose an instance of `Task` except in rare performance-related circumstances. (See http://blogs.msdn.com/b/pfxteam/archive/2012/03/25/10287435.aspx). The guideline that `Dispose` should never throw is still one that applies whenever it's recommended that calling `Dispose` is generally the right thing to do.

Tuatan:
Dave, you are also correct here, usually task instance does not hold any disposable resources. I do realize that Try/Catch/Swallow is not particularly clean pattern to put it mildly, but an alternative option does not look really good as well:

`disposableInstance is Task ? Disposable.Empty : new SingleAssignmentDisposable{ Disposable =  disposableInstance }`

And in this cause you have to rely on application components that they did that as well.

shiftkey:
Thanks for opening this @Tuatan but I'm :thumbsdown: on this change without understanding more about the context for the problem we're tackling here.

Now that we're close to getting 3.0 out the door, I think we can do something more elegant here.

PR #167, opened 2016-03-10
Fix the logic of the Sample operator

### LINKED ISSUE #134: WeakSubscribe
Please consider adding WeakSubscribe extension methods that have the same signatures as the Subscribe extension methods, but use an AnonymousWeakObserver class that holds weak references to the onNext, onError, and onComplete delegates instead of the existing AnonymousObserver class.

The AnonymousWeakObserver should dispose itself if the target of the any of the weak references is collected as garbage.  It should also emulate AutoDetachObserver and dispose an attached Disposable (from the Observable) when it is disposed.

WeakSubscribe is useful when you need a subscription lifetime tied to an object that doesn't provide a hook for calling dispose on the IDisposable returned by IObservable.Subscribe.  WPF controls are an example.  The Presentation Framework internally makes liberal use of weak references to prevent data bindings from tying the lifetime of a target to the source.  Rx can do the same, in a more elegant manner, with WeakSubscribe extension methods.

If there is some other mechanism in Rx that makes this unnecessary or a bad idea? If so, I would love to learn what I am doing wrong.

bnayae:
Fixes #134

When the source observable complete
the Sample immediately sending the last value 
and dispose the periodic schedule
rather than waiting for the periodic schedule event 
(which may take take long)

### DIFF
--- a/Rx.NET/Source/System.Reactive.Linq/Reactive/WeakObservable.Extensions.cs
+++ b/Rx.NET/Source/System.Reactive.Linq/Reactive/WeakObservable.Extensions.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
+
+using System.ComponentModel;
+using System.Reactive;
+using System.Reactive.Disposables;
+using System.Threading;
+
+namespace System.Reactive.Linq
+{
+    /// <summary>
+    /// Provides a set of static methods for subscribing delegates to observables.
+    /// </summary>
+    public static class WeakObservableExtensions
+    {
+        #region ToWeakObservable
+
+        /// <summary>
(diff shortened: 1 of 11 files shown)


Rx.NET/Source/System.Reactive.Linq/Reactive/WeakObservable.Extensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Reactive;
+using System.Reactive.Disposables;
+using System.Threading;
+
+namespace System.Reactive.Linq
+{
+    /// <summary>
+    /// Provides a set of static methods for subscribing delegates to observables.
+    /// </summary>
+    public static class WeakObservableExtensions
```
bartdesmet: These changes are unrelated? Please narrow down the PR to the issue described.

PR #170, opened 2016-04-08
Using async/await in System.Interactive.Async

aelij:
Hi,

This PR is a precursor to implementing all the operators with async delegates. With async/await this will be very easy to implement. It also simplifies the code a great deal.

For example:

```csharp
IAsyncEnumerable<T> WhereAsync<T>(this IAsyncEnumerable<T> enumerable, Func<T, Task<bool>> predicate)
```

My motivation is the release of Service Fabric, which exposes its own `IAsyncEnumerable<T>` interface. With a simple wrapper (such as the one [here](https://gist.github.com/aelij/987d974c811865029564f1bbeffb6b47)) it would be possible to use `System.Interactive.Async` with it. Since SF reliable collections only expose an async enumerable, this is the only option to use LINQ. The async delegate versions of the operators are also very useful in cross-collection queries (as you can see in the aforementioned Gist).

I realize this is a pretty big (unsolicited ;) change, but hopefully you'll consider it.

Eli.

---
This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/170)


onovotny:
Hi @aelij, can you please rebase this on to the `EF-Fixes` branch?

aelij:
Sure. Does it mean I'll need to open a new PR?

onovotny:
@aelij if you rebase and force push to the same branch, it'll keep it associated with this PR

onovotny:
Does this PR only contain async/await impls or does it also include things like `WhereAsync`? I'd keep new methods as a separate discussion/PR.

aelij:
Just async/await for current methods.

onovotny:
I was about to do the same in the `EF-Fixes` branch, so I'm glad I noticed this :) The biggest thing in that branch is a copy/use of `Lookup<T,V>` to better implement `GroupJoin`

aelij:
@onovotny, I'm trying to run the tests after the rebase with no success. The Test Explorer doesn't discover them, ReSharper won't run project.json test projects at all, and `dotnet test` produces the following error:

```
System.Interactive.Async>dotnet test
dotnet-test Error: 0 : System.InvalidOperationException: Can not find runtime target for framework '.NETStandard,Version=v1.0' compatible with one of the target runtimes: 'win10-x64, win81-x64, win8-x64, win7-x64'. Possible causes:
1. The project has not been restored or restore failed - run `dotnet restore`
2. The project does not list one of 'win10-x64, win81-x64, win8-x64, win7-x64' in the 'runtimes' section.
   at Microsoft.DotNet.ProjectModel.BuildWorkspace.GetRuntimeContext(ProjectContext context, IEnumerable`1 runtimeIdentifiers)
   at Microsoft.DotNet.Tools.Test.TestCommand.<>c__DisplayClass2_0.<DoRun>b__0(ProjectContext c)
   at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Microsoft.DotNet.Tools.Test.TestCommand.DoRun(String[] args)
```

Any idea how to fix this? Thanks.

onovotny:
For the code currently in `master`/`EF-Fixes`, it requires a newer set of CLI tools -- basically ones close to the RTM. You should be able to get the latest .NET Core SDK installer from here:

https://github.com/dotnet/cli

BTW, on/after Monday, you'll want to run `nuget locals all -clear` to ensure you have the "real" versions of the final packages.

aelij:
Can this wait until Monday? I'd rather not install nightlies :)

onovotny:
Indeed, thanks!

aelij:
I've run the tests, and all of them pass except `Do5` and `Do6`. This is because the exception received in delegate of the `Do` operator is no longer an `AggregateException` but rather the inner one, as thrown by the `async` operator.

The question now is whether we take this breaking change and conform to the `async` way, or keep the original behavior.

WDYT?

onovotny:
@aelij Changing from `AggregateException` to the inner one is ok and would be expected.

aelij:
I've force-pushed the branch. Cool, I see you've added CI builds!

Regarding the `Do` change, if you're planning on writing some release notes, this should be mentioned.

onovotny:
As this is a large PR, this tool may help navigate the diffs better than the built-in viewer:
https://reviewable.io/reviews/Reactive-Extensions/Rx.NET/170

onovotny:
Review status: 0 of 21 files reviewed at latest revision, 1 unresolved discussion.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 1712 \[r2\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLSs-dPxrD9l99i-Aio:-KLSs-dQ2Qy0TdvWU6YC:2099085386) ([raw file](https://github.com/reactive-extensions/rx.net/blob/5d350fb041918a86a284318b5186103218d93215/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L1712)):*
> ```Smalltalk
>             {
>                 if (!await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                     throw new InvalidOperationException("Source sequence doesn't contain any elements.");
> ```

Please update this to be `Strings.NO_ELEMENT`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

onovotny:
Reviewed 6 of 13 files at r1, 15 of 15 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
```
C:\Users\shiftkey\Documents\GitHub\rx.net\Ix.NET\Source\System.Interactive.Async\AsyncEnumerable.Single.cs(1289,13): error CS1997: Since 'AsyncEnumerable.ForEachAsync<TSource>(IAsyncEnumerable<TSource>, Action<TSource, int>, CancellationToken)' is an async method that returns 'Task', a return keyword must not be followed by an object expression. Did you intend to return 'Task<T>'?
```

aelij:
@shiftkey Should be okay now. I missed some files in the EF-fixes merge.

onovotny:
Reviewed 6 of 6 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 37 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiUquYEOeTriVZc7w4:-KLiUquYEOeTriVZc7w5:-567616862) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L37)):*
> ```Smalltalk
>                 }
>             }
>             var result = resultSelector(acc);
> ```

:lipstick: skip declaring this variable and just return it directly

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 370 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiWGC1wInxQVM5w1kj:-KLiWGC2RGorptIUrC_6:1387432558) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L370)):*
> ```Smalltalk
>                 }
> 
>                 TSource result = e.Current;
> ```

:lipstick: `var result` here

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 611 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiXah9ybqX6b1WPQWB:-KLiXah9ybqX6b1WPQWC:831826304) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L611)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     int? v = e.Current;
> ```

:lipstick: `var v = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 654 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiXmefOUbY1y4cAIrz:-KLiXmefOUbY1y4cAIs-:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L654)):*
> ```Smalltalk
>                 }
> 
>                 long sum = e.Current;
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 683 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiXwogUJ7d8LfOiTLj:-KLiXwogUJ7d8LfOiTLk:831826304) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L683)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     long? v = e.Current;
> ```

:lipstick: `var v = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 686 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiY17393e94XVd7ffn:-KLiY17393e94XVd7ffo:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L686)):*
> ```Smalltalk
>                     if (v.HasValue)
>                     {
>                         long sum = v.GetValueOrDefault();
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 726 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiY7z8CTU4ja570vuF:-KLiY7z8CTU4ja570vuG:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L726)):*
> ```Smalltalk
>                 }
> 
>                 double sum = e.Current;
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 755 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiYIVH-vnxgRORdoR3:-KLiYIVH-vnxgRORdoR4:831826304) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L755)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     double? v = e.Current;
> ```

:lipstick: `var v = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 758 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiYLygZ4DIbOPvj41M:-KLiYLygZ4DIbOPvj41N:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L758)):*
> ```Smalltalk
>                     if (v.HasValue)
>                     {
>                         double sum = v.GetValueOrDefault();
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 824 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLiYWQlB2AU95tTrCrT:-KLiYWQlB2AU95tTrCrU:831826304) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L824)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     float? v = e.Current;
> ```

:lipstick: `var v = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 867 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLib_ja1AMYCROC38yU:-KLib_ja1AMYCROC38yV:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L867)):*
> ```Smalltalk
>                 }
> 
>                 decimal sum = e.Current;
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 893 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLibj7LPF4gYmn2IDai:-KLibj7LPF4gYmn2IDaj:831826304) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L893)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     decimal? v = e.Current;
> ```

:lipstick: `var v = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 896 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLibmNFrLuPWnL01uja:-KLibmNFrLuPWnL01ujb:267906507) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L896)):*
> ```Smalltalk
>                     if (v.HasValue)
>                     {
>                         decimal sum = v.GetValueOrDefault();
> ```

:lipstick: `var sum = ...`

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 1623 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLicAsM_f6USmkziBqq:-KLicAsM_f6USmkziBqr:-1139686413) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L1623)):*
> ```Smalltalk
>                 throw new ArgumentNullException(nameof(source));
> 
>             return IsEmpty_(source, cancellationToken);
> ```

:lipstick: maybe inline this method as there are no other overloads?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLidNlDupPKoNoLfdzG:-2058957729) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
> ```Smalltalk
> 
>                 return Create(
>                     ct =>
> ```

We don't seem to respect the provided `CancellationToken` inside the provided `moveNext` - do you think we should?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLif7CWScnXpaiXrgGX:-1099820686) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
> ```Smalltalk
>         }
> 
> #if !NO_RXINTERFACES
> ```

I gather this is just some vestigal code that we no longer #ifdef away @onovotny?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 34 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLifqohiBXutvg1OU9n:-KLifqohiBXutvg1OU9o:-1583720213) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L34)):*
> ```Smalltalk
>         }
> 
>         static IAsyncEnumerator<T> Create<T>(Func<CancellationToken, Task<bool>> moveNext, Func<T> current,
> ```

add explicit `private` modified to make this consistent with the rest of the coding style

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
Note: I'm really not sure how to say "review still in progress" with reviewable but I still need to look at `AsyncEnumerable.Single.cs` and below. Also, the stuff I'm finding is mostly pedantic/coding style so it should be easy to resolve.

onovotny:
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLiwBF5Otvdf6gkI6Vi:727380982) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> I gather this is just some vestigal code that we no longer #ifdef away @onovotny?
</details>
I'd say that we should clean up vestigial #ifdef's in a separate issue/PR and to ignore it for now here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLnRNJeejSNOrNxP450:-1392062349) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
> to ignore it for now here

So we should leave this as-is or undo this change?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 358 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnTadYgZY2VA8a2KFW:-KLnTadYgZY2VA8a2KFX:504300574) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L358)):*
> ```Smalltalk
>                         if (await e.MoveNext(cts.Token).ConfigureAwait(false))
>                         {
>                             if (predicate(e.Current))
> ```

:lipstick: `return predicate(e.Current);` for a bit of concise-ness?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 392 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnTrYg2zjnoZxwQQ-P:-KLnTrYg2zjnoZxwQQ-Q:-73695766) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L392)):*
> ```Smalltalk
>                         if (await e.MoveNext(cts.Token).ConfigureAwait(false))
>                         {
>                             if (predicate(e.Current, checked(index++)))
> ```

:lipstick: `return predicate(e.Current, checked(index++));` for a bit of concise-ness?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 900 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnUynefIkiDOitbsDO:-KLnUynefIkiDOitbsDP:-2050360982) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L900)):*
> ```Smalltalk
>                     {
>                         res = await e.MoveNext(ct).ConfigureAwait(false);
>                         if (res == true)
> ```

:lipstick: do we need the `== true` bit here?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1120 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnVSM2oBbUfuoa2shJ:-KLnVSM2oBbUfuoa2shK:972441860) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1120)):*
> ```Smalltalk
>                     {
>                         exception?.Throw();
>                         return false;
> ```

I *think* this is fine to leave here, as the old code using `TaskCompletionSource` meant that `TrySetException` wasn't going to throw the exception (but `.Throw()` here should).

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1232 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnW-7NjC5FlNBJs6Hf:-KLnW-7OiYLwNQ1-gcDe:-1305506739) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1232)):*
> ```Smalltalk
>                         return result;
>                     }
>                     catch (OperationCanceledException)
> ```

Can we document why this is an important scenario to catch and rethrow explicitly, rather than fall through to the `onError`?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1299 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnXJjgjb_MYAdUJRO2:-KLnXJjgjb_MYAdUJRO3:362741845) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1299)):*
> ```Smalltalk
>                 while (await e.MoveNext(cancellationToken).ConfigureAwait(false))
>                 {
>                     action(e.Current, checked(index++));
> ```

This is an introduced `checked` here - is this just to be consistent or did you encounter something unexpected here during the migration?

---

*[Ix.NET/Source/System.Interactive.Async/EmptyArray.cs, line 1 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnYp_vMsCP_B0TzfIc:-KLnYp_wrqf2zDtJx1tj:827332171) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/EmptyArray.cs#L1)):*
> ```Smalltalk
> // Copyright (c) Microsoft Open Technologies, Inc. All rights reserved. See License.txt in the project root for license information.
> ```

As this is a new file, please use the new license header:

```
// Licensed to the .NET Foundation under one or more agreements.
```

---

*[Ix.NET/Source/System.Interactive.Async/Grouping.cs, line 1 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnYuPFGhzZPgOZLLnG:-KLnYuPFGhzZPgOZLLnH:827332171) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Grouping.cs#L1)):*
> ```Smalltalk
> using System;
> ```

As this is a new file, please use the new license header:

```
// Licensed to the .NET Foundation under one or more agreements.
```

---

*[Ix.NET/Source/System.Interactive.Async/Grouping.cs, line 9 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZpWWHciRyDKLLjsa:-KLnZpWWHciRyDKLLjsb:639229180) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Grouping.cs#L9)):*
> ```Smalltalk
> namespace System.Linq.Internal
> {
>     internal class Grouping<TKey, TElement> : IGrouping<TKey, TElement>, IList<TElement>
> ```

Actually, this looks really familiar - if we're pulling in this from corefx can we link back and attribute accordingly?

Something like this:

```csharp
/// adapted from System.Linq.Grouping from .NET Framework
/// source: https://github.com/dotnet/corefx/blob/b90532bc97b07234a7d18073819d019645285f1c/src/System.Linq/src/System/Linq/Grouping.cs#L64
```

I've just used the latest commit on `master` as the reference here, you might have used a different source.

---

*[Ix.NET/Source/System.Interactive.Async/Lookup.cs, line 1 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnYxteXgHrHzmhzuRU:-KLnYxteXgHrHzmhzuRV:827332171) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Lookup.cs#L1)):*
> ```Smalltalk
> using System;
> ```

As this is a new file, please use the new license header:

```
// Licensed to the .NET Foundation under one or more agreements.
```

---

*[Ix.NET/Source/System.Interactive.Async/Lookup.cs, line 9 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLn_g0ZgbQBW7OSQx_s:-KLn_g0ZgbQBW7OSQx_t:228356049) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Lookup.cs#L9)):*
> ```Smalltalk
> using System.Threading.Tasks;
> 
> namespace System.Linq.Internal
> ```

Same thing here - I'd like us to cite sources wherever appropriate for better visibility.

I gather this is coming from CoreFx, and so we should use `System.Linq` for the namespace to be consistent with upstream.

---

*[Ix.NET/Source/System.Interactive.Async/Strings.cs, line 10 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZ1-KyNIIZSfvMWqp:-KLnZ1-KyNIIZSfvMWqq:-1161337394) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Strings.cs#L10)):*
> ```Smalltalk
>         public static string NO_ELEMENTS = "Source sequence doesn't contain any elements.";
>         public static string MORE_THAN_ONE_ELEMENT = "Source sequence contains more than one element.";
>         public static string NOT_SUPPORTED = "NOT SUPPORTED";
> ```

Could we make this more user-friendly? cc @onovtny for feels

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLnZAMR6zrWr-qbmyjx:548409269) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
> ```Smalltalk
>         }
> 
>         public static void Handle<T, R>(this Task<T> task, TaskCompletionSource<R> tcs, Action<T> success)
> ```

I'm not sure if external callers would have relied on this code (let alone should have), but given it is `public static` I'm hesitant to remove this right now.  Feels @onovotny @bartdesmet?

---

*[Ix.NET/Source/Tests/AsyncTests.Aggregates.cs, line 21 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLndaxz--LNA_SHspaQ:-KLndaxz--LNA_SHspaR:-346314461) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.Aggregates.cs#L21)):*
> ```Smalltalk
>     public partial class AsyncTests
>     {
>         private const int WaitTimeoutMs = 5000;
> ```

I'm not averse to putting timeouts here in the tests, especially where `.Wait()` is used, but I'd love to understand more about what you found while running the tests that necessitated this change...

---

*[Ix.NET/Source/Tests/AsyncTests.cs, line 24 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnf0pEHbOJyym83obV:-KLnf0pEHbOJyym83obW:-915851297) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.cs#L24)):*
> ```Smalltalk
> 
>         [Obsolete("Don't use this, use Assert.ThrowsAsync and await it", true)]
>         public Task AssertThrows<E>(Func<Task> func)
> ```

I don't see any usages of this. Shall we clean this up?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
@aelij thanks for contributing this! Don't worry about the volume of comments I've added, that's just me trying to be thorough and ensure nothing slips through the review process.

For the most part, the stuff I've raised is just cleanup jobs (prefixed with a :lipstick: emoji) but there's a few questions I have as well.

onovotny:
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLnhfMzm8oVh6ukg3cq:-628564895) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> > to ignore it for now here
> 
> So we should leave this as-is or undo this change?
> 
</details>
Was this new or in there before, that's what I'm confused about. If it's new, then probably nuke it, otherwise, leave it as it's harmless and we'll clean up ifdef's later.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1299 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnXJjgjb_MYAdUJRO2:-KLnhyxy1DjFO2QIuL6w:-284567654) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1299)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> This is an introduced `checked` here - is this just to be consistent or did you encounter something unexpected here during the migration?
</details>
Seems like this is logically the correct behavior though -- if we get past `int.MaxValue` it should throw, shouldn't it?

---

*[Ix.NET/Source/System.Interactive.Async/Grouping.cs, line 9 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZpWWHciRyDKLLjsa:-KLnhUgYqc7Bj30_WleT:1065916734) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Grouping.cs#L9)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> Actually, this looks really familiar - if we're pulling in this from corefx can we link back and attribute accordingly?
> 
> Something like this:
> 
> ```csharp
> /// adapted from System.Linq.Grouping from .NET Framework
> /// source: https://github.com/dotnet/corefx/blob/b90532bc97b07234a7d18073819d019645285f1c/src/System.Linq/src/System/Linq/Grouping.cs#L64
> ```
> 
> I've just used the latest commit on `master` as the reference here, you might have used a different source.
</details>
Yup, makes sense.

---

*[Ix.NET/Source/System.Interactive.Async/Lookup.cs, line 9 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLn_g0ZgbQBW7OSQx_s:-KLnh4STuYLDFvoXbUnq:1339893389) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Lookup.cs#L9)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> Same thing here - I'd like us to cite sources wherever appropriate for better visibility.
> 
> I gather this is coming from CoreFx, and so we should use `System.Linq` for the namespace to be consistent with upstream.
</details>
Agreed. I took this from CoreFX, but a link would be good here. In this case, we cannot use `System.Linq` as the namespace as `Lookup<T,U>` is public there and it'd conflict. They don't expose the public ctor or methods we need to use, so I adapted their impl but made it internal. It's only returned to callers as an `ILookup<T,U>` so it's functionally the same.

---

*[Ix.NET/Source/System.Interactive.Async/Strings.cs, line 10 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZ1-KyNIIZSfvMWqp:-KLngyHHphB_r5RDMYl5:-1006975100) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Strings.cs#L10)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> Could we make this more user-friendly? cc @onovtny for feels
</details>
Up for any suggestions on this. Also will be easier to make into a ResX once back on MSBuild in the fall.

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLngbPi7wLXTPlRP61T:1690898392) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> I'm not sure if external callers would have relied on this code (let alone should have), but given it is `public static` I'm hesitant to remove this right now.  Feels @onovotny @bartdesmet?
</details>
I would recommend leaving it in with an `Obsolete` attribute added as a warning.

---

*[Ix.NET/Source/Tests/AsyncTests.cs, line 24 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnf0pEHbOJyym83obV:-KLngA--PlByN96ZGgrh:1265246387) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.cs#L24)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> I don't see any usages of this. Shall we clean this up?
</details>
I left it there as a preventative measure as there's still the non-task overload that would work as an async void. This is there to catch accidental usages and fail.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
Reviewed 6 of 13 files at r1, 9 of 15 files at r2, 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLniwr2jGIwcdsg7b0A:928778788) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Was this new or in there before, that's what I'm confused about. If it's new, then probably nuke it, otherwise, leave it as it's harmless and we'll clean up ifdef's later.
</details>
It was introduced in 6180f069edda1a50a65a03aa25d1c5a70fe2eda9

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1299 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnXJjgjb_MYAdUJRO2:-KLnj0ElQfSDrSNTqcBa:1411571371) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1299)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Seems like this is logically the correct behavior though -- if we get past `int.MaxValue` it should throw, shouldn't it?
</details>
Yeah, I'm fine with playing it safe here. No further context on this beyond when the code was OSSed in the first commit.

---

*[Ix.NET/Source/System.Interactive.Async/Strings.cs, line 10 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZ1-KyNIIZSfvMWqp:-KLnlVquSzjW2MBRPXIZ:-769267411) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Strings.cs#L10)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Up for any suggestions on this. Also will be easier to make into a ResX once back on MSBuild in the fall.
</details>
What about `"Specified method is not supported."` (taken from the default for the .NET Framework) for the message?

---

*[Ix.NET/Source/Tests/AsyncTests.cs, line 24 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnf0pEHbOJyym83obV:-KLnks9jnIpEkKZ5-hcz:304072805) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.cs#L24)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> I left it there as a preventative measure as there's still the non-task overload that would work as an async void. This is there to catch accidental usages and fail.
</details>
:thumbsup: let's leave it in

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

aelij:
@shiftkey I don't mind, although I would've done this in a separate PR (most of the 'lipsticks' existed in the original code).

Regarding the 'var' changes, would anyone mind if I added a ReSharper rule (in a committed .DotSettings file)?

---

Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 1623 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLicAsM_f6USmkziBqq:-KLoJ5l2d39PxNrV9x_J:-514456655) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L1623)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> :lipstick: maybe inline this method as there are no other overloads?
</details>
The purpose of separating each operator to private methods is to preserve the semantics of the previous version - throw immediately and not asynchronously.  If you take a look at the LINQ implementation (that uses compiler-generated iterators), you'll see the same thing.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLoKyOV7YeRSaza6qM4:-1597184644) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> We don't seem to respect the provided `CancellationToken` inside the provided `moveNext` - do you think we should?
</details>
Sure, we could add `ct.ThrowIfCancellationRequested()`. I added this method because I needed a way to transform an enumerable to an async enumerable without creating thread-pool tasks.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1120 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnVSM2oBbUfuoa2shJ:-KLoMTHhXoVqOURhbDta:-1316211658) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1120)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> I *think* this is fine to leave here, as the old code using `TaskCompletionSource` meant that `TrySetException` wasn't going to throw the exception (but `.Throw()` here should).
</details>
We can also do what Roslyn does in such cases `throw ExceptionUtils.Unreachable` (which throws an exception saying "This program location is thought to be unreachable")

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1232 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnW-7NjC5FlNBJs6Hf:-KLoNHexj4On8YuDsq_g:-810135685) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1232)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> Can we document why this is an important scenario to catch and rethrow explicitly, rather than fall through to the `onError`?
</details>
I think it's self-evident. Cancellation has special handling throughout the code.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1299 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnXJjgjb_MYAdUJRO2:-KLoNw-4_qa2KyzxDWZp:-1098065945) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1299)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> Yeah, I'm fine with playing it safe here. No further context on this beyond when the code was OSSed in the first commit.
</details>
I added it because it exists in all other operators.

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLoOYTfDD3aNmb-Q9gS:885251030) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> I would recommend leaving it in with an `Obsolete` attribute added as a warning.
</details>
This entire class is internal. We can definitely remove unused methods.

---

*[Ix.NET/Source/Tests/AsyncTests.Aggregates.cs, line 21 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLndaxz--LNA_SHspaQ:-KLoOearoiC9-QMUBxKS:1289933613) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.Aggregates.cs#L21)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> I'm not averse to putting timeouts here in the tests, especially where `.Wait()` is used, but I'd love to understand more about what you found while running the tests that necessitated this change...
</details>
When I transitioned the code to use await, it became apparent that Dispose() played a significant role here, and there were some cases where it didn't get called properly (bugs I introduced and fixed), but it made the tests hang, which was annoying.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-:-KLoPHAHS-6ASDzFK6mR:-223564862)*

aelij:
Review status: all files reviewed at latest revision, 32 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 900 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnUynefIkiDOitbsDO:-KLoViW9g91fRT4AxZs4:-2086892314) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L900)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> :lipstick: do we need the `== true` bit here?
</details>
It's a nullable bool

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
@aelij :thumbsup:

---

Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLoaC8zYuGoGEcp8vwS:958750068) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> Sure, we could add `ct.ThrowIfCancellationRequested()`. I added this method because I needed a way to transform an enumerable to an async enumerable without creating thread-pool tasks.
</details>
@onovotny thoughts? or just skip it?

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 900 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnUynefIkiDOitbsDO:-KLoaKIBmzgxqn3ZLgiA:1424244679) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L900)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> It's a nullable bool
</details>
Oops, my bad. :thumbsup:

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1120 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnVSM2oBbUfuoa2shJ:-KLoa_WDxIDnwk2hM37K:-1359950602) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1120)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> We can also do what Roslyn does in such cases `throw ExceptionUtils.Unreachable` (which throws an exception saying "This program location is thought to be unreachable")
</details>
@onovotny thoughts?

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLob1klCemrwOxLkC8Q:174732578) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> This entire class is internal. We can definitely remove unused methods.
</details>
@aelij good point!

---

*[Ix.NET/Source/Tests/AsyncTests.Aggregates.cs, line 21 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLndaxz--LNA_SHspaQ:-KLob6CsQfx_-bvTA-LN:1788762346) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.Aggregates.cs#L21)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> When I transitioned the code to use await, it became apparent that Dispose() played a significant role here, and there were some cases where it didn't get called properly (bugs I introduced and fixed), but it made the tests hang, which was annoying.
</details>
@onovotny are we cool with this, at least until we get around to updating the tests to behave well with xUnit 2's infrastructure?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-:-KLoa2zKgl83RWZQ9Jn5:1981514234)*

onovotny:
There is already a solution level settings file for each project but that particular rule may not be set to always switch/use `var`. Updating that solution level setting for each is :+1: for me.

---

Review status: all files reviewed at latest revision, 29 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs, line 1623 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLicAsM_f6USmkziBqq:-KLpsd1lAgkJRZObV0Yo:1052413458) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Aggregates.cs#L1623)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> The purpose of separating each operator to private methods is to preserve the semantics of the previous version - throw immediately and not asynchronously.  If you take a look at the LINQ implementation (that uses compiler-generated iterators), you'll see the same thing.
</details>
Makes sense. This will get cleaner with C# 7 with inline functions as it's that exact behavior that's one of the goals of doing it.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLpsmFs5Ka_6vZXq3Aq:-590272643) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> @onovotny thoughts? or just skip it?
</details>
I tend to think that `CancellationToken`'s should be honored if present.  Seems like calling `ThrowIfCancellationRequested` is a good idea.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 102 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLif7CVZNibu4D5D12h:-KLptRX6HlzpUtgF-mSt:340390953) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L102)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> It was introduced in 6180f069edda1a50a65a03aa25d1c5a70fe2eda9
</details>
got it. I'd say leave it as-is. That ifdef was for platforms that didn't have `IObserver`/`IObservable` in the BCL (Silverlight). Kind of a clean-as-we-go approach, though I can do a "big bang" cleanup too.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1120 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnVSM2oBbUfuoa2shJ:-KLpvPecsuxCC8gurBBO:-229735301) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1120)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> @onovotny thoughts?
</details>
Not sure I have a strong opinion here.  If it's the same situation as Roslyn, we can follow their lead.

---

*[Ix.NET/Source/System.Interactive.Async/Strings.cs, line 10 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZ1-KyNIIZSfvMWqp:-KLpvhcrMxjistvh0Q38:-1804098361) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/Strings.cs#L10)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> What about `"Specified method is not supported."` (taken from the default for the .NET Framework) for the message?
</details>
Fair enough, works for me :)

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLpvlbLkoYJ1pjVf5DF:810082289) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> @aelij good point!
</details>
:fire: 

---

*[Ix.NET/Source/Tests/AsyncTests.Aggregates.cs, line 21 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLndaxz--LNA_SHspaQ:-KLpvowditW2fCFHPhzg:98232464) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/Tests/AsyncTests.Aggregates.cs#L21)):*
<details><summary><i>Previously, shiftkey (Brendan Forster) wrote…</i></summary>
> @onovotny are we cool with this, at least until we get around to updating the tests to behave well with xUnit 2's infrastructure?
</details>
I'm definitely in favor of non-hanging tests, even if they're just exposing bugs we need to fix. Figuring out which test(s) hang is a time sink

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-:-KLpsGWFHSfg-e4vNiEC:-1220885737)*

aelij:
Review status: all files reviewed at latest revision, 29 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLq0XlLHgK2NseGQE6p:-1312991594) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> I tend to think that `CancellationToken`'s should be honored if present.  Seems like calling `ThrowIfCancellationRequested` is a good idea.
</details>
Following the discussion in #202 perhaps I should remove this method and add it in a different PR?

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 35 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnZAMR6zrWr-qbmyjw:-KLq0y0ZbMh_qA7L12S2:922598708) ([raw file](https://github.com/reactive-extensions/rx.net/blob/8374412c38dcc866c5fa8c79283f33f7de0b584f/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L35)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> :fire: 
</details>
I've removed the Finally method (there was one use but I could convert it to a try..finally block).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

onovotny:
Review status: all files reviewed at latest revision, 28 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs, line 23 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLidNlDupPKoNoLfdzF:-KLqLzWIpaPSXJs8Kv_-:1258253021) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Conversions.cs#L23)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> Following the discussion in #202 perhaps I should remove this method and add it in a different PR?
</details>
Yeah, makes sense -- let's track this separately. I don't like the current behavior, but I'd like to get the rest of this in while we figure out the right answer to the other this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

onovotny:
Reviewed 10 of 10 files at r4.
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 80 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuwguNG2ErpN0bYVJn:-KLuwguNG2ErpN0bYVJo:-1145985622) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80)):*
> ```Smalltalk
>                     using (ct.Register(stop))
>                     {
>                         return await moveNext(ct, tcs);
> ```

.ConfigureAwait(false)?

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 13 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuxGfSdiEEd8429YBF:-KLuxGfSdiEEd8429YBG:2097266861) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L13)):*
> ```Smalltalk
>         public static readonly Task<bool> False = Task.FromResult(false);
>         
>         public static Task<T> Throw<T>(Exception exception)
> ```

Where is this still being used, and can we get rid of it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

aelij:
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 80 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuwguNG2ErpN0bYVJn:-KLuxlUCoTTzTNIkAaOk:-154063849) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> .ConfigureAwait(false)?
</details>
Sigh. Serves me right for not supporting the ReSharper EAP for my **own** extension (ConfigureAwaitChecker). :)

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 13 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuxGfSdiEEd8429YBF:-KLuy6TIXylH7DqZV91i:-50980382) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L13)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Where is this still being used, and can we get rid of it?
</details>
I looked into getting rid of it, but it's used in a few legitimate places (e.g. the `Throw` operator). This is a useful method - even .NET 4.6 added a `Task.FromException()`, but it's unavailable in the selected netstandard.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs, line 1120 \[r3\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLnVSM2oBbUfuoa2shJ:-KLuyHd_d6eMhwNBFKYb:2105501798) ([raw file](https://github.com/reactive-extensions/rx.net/blob/562316ad93b26ba38efc3e6205919c15191ca02c/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs#L1120)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Not sure I have a strong opinion here.  If it's the same situation as Roslyn, we can follow their lead.
</details>
This interests me, but not really strongly enough to address it here. Let's leave this as-is.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

onovotny:
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 80 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuwguNG2ErpN0bYVJn:-KLv1JdrsSFibuAORH4n:-93655082) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> Sigh. Serves me right for not supporting the ReSharper EAP for my **own** extension (ConfigureAwaitChecker). :)
</details>
Will happily install that extension once available for the EAP. I do wish ReSharper's EAP's didn't constantly break add-ins.

---

*[Ix.NET/Source/System.Interactive.Async/TaskExt.cs, line 13 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuxGfSdiEEd8429YBF:-KLv1W8Dj-qk0MQcASuG:-510466975) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/TaskExt.cs#L13)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> I looked into getting rid of it, but it's used in a few legitimate places (e.g. the `Throw` operator). This is a useful method - even .NET 4.6 added a `Task.FromException()`, but it's unavailable in the selected netstandard.
</details>
:+1: 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

aelij:
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 80 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuwguNG2ErpN0bYVJn:-KLv1yxiRUNyUyEonCxn:1483328371) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80)):*
<details><summary><i>Previously, onovotny (Oren Novotny) wrote…</i></summary>
> Will happily install that extension once available for the EAP. I do wish ReSharper's EAP's didn't constantly break add-ins.
</details>
Hmm, how about installing the [Microsoft.ApiDesignGuidelines.Analyzers](https://www.nuget.org/packages/Microsoft.ApiDesignGuidelines.Analyzers/1.2.0-beta2) pack for this project? It has several useful diagnostics, including `CA2007` - Do not directly await a Task. This way we're not R#-dependent, and the warnings would show up in the CI build.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

onovotny:
Review status: all files reviewed at latest revision, 30 unresolved discussions.

---

*[Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs, line 80 \[r4\]](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170#-KLuwguNG2ErpN0bYVJn:-KLv4bi-p0b0Clsehp2a:-1402257907) ([raw file](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80)):*
<details><summary><i>Previously, aelij (Eli Arbel) wrote…</i></summary>
> Hmm, how about installing the [Microsoft.ApiDesignGuidelines.Analyzers](https://www.nuget.org/packages/Microsoft.ApiDesignGuidelines.Analyzers/1.2.0-beta2) pack for this project? It has several useful diagnostics, including `CA2007` - Do not directly await a Task. This way we're not R#-dependent, and the warnings would show up in the CI build.
</details>
Will take a look at that, thanks for the tip!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/reactive-extensions/rx.net/170)*

shiftkey:
:shipit:

onovotny:
@aelij Thanks for your hard work here!

aelij:
🎉 

@onovotny Please fix the ConfigureAwait() [here](https://github.com/reactive-extensions/rx.net/blob/62b7ef4fb6a7df4f592d0aacbbeab97185e2a69a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs#L80). I was waiting for all the discussions to be resolved, and didn't realize you were going to merge it :)

Thanks!

onovotny:
@aelij done

aelij:
Thanks guys. It's been fun.

Now it's time to start thinking about #194 :)

PR #171, opened 2016-04-16
Fix Ix Catch method invalid behaviour with Array.

RyotaMurohoshi:
I fixed Ix Catch methods with Array.

At 2252cb4, next test passes.

```csharp
[TestMethod]
public void Catch5()
{
    var xss = new[] { Enumerable.Range(0, 5), Enumerable.Range(5, 5) };
    var res = EnumerableEx.Catch(xss);
    Assert.IsTrue(res.SequenceEqual(Enumerable.Range(0, 5)));
}
```

But next test failed.

```csharp
[TestMethod]
public void Catch5_Array()
{
    var xss = new[] { new[] { 0, 1, 2, 3, 4 }, new[] { 5, 6, 7, 8, 9 } };
    var res = EnumerableEx.Catch(xss);
    Assert.IsTrue(res.SequenceEqual(Enumerable.Range(0, 5)));
}
```

`IEnumerator<T>` `Current` property when `IEnumerator<T>` `MoveNext()` return false is undefined.

Array throws Exception.`List<T>` returns defaualt(T).

Unexpected throwing exceptions are happen at [here](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs#L44) and [here](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs#L129) with Array.

I fixed it.

### DIFF
--- a/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
+++ b/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
@@ -37,13 +37,11 @@ namespace System.Linq
             {
                 while (true)
                 {
-                    var b = default(bool);
                     var c = default(TSource);
 
                     try
                     {
-                        b = e.MoveNext();
-                        if (!b)
+                        if (!e.MoveNext())
                             break;
 
                         c = e.Current;
@@ -122,13 +120,11 @@ namespace System.Linq
 
                     while (true)
                     {
-                        var b = default(bool);
                         var c = default(TSource);
 
                         try
                         {
-                            b = e.MoveNext();
-                            if (!b)
+                            if (!e.MoveNext())
                                 break;
 
                             c = e.Current;
--- a/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
+++ b/Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
@@ -43,6 +43,9 @@ namespace System.Linq
                     try
                     {
                         b = e.MoveNext();
+                        if (!b)
+                            break;
+
                         c = e.Current;
                     }
                     catch (TException ex)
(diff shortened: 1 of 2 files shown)


mattpodwysocki:
@RyotaMurohoshi please rebase this before we can merge.  Thanks for the catch!

Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
```diff
@@ -126,6 +126,9 @@ private static IEnumerable<TSource> Catch_<TSource>(this IEnumerable<IEnumerable
                         try
                         {
                             b = e.MoveNext();
+                            if (!b)
```
bartdesmet: Remove local variable `b` now it doesn't have to span across the `try` statement anymore.

Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
```diff
@@ -41,6 +41,9 @@ public static partial class EnumerableEx
                     try
                     {
                         b = e.MoveNext();
+                        if (!b)
```
bartdesmet: Remove local variable `b` now it doesn't have to span across the `try` statement anymore.

Ix.NET/Source/System.Interactive/EnumerableEx.Exceptions.cs
RyotaMurohoshi: @bartdesmet 

Thanks!
OK, I fixed at e858877 .

RyotaMurohoshi:
@mattpodwysocki 

I `rebased` and `pushed -f`! 

OK?

mattpodwysocki:
LGTM, @bartdesmet?

PR #175, opened 2016-04-27
Weak Observable (ISSUE: WeakSubscribe #134)

bnayae:
Weak subscription implementation + tests +
Addition overload for AdvanceBy of TestScheduler


shiftkey:
@bnayae there's currently a merge conflict here, as well as a CLA that needs to be signed, but I'm also not sure of the backstory behind this change - could you tell me a bit about the motivation for it?

shiftkey:
@bnayae there's currently a merge conflict here, as well as a CLA that needs to be signed - are you still interested in wrapping this up?

bnayae:
I had sign the .NET Core CLA is it the same CLA?
if not where can I sign it?

the motivation behind the feature is to be able to subscribe consumer which will not kept alive
because of the subscription.
 
it can be UI view that may be close any class what has it's own lifetime management.
if the lifetime managed component is doing the subscription it can easily dispose it when it disposed,
but if it expose observer it may not be aware of the subscription.

akarnokd:
I'm not sure what the intended use case is. The general problem I see is that when a weak observer is gone, nothing will dispose the sequence and thus it may leak resources. I'd say Rx.NET project is not the best target for such very special feature.

danielcweber:
>The general problem I see is that when a weak observer is gone, nothing will dispose the sequence and thus it may leak resources.

The PR might be fixed to dispose the subscription is the observer is gone. @bnayae Can you elaborate on the use case for this? I could imagine both: that there are use cases as well as this operator being much too special and dangerous for this library.

danielcweber:
Closing due to unclear use case, improbability to be considered as a feature and the general risk to make it a operator in the basic library.

PR #178, opened 2016-05-04
Refactor Testing API for better clearability and ease of use.

bnayae:
The TestScheduler.Start didn't reflect its default via the use of overloads,
using default parameters reflect the default value on the intellisense.

At ReactiveTest
OnNext, OnError and OnComplete
is commonlly use to reflect long period and it more convinent to have TimeSpan overload


Rx.NET/Source/Microsoft.Reactive.Testing/ReactiveTest.cs
```diff
@@ -93,6 +116,22 @@ public static Recorded<Notification<T>> OnError<T>(long ticks, Exception excepti
...
+        /// <param name="duration">Recorded virtual time the OnError notification occurs.</param>
+        /// <param name="exception">Recorded exception stored in the OnError notification.</param>
+        /// <returns>Recorded OnError notification.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="exception"/> is null.</exception>
+        public static Recorded<Notification<T>> OnError<T>(TimeSpan duration, Exception exception)
+        {
+            if (exception == null)
+                throw new ArgumentNullException("exception");
+
+            return new Recorded<Notification<T>>(duration.Ticks, Notification.CreateOnError<T>(exception));
```
paulpdaniels: Why not just pass this to the overload?

```csharp
return OnError(duration.Ticks, Notification.CreateOnError<T>(exception));

```

LeeCampbell:
I see what you are trying to do. I am also not a big fan of the longs-as-ticks which everyone uses Timespan and the converts back to Ticks.

I would also love to see the default magic values removed from the `Start` operator

martinwoodward:
Hi @bnayae  - would you mind signing the [.NET Foundation Contribution License agreement](http://cla2.dotnetfoundation.org)? Shouldn't take long and is hopefully pretty painless. Many thanks!

bnayae:
I had sign the License

Rx.NET/Source/Microsoft.Reactive.Testing/TestScheduler.cs
```diff
@@ -88,38 +92,6 @@ public ITestableObserver<T> Start<T>(Func<IObservable<T>> create, long created,
             return observer;
         }
 
-        /// <summary>
```
akarnokd: Would the removal of these methods count as a breaking change, even if another overload now features default parameter values?

vgw-LeeCampbell: I believe so as any compiled code would not be able to just redirect to these other methods as they are different signatures. From my understanding you would need a perform a recompile of the consuming classes (which would put the default values in at compile time) , and thus this is a breaking change.

danielcweber:
Are this PR's improvements enough to go for a breaking change? @bnayae Can we have the improvements without breaking changes? There are merge conflicts anyway.

bnayae:
I update the branch with the current master version.
I think that if this change will be part of a new NuGet version, It will make no problem to any consumer because the consumer will have to build his code anyway against the new NuGet.
existing signature should have kept working.

akarnokd:
Well, I'm not convinced we'd need those new record creation methods. Usually, I prefer API changes that either the library will use then on, is by itself is non-trivial to implement or requires deep changes to support some 3rd party behavior. None of these is seems to apply here.

danielcweber:
Besides, for testing purposes, the unit of time here is completely irrelevant because we're dealing with virtual time only...am I correct ?

danielcweber:
Closing due to a breaking change we won't go for at this moment. Also, the unit of time is irrelevant for virtual time schedulers.

Issue #183, opened 2016-06-08
IX-Async implementation of ForEachAsync possibly failing to dipose enumerator


divega:
We got an EF Core 1.0.0 RC2 bug report at https://github.com/aspnet/EntityFramework/issues/5440 that could be explained by `ForEachAsync()` not timely calling `Dispose()` on the `IAsyncEnumerator<T>`.

We didn't get a full repro but the customers confirmed that the issue went away when we replaced IX-Async's implementation of `ForEachAsync()` with our own.

cc @anpete

onovotny:
Is there any chance you can submit your version as a PR?

onovotny:
This should be fixed in #170. Can you please verify with 3.0.1-build00046 or later?

PR #185, opened 2016-06-16
Fixes for EF issues 

onovotny:
This PR fixes 

https://github.com/Reactive-Extensions/Rx.NET/issues/182
and 
https://github.com/Reactive-Extensions/Rx.NET/issues/183

Note that the fixes for these involve incorporating `await`. This means we need the `Microsoft.Bcl.Async` library to support .NET 4.0....or we drop .NET 4.0 support. Dropping .NET 4 means we can avoid any potential issues with the BCL Async tools.

/cc @bartdesmet @shiftkey @natemcmaster @mattpodwysocki

---
This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/185)

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Multiple.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Multiple.cs
@@ -356,87 +356,8 @@ namespace System.Linq
             if (comparer == null)
                 throw new ArgumentNullException("comparer");
 
-            return Create(() =>
-            {
-                var innerMap = default(Task<ILookup<TKey, TInner>>);
-                var getInnerMap = new Func<CancellationToken, Task<ILookup<TKey, TInner>>>(ct =>
-                {
-                    if (innerMap == null)
-                        innerMap = inner.ToLookup(innerKeySelector, comparer, ct);
 
-                    return innerMap;
-                });
-
-                var outerE = outer.GetEnumerator();
-                var current = default(TResult);
-
-                var cts = new CancellationTokenDisposable();
-                var d = Disposable.Create(cts, outerE);
-
-                var f = default(Action<TaskCompletionSource<bool>, CancellationToken>);
-                f = (tcs, ct) =>
-                {
-                    getInnerMap(ct).Then(ti =>
-                    {
-                        ti.Handle(tcs, map =>
-                        {
-                            outerE.MoveNext(ct).Then(to =>
-                            {
-                                to.Handle(tcs, res =>
-                                {
-                                    if (res)
-                                    {
-                                        var element = outerE.Current;
-                                        var key = default(TKey);
-
-                                        try
-                                        {
-                                            key = outerKeySelector(element);
-                                        }
-                                        catch (Exception ex)
-                                        {
-                                            tcs.TrySetException(ex);
-                                            return;
-                                        }
-
-                                        var innerE = default(IAsyncEnumerable<TInner>);
-                                        if (!map.Contains(key))
-                                            innerE = AsyncEnumerable.Empty<TInner>();
-                                        else
-                                            innerE = map[key].ToAsyncEnumerable();
-
-                                        try
-                                        {
-                                            current = resultSelector(element, innerE);
-                                        }
-                                        catch (Exception ex)
-                                        {
-                                            tcs.TrySetException(ex);
-                                            return;
-                                        }
-
-                                        tcs.TrySetResult(true);
-                                    }
-                                    else
-                                    {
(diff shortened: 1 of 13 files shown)


onovotny:
Note that this is 3.0.1 as per request of the EF team to keep this out of 3.0.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Multiple.cs
```diff
@@ -453,6 +374,169 @@ public static Task<bool> SequenceEqual<TSource>(this IAsyncEnumerable<TSource> f
...
+                    List<TInner> group;
+
+                    if (!await _outer.MoveNext(cancellationToken).ConfigureAwait(false))
+                    {
+                        return false;
+                    }
+
+                    if (_innerGroups == null)
+                    {
+                        _innerGroups = new Dictionary<TKey, List<TInner>>();
```
bartdesmet: Use comparer?

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Multiple.cs
```diff
@@ -453,6 +374,169 @@ public static Task<bool> SequenceEqual<TSource>(this IAsyncEnumerable<TSource> f
...
+
+                        while (await _inner.MoveNext(cancellationToken).ConfigureAwait(false))
+                        {
+                            var inner = _inner.Current;
+                            var innerKey = _innerKeySelector(inner);
+
+                            if (innerKey != null)
+                            {
+                                if (!_innerGroups.TryGetValue(innerKey, out group))
```
bartdesmet: Can the key be null? Dictionary doesn't support that.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
```diff
@@ -1652,49 +1652,22 @@ public static void ForEach<TSource>(this IAsyncEnumerable<TSource> source, Actio
...
-
-            var i = 0;
-
-            var f = default(Action<CancellationToken>);
-            f = ct =>
+            using (var asyncEnumerator = source.AsAsyncEnumerable().GetEnumerator())
             {
-                e.MoveNext(ct).Then(t =>
+                var i = 0;
+                while (await asyncEnumerator.MoveNext(cancellationToken))
```
bartdesmet: ConfigureAwait(false)

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
```diff
@@ -1652,49 +1652,22 @@ public static void ForEach<TSource>(this IAsyncEnumerable<TSource> source, Actio
             source.ForEachAsync(action, cancellationToken).Wait(cancellationToken);
         }
 
-        public static Task ForEachAsync<TSource>(this IAsyncEnumerable<TSource> source, Action<TSource, int> action, CancellationToken cancellationToken)
+        public static async Task ForEachAsync<TSource>(this IAsyncEnumerable<TSource> source, Action<TSource, int> action, CancellationToken cancellationToken)
```
bartdesmet: What's the bug being addressed here?

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
onovotny: In this case, is that ok or not as it'll call the `action()` on the next line? That could mean the user's callback is happening on a different thread. Probably ok, just confirming.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Single.cs
onovotny: The bug is #183. There's a repro using EF RC2 that we need to extract an Ix-only repro for. https://github.com/aspnet/EntityFramework/issues/5440

Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Multiple.cs
onovotny: I completely changed the impl to something simpler based on https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/GroupJoin.cs. It uses Lookup and now has the same behavior as CoreFX linq.

onovotny:
I'm going to expand this PR to include #200 as the EF specific fixes are a subset of that.

onovotny:
closing this in favor of #170

PR #208, opened 2016-06-30
VS environment cleanup

shiftkey:
After installing VS2015 Update 3 and the latest stable VS tooling for .NET Core, you'll probably see this message when opening the Rx.NET solution:

screen shot 2016-06-30 at 5 03 57 pm

This PR makes the error go away, and also adds some of the solution-level changes to the Ix.NET solution so these have a consistent structure.

- [x] don't lock SDK versions

---
This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/208)

### DIFF
--- a/Rx.NET/Source/global.json
+++ b/Rx.NET/Source/global.json
@@ -1,6 +1,3 @@
 {
-    "sources": [ ".", "Tests.System.Reactive" ],
-    "sdk": {
-      "version": "1.0.0-preview2-003093"
-  }
+  "projects": [ ".", "Tests.System.Reactive" ]
 }
--- a/Ix.NET/Source/Ix.NET.sln
+++ b/Ix.NET/Source/Ix.NET.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25402.0
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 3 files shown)


Rx.NET/Source/global.json
```diff
@@ -1,6 +1,6 @@
 {
-    "sources": [ ".", "Tests.System.Reactive" ],
-    "sdk": {
-      "version": "1.0.0-preview2-003093"
+  "sources": [ ".", "Tests.System.Reactive" ],
+  "sdk": {
+    "version": "1.0.0-preview2-003121"
```
shiftkey: Is there any benefit to locking these? I'm not sure of the tradeoffs here, and I already have accrued four different versions under `C:/Program Files/dotnet/` without thinking...

onovotny: From what I've read, leaving the sdk node out means it'll use the latest version installed, so it everything floats.

shiftkey: Cool, I'll pull these out and see what happens in VS

Rx.NET/Source/global.json
```diff
@@ -1,6 +1,6 @@
 {
-    "sources": [ ".", "Tests.System.Reactive" ],
-    "sdk": {
-      "version": "1.0.0-preview2-003093"
+  "sources": [ ".", "Tests.System.Reactive" ],
```
onovotny: `sources` should also be renamed to `projects` while you're in there.

shiftkey:
:eyes:

onovotny:
:shipit:

Issue #97, opened 2015-05-18
NET 4.0 and .NET 4.5. versions need to be signed with different keys


forki:
As discussed in https://github.com/fsprojects/Paket.VisualStudio/pull/38#issuecomment-103008730 we have a very weird issue with using Rx in a Visual Studio extension. It seems the Xamarin addin uses also Rx and this breaks [Paket.VisualStudio](https://github.com/fsprojects/Paket.VisualStudio) and AFAIK also "GitHub for Visual Studio 2015" (which ships with VS).

Quoting @shana from the other issue:

      Microsoft is shipping two separate builds of system.reactive, one for .NET 4.0 and another for .NET 4.5. Both with the same signing key snd version. As a result, the runtime doesn't distinguish them (runtime version is not a part of the assembly identification) and will only load one of them, whichever comes first. If the first is the one for .net 4.0, any 4.5 methods will not be a part of it.
      
      Result: MethodNotFoundException
      
      The only way to fix this is to either load your extension before they do (if you can, it's a race), or ship your own copy of system.reactive with a different signing key so it can be loaded side-by-side with the one Xamarin is loading, which is what I'm doing. I have a fork at https://github.com/shana with a different key, you can also use the ones that ship with the GitHub Extension for VS2015.

So @shana has already "fixed it" for "GitHub for Visual Studio" and uses a custom build from https://github.com/shana/Rx.NET/tree/custom-key

/cc @shana please upload your nuget package to your releases page so that we can reuse it as a temporary fix.

onovotny:
For things like VS-addins using shared components, it seems like you'll always be subject to this issue if you use a common library. If you can't load into a separate appdomain (and the VS model may not allow it), you may be better-off using ILMerge to internalize your references. That way you'll always get the dependencies you compiled and tested against. That approach would shield you against more than just Rx version differences.

shana:
Ooops, looks like I missed this discussion 😞

It's probably not relevant that I comment on this now, but for future reference:

@forki The nuget packages are at https://github.com/github/VisualStudio/tree/master/lib.

@onovotny Two DLLs with the same name, same version and same key mean that they are ABI-compatible. Having breaking ABI changes without changing anything in the identification of the library is a critical compatibility-breaking bug. ILMerge or custom signing are workarounds, consequences to the problem, not the actual solution to the underlying problem.

onovotny:
[...]

Issue #210, opened 2016-07-01
Why does the 3.0 test library depend on XUnit.Assert?


RxDave:
I'm using MS Test in my projects and I'd rather not have any dependencies on XUnit. Is there any chance that you'll drop this dependency? Why include it in the first place given that the 2.2.5 release worked just fine without it?

onovotny:
This was done during the port to .NET Core as there was no .NET Core support for MSTest at the time. The 2.2.5 version did have a dependency on MSTest's assertions.

The assertions are an implementation detail of `ReactiveAssert`, but we can look into using a source version and making it internal. Would that work for you?

RxDave:
Thanks for the answer! I didn't realize that there was a dependency on MSTest. I'd agree that, if possible, the Rx library should be test library-agnostic.

I think that `Debug.Assert` typically works with any test framework. Would you consider using that instead?

onovotny:
`Debug.Assert` doesn't really work for a couple of reasons:

1. It's conditionally-compiled for `DEBUG`, so in our release builds of the library it'd be a no-op.
2. Even if we were to `#define DEBUG` in the file to get around that, `Debug.Assert` throws up a modal dialog. What we need is an exception that gets translated into a test failure.
3. What we really need are things like `Assert.Same`, `Assert.SequenceEqual`, etc. No good reason for us to reimplement that logic when there are existing assertion libraries that already do that.

Using a source-version of the assertions should address this though to remove the external dependency, right?

RxDave:
Point taken about conditional compilation, and actually `Debug.Assert` does show a modal dialog, which I just verified. I really thought that MSTest and NUnit intercepted it to simply fail the test (never used XUnit). Not sure why I thought that, but I guess I was wrong.

Embedding parts of the source is of course fine too, and it would prevent having to build custom assertion methods, but I was just thinking that the Rx test framework probably doesn't need any dependencies on semantic assertion methods in the first place. It defines its own semantic assertions, via `ReactiveAssert`, thus it should be able to use some kind of standard assertion mechanism internally to alert _any_ test runner that there is a failure; e.g., a simple test-framework agnostic `Assert.Fail` mechanism should technically be good enough to implement `ReactiveAssert`. Or am I missing something?

Anyway, that's perhaps a limitation in the FCL. It would be great to have a general-purpose `Test.Assert` method or something like that. Barring that, we're in agreement.

RxDave:
[...]

PR #212, opened 2016-07-03
Implement assembly version's per "band" to ensure binding redirects work

### LINKED ISSUE #205: Implement assembly version strategy
After a long discussion with @ericstj about #97, it seems like there's no really good options here, just one possibly "least bad" one.

To summarize the issue, when a package provides multiple assemblies that work on a platform the AssemblyVersion for each "higher/wider" version needs to be greater to ensure binding redirects work as expected. This means we cannot simply use the NuGet package version for all assemblies in the package. This is true whether it's `net40` and `net45` or `netstandard1.0`, `netstandard1.1` and `netstandard1.3` as we have today. The core issue is the same -- higher versions have a superset of the functionality as lower-platform versions.

In assembly versioning, it's generally considered the case that a higher assembly version contains all of the functionality of a lower one. This leads to the following "least bad" option:

- We will keep the major version of the assembly equal to the NuGet package (this is the one area we may have to reconsider)
- For each assembly in a compatibility level, we'll use a range within the assembly version. This gives us room for 1000 releases before we need to bump anything. Platform specific versions would use the range that matches its [.NET Platform Standard version](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#mapping-the-net-platform-standard-to-platforms):
 - `netstandard1.0`: `3.0.0`-`3.0.999`
 - `netstandard1.1`: `3.0.1000`-`3.0.1999`
 - `netstandard1.3`: `3.0.3000`-`3.0.3999`
 - `net45`: `3.0.0`-`3.0.999`
 - `net46`: `3.0.3000`-`3.0.3999`
 - `win8`: `3.0.1000`-`3.0.1999`
 - `wp8`: `3.0.0`-`3.0.999`
 - `win81`: `3.0.2000`-`3.0.2999`
 - `wpa81`: `3.0.2000`-`3.0.2999`
 - `uap10.0`: `3.0.4000`-`3.0.4999`
 - `netcoreapp1.0`: `3.0.6000`-`3.0.6999` 

Given this model, we would increment `Minor` and `Patch` versions on the NuGet package as appropriate for the release, but each release would increment a value in the given range for the assembly. This will ensure that something isn't accidentally redirected to a higher assembly version with fewer/missing API surface area.

One open question is `Major` versions. Given that major versions are generally considered able to have breaking changes, it may be fair to bump the major assembly version to match. This could potentially lead to an issue if a `3.0.3000` assembly gets redirected to a `4.0.0` version that has less surface area. For a major release, this might be acceptable.

Thoughts? 

/cc @bartdesmet @paulcbetts @mattpodwysocki @shiftkey @LeeCampbell @shana @forki @OmerRaviv @natemcmaster

onovotny:
Fixes #205 and #97.

/cc @shiftkey @mattpodwysocki @natemcmaster @shana @forki @omerraviv for review.

---
This change is [Reviewable](https://reviewable.io/reviews/reactive-extensions/rx.net/212)

### DIFF
--- a/Ix.NET/Source/Ix.NET.sln
+++ b/Ix.NET/Source/Ix.NET.sln
@@ -18,7 +18,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{B733D97A-F1ED-4FC3-BF8E-9AC47A89DE96}"
 	ProjectSection(SolutionItems) = preProject
 		GlobalAssemblyVersion.cs = GlobalAssemblyVersion.cs
-        build-new.ps1 = build-new.ps1
+		build-new.ps1 = build-new.ps1
 		global.json = global.json
 		NuGet.Config = NuGet.Config
 	EndProjectSection
--- a/Ix.NET/Source/Ix.NET.sln
+++ b/Ix.NET/Source/Ix.NET.sln
@@ -17,7 +17,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "System.Interactive.Tests", 
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{B733D97A-F1ED-4FC3-BF8E-9AC47A89DE96}"
 	ProjectSection(SolutionItems) = preProject
-		build-new.ps1 = build-new.ps1
+		GlobalAssemblyVersion.cs = GlobalAssemblyVersion.cs
+        build-new.ps1 = build-new.ps1
 		global.json = global.json
 		NuGet.Config = NuGet.Config
 	EndProjectSection
--- a/Rx.NET/Source/System.Reactive.Observable.Aliases/Properties/AssemblyInfo.cs
+++ b/Rx.NET/Source/System.Reactive.Observable.Aliases/Properties/AssemblyInfo.cs
@@ -4,16 +4,16 @@ using System.Resources;
 using System.Runtime.InteropServices;
 using System.Security;
 
-[assembly: AssemblyTitle("System.Interactive.Async")]
-// Notice: same description as in the .nuspec files; see Source/Interactive Extensions/Setup/NuGet
-[assembly: AssemblyDescription("Interactive Extensions Async Library used to express queries over asynchronous enumerable sequences.")]
+[assembly: AssemblyTitle("System.Reactive.Observable.Aliases")]
+// Notice: same description as in the .nuspec files; see Source/Rx/Setup/NuGet
+[assembly: AssemblyDescription("Reactive Extensions provides the aliases Map, FlatMap, and Filter.")]
 #if DEBUG
 [assembly: AssemblyConfiguration("Debug")]
 #else
 [assembly: AssemblyConfiguration("Retail")]
 #endif
 [assembly: AssemblyCompany(".NET Foundation and Contributors.")]
-[assembly: AssemblyProduct("Interactive Extensions")]
+[assembly: AssemblyProduct("Reactive Extensions")]
 [assembly: AssemblyCopyright("\x00a9 .NET Foundation and Contributors.  All rights reserved.")]
 [assembly: NeutralResourcesLanguage("en-US")]
 
@@ -32,5 +32,5 @@ using System.Security;
 //  Version numbers are automatically generated in the msbuild files based on regular expressions 
 // ===========================================================================
 
-[assembly: AssemblyFileVersion("1.2.0.0")]
-[assembly: AssemblyInformationalVersion("1.2.0.0")]
+[assembly: AssemblyFileVersion("2.2.0.0")]
+[assembly: AssemblyInformationalVersion("2.2.0.0")]
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,4 +1,4 @@
-assembly-versioning-scheme: MajorMinorPatch
+assembly-versioning-scheme: None
 mode: ContinuousDeployment
 tag-prefix: '[vV]'
 continuous-delivery-fallback-tag: ci
--- a/Ix.NET/Source/GlobalAssemblyVersion.cs
+++ b/Ix.NET/Source/GlobalAssemblyVersion.cs
@@ -1,0 +1,22 @@
+﻿
+using System.Reflection;
+
+
+#if NETSTANDARD1_0 || WP8
+[assembly: AssemblyVersion("3.0.0.0")]
+#elif NETSTANDARD1_1 || WINDOWS8 || NET45 || NETCORE45
+[assembly: AssemblyVersion("3.0.1000.0")]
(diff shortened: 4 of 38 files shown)


shiftkey:
Trying to build the project errors out here. Am I missing something?

```
Attempting to gather dependency information for package 'gitversion.commandline.3.5.5-pullrequest0921' with respect to project 'packages', targeting 'Any,Version=v0.0'
Package 'gitversion.commandline,3.5.5-pullrequest0921' is not found in the following primary source(s): 'C:\Users\shiftkey\AppData\Local\NuGet\Cache,C:\Users\shiftkey\.nuget\packages\,https://api.nuget.org/v3/index.json,C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\'. Please verify all your online package sources are available (OR) package id, version are specified correctly.
```

onovotny:
It should be using the Build Packages feed with the guid in it.... That's where I put that package while I wait for it to be officially merged and published.

shiftkey:
@onovotny I think this is related to where you execute the `build-new` scripts from.

On my machine, when I run `.\Rx.NET\Source\build-new` the location isn't changed, so it:

 - checks and downloads nuget.exe to root of the repository
 - `nuget restore ...` doesn't resolve a `nuget.config` at the root
 - it misses the additional `Build Packages` feed, so it can't find this new package

The workaround I tested was to do this:

```powershell
Push-Location ".\Rx.NET\Source\"
Write-Host "Building Rx.NET" -ForegroundColor Green
.\build-new
Pop-Location
```

There's probably a much cleaner way to do this, but I also haven't looked into why Appveyor isn't failing in the same way...

onovotny:
@shiftkey there's a nuget.config at the root here that should work https://github.com/Reactive-Extensions/Rx.NET/blob/master/NuGet.Config. Was that not getting picked up for you?

Ix.NET/Source/Ix.NET.sln
```diff
@@ -17,7 +17,8 @@ Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "System.Interactive.Tests",
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{B733D97A-F1ED-4FC3-BF8E-9AC47A89DE96}"
 	ProjectSection(SolutionItems) = preProject
-		build-new.ps1 = build-new.ps1
+		GlobalAssemblyVersion.cs = GlobalAssemblyVersion.cs
+        build-new.ps1 = build-new.ps1
```
shiftkey: :lipstick: indenting (lol)

Rx.NET/Source/System.Reactive.Observable.Aliases/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
```
shiftkey: Is there a particular reason why we need this file?

shiftkey:
> Was that not getting picked up for you?

It was not, but let me have another shot at it now...

Rx.NET/Source/System.Reactive.Observable.Aliases/Properties/AssemblyInfo.cs
onovotny: this file was missing before actually...so this assembly never had its attributes set.

Ix.NET/Source/Ix.NET.sln
onovotny: bah, I'll fix it -- was a merge conflict, so VS didn't generate that.

shiftkey:
Build script is fine now on my machine, I just can't create packages:

```
Building Packages

    Directory: C:\Users\shiftkey\Documents\GitHub\rx.net


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       11/07/2016   8:07 AM                artifacts
Attempting to build package from 'Microsoft.Reactive.Testing.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Core.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Experimental.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Interfaces.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Linq.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Observable.Aliases.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.PlatformServices.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Providers.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Runtime.Remoting.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Windows.Forms.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.Windows.Threading.nuspec'.
The special version part cannot exceed 20 characters.
Attempting to build package from 'System.Reactive.WindowsRuntime.nuspec'.
The special version part cannot exceed 20 characters.
```

Looking at the diff for the generated `AssemblyInformationVersion` entry, it seems like GitVersion is being very verbose:

```
-[assembly: AssemblyFileVersion("2.2.0.0")]
-[assembly: AssemblyInformationalVersion("2.2.0.0")]
+[assembly: AssemblyFileVersion("3.0.1.0")]
+[assembly: AssemblyInformationalVersion("3.0.1-assembly-versions.27+Branch.assembly-versions.Sha.ab35480481315a6703909afab95b4fb769b6d898")]
```

Do we care about this constraint?

shiftkey:
After a fresh build there are two changes to dependencies:

- `Rx.NET/Source/NuSpecs/Microsoft.Reactive.Testing.nuspec` bumps `xunit.assert` from v2.1.0 to 2.2.0-beta2-build3300
- `Rx.NET/Source/NuSpecs/System.Reactive.Observable.Aliases.nuspec` adds `System.Resources.ResourceManager` v4.0.1

shiftkey:
@onovotny just a few little things - would love to be able to poke the packages and verify the changes but I won't hold that up because of how GitVersion names branches in this situation...

onovotny:
Yeah, I wouldn't worry about any package failures due to branch names. I wish NuGet didn't have that limit; otherwise, I can also look into ways to truncate to fit. Prob another PR.

onovotny:
As for versions of BCL refs, I'm going to create a separate PR to remove those granular deps and just reference NETStandard.Library. That's the new/current recommendation of the NuGet/BCL team.

shiftkey:
@onovotny :thumbsup:

onovotny:
As for the `xunit.assert` version, #210 will turn that into a source reference, internal, so that dependency will go too.

PR #219, opened 2016-07-11
Remove granular dependencies in the NuSpecs to fix #217

### LINKED ISSUE #217: Remove granular package dependencies
As per the current guidance from @yishaigalatzer, the NuGet team and the BCL team, .NET Standard packages should have a dependency on the `NETStandard.Library` version, not the individual packages.

onovotny:
Fixes #217 
/cc @shiftkey @natemcmaster


natemcmaster:
:eyes: this may have implications for EF Core as it still uses granular dependencies. I'll discuss internally with EF/aspnet team, but don't feel you have to wait to merge due to this.

shiftkey:
@onovotny is there a rush on getting this in for 3.0.1?

onovotny:
I guess there's no real rush but it's the correct thing to do according to the team's I've spoken to.

PR #220, opened 2016-07-11
Add Volatile Read/Write around isStopped for thread safety

### LINKED ISSUE #143: ObserverBase isStopped thread safety question
I was looking at the code of ObserverBase and I found it strange that OnNext and Dispose don't do anything special to ensure thread safety of isStopped.

If thread A calls Dispose, how is it ensured that thread B which calls OnNext sees this change of isStopped?

onovotny:
Should fix #143 

/cc @bartdesmet

### DIFF
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/ObserverBase.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/ObserverBase.cs
@@ -29,7 +29,7 @@ namespace System.Reactive
         /// <param name="value">Next element in the sequence.</param>
         public void OnNext(T value)
         {
-            if (isStopped == 0)
+            if (Volatile.Read(ref isStopped) == 0)
                 OnNextCore(value);
         }
 
@@ -98,7 +98,7 @@ namespace System.Reactive
         {
             if (disposing)
             {
-                isStopped = 1;
+                Volatile.Write(ref isStopped, 1);
             }
         }
 


Rx.NET/Source/System.Reactive.Core/Reactive/ObserverBase.cs
```diff
@@ -98,7 +98,7 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                isStopped = 1;
+                Volatile.Write(ref isStopped, 1);
```
akarnokd: The constructor sets the field to zero; either that is unnecessary or the constructor should use Volatile.Write() as well.

onovotny: The constructor's calls don't have thread contention as the object isn't visible to a caller until fully created. While 0 is the default value of an int, I don't see any harm in being explicit about it.

shiftkey:
:shipit:

Rx.NET/Source/System.Reactive.Core/Reactive/ObserverBase.cs
tmds: @onovotny there is indeed no thread contention in the constructor. The question is: if you don't use a volatile write, are you sure all threads will see the value?
I guess this will be true in this case, because the value is zero. I am not sure this is true in general.

PR #221, opened 2016-07-11
Made Ix-Async factory methods public

### LINKED ISSUE #159: Internal factory methods in Ix-Async
Hi,
I'm trying to write a method that continuously pull data from storage based on task api that use paging with continuation token.
I implemented it with Rx Observable, but I prefer to move the control to the consumer.
The methods I saw that felt suitable for the creation of the AsyncEnumbrable were internal:
```csharp
static IAsyncEnumerator<T> Create<T>(Func<CancellationToken, Task<bool>> moveNext, Func<T> current, Action dispose)

static IAsyncEnumerable<T> Create<T>(Func<IAsyncEnumerator<T>> getEnumerator)
```
Is there any other recommended way to do it? (Like implementing the interfaces with classes)
What's the reason these methods are internal?

slang25:
Fixes #159

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncEnumerable.Creation.cs
@@ -11,7 +11,7 @@ namespace System.Linq
 {
     public static partial class AsyncEnumerable
     {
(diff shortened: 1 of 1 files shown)


shiftkey:
@onovotny do we want this in for 3.0.1?

Issue #158, opened 2016-01-19
SubjectBase


grokys:
I notice that `SubjectBase` has been added (though it's not yet available in any release) however it only defines abstract methods, which is kinda useless. It would be good if it also handled the low-level plumbing with `ImmutableList` etc.

grokys:
Especially as Rx's ImmutableList is internal, so we either need to take a dependency on Immutable Collections, or copy Rx's ImmutableList into our code.

shiftkey:
@grokys is this still something you're interested in?

grokys:
Yeah, it's definitely something that would be useful!

shiftkey:
@grokys ok cool, let's tease this stuff out:

 - what would you like to see changed with `SubjectBase`? For reference, it's mostly being used as a base class, and I get the feeling that's why it's necessary to make public:

<img width="544" src="https://cloud.githubusercontent.com/assets/359239/16738682/a25778a4-474b-11e6-9351-30e414bd9c7e.png">

 - do you think we should adopt `ImmutableList` internally and remove that code? That feels better for the overall ecosystem, rather than making Yet Another Implementation public and encouraging more fragmentation...

shiftkey:
[Here's a rough PR which migrates our usage of the internal `ImmutableList` over to use the `System.Collections.Immutable` version](https://github.com/Reactive-Extensions/Rx.NET/compare/master...shiftkey:be-more-immutable?expand=1) - much easier than I anticipated...

onovotny:
Definitely a 👍 to using the external ImmutableList now and removing our impl now that we can.

flagbug:
We have to be careful with swapping those, see https://github.com/Reactive-Extensions/Rx.NET/pull/222#issuecomment-231825104

grokys:
Ok, it seems that the use-case I was thinking of when I created this issue is no longer relevant, but I can explain what I was thinking:

The problem was that the `ImmutableList` impl in Rx was private so that to create my own subject I would've had to add a new dependency that I didn't previously have. Moving to `System.Collections.Immutable` would of course make that no longer true.

However, there's a lot of commonality to the implementations of `OnNext`/`OnError`/`OnCompleted` in the different subject implementations, which could maybe be exposed, but I think the recently added `Subject.Create` method should make creating a new new subject less common; so probably not worth messing with this.

Closing this for now.

PR #222, opened 2016-07-11
[Proposal] Replace internal ImmutableList<T> with System.Collections.Immutable dependency

shiftkey:
This was suggested in #158 and was a rather trivial change to apply.

I'm still pondering on the implications for having an additional dependency for `System.Reactive.Core`, but I'm mostly okay with it...

I need to go back and tweak the commit history, but the diff is good enough to gather feedback...


flagbug:
@shiftkey Caution, I've already tried to do that a few weeks ago! The internal Rx `ImmutableList` is extremely simple and fast. The `System.Collections.Immutable` list is way more complicated and orders of magnitude slower.

shiftkey:
@flagbug thanks for the heads up, I *knew* this was too easy!

flagbug:
@shiftkey Just FYI, I also tried it with `System.Collections.Immutable.ImmutableArray`, performance was also :hankey:

onovotny:
Are there perf issus filed against the BCL versions? Seems like this is something they'd want to tackle? Even if it's a "LightweightImmutableList", etc?

shiftkey:
@flagbug what sort of testing did you do around this that highlighted the issue?

flagbug:
@shiftkey Can't remember the exact details, but I was experimenting around the `Subscribe` method that was a performance bottleneck in one of my Android applications. Replacing the `ImmutableList` resulted in even worse performance.

shiftkey:
Parking this for now.

danielcweber:
Reopening this because it is still highly desirable and the performance bottleneck could be tried to work around. I guess the poor performance is due to the ImmutableList having a complexity in O(n) when appending elements to the list. It should be examined if performance converges to O(1) when prepending elements. That, however would lead to a reversed order of observer-notification. It's up to debate if we can live with that. Let's discuss.

danielcweber:
First tests show that a reversed order of notifying IObservers does not result in any failed unit tests, i.e. there is apparently nothing depending on the current behaviour.

akarnokd:
There is no guarantee which observer will leave so worst case is O(N) for removal. Also `AsyncSubject` and `Subject` no longer use it and instead have their copy-on-write array management.

> That, however would lead to a reversed order of observer-notification. 

There is a chance users already rely on the ordering, for example, one observer modifying a UI state as it knows  the next observer can pick that updated state.

At the end, it is a tradeoff between supporting a lot of come-and-go observers or supporting a lot of onNexts for a few observers.

danielcweber:
#779, #780 and #781 experimented with different approaches on how to get the immutable collections in. @akarnokd did the benchmarks. It showed massive performance drops for the benchmarked scenarios.

My interpretation of the results:

First, only Subject was affected by the changes, any results on AsyncSubject or BehaviorSubject are noise.

If there ever was some performance gain expected from Immutable Collections, it would have been during subscription, when repeated re-allocation of an array/copying (and thus, garbage) may be avoided. The benefits might actually show with a lot of upfront subscriptions and/or randomized subscription/unsubscription.

The performance penalty we see is mostly due to enumeration, which can't beat the array since the immutable collection traverses a data structure while the array is, well, aligned in memory and thus very cache friendly.

The benefits might show eventually for massive amounts of live subscriptions, however, it's probably not the case for 99% of Subject consumers. The penalty is way too high to go in that direction I guess. However, the naive approach of just copying an array every time is still far from optimal for large changing numbers of subscribers.

akarnokd:
Array copying is cheap for sizes that is smaller than the cache line size (8/16 elements) and I haven't encountered usages that were having that many observers deliberately in practice (with RxJava). There is the occasional power-user with hundreds of observers where I usually suggest implementing a fresh `Subject` on top of `HashSet` with or without an array-based cache to make `onNext` fast during calm times.

danielcweber:
Closing this since we found out that immutable collections are, for most of Rx usage scenarios, inferior in performance to simple array copying. I recently had a look into doubly linked lock free lists, which have very nice complexity characteristics, but as long as Subjects aren't determined to be a bottleneck for anybody, it's hardly worth the effort.

PR #237, opened 2016-07-20
Ix optimizations

onovotny:
This PR is to optimize Ix Async by incorporating the patterns used in https://github.com/dotnet/corefx/tree/master/src/System.Linq/src/System/Linq, taking advantage of the thought leadership and performance reviews of that codebase.


Ix.NET/Source/System.Interactive.Async/AsyncEnumerableHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+namespace System.Collections.Generic
+{
+    // Based on https://github.com/dotnet/corefx/blob/ec2685715b01d12f16b08d0dfa326649b12db8ec/src/Common/src/System/Collections/Generic/EnumerableHelpers.cs
+    internal static class AsyncEnumerableHelpers
+    {
+        internal static async Task<T[]> ToArray<T>(IAsyncEnumerable<T> source, CancellationToken cancellationToken)
+        {
+            var result = await ToArrayWithLength(source, cancellationToken)
+                             .ConfigureAwait(false);
+            Array.Resize(ref result.array, result.length);
```
bartdesmet: Nit: can move this to a `ToArray` method on`ArrayWithLength<T>`.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerableHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+            return result.array;
+        }
+
+        internal static async Task<ArrayWithLength<T>> ToArrayWithLength<T>(IAsyncEnumerable<T> source, CancellationToken cancellationToken)
+        {
+            var result = new ArrayWithLength<T>();
+            // Check for short circuit optimizations. This one is very unlikely
+            // but could be here as a group
+            var ic = source as ICollection<T>;
```
bartdesmet: Interesting cross-over between the `async` and `sync` domains here. I'd be ok omitting this part for the time being until we rationalize async collections more with the BCL team (after I get back from vacation, I'll pick up that thread with the appropriate set of folks again). The rest with dynamic array growth etc. is goodness though; very similar to the internal buffering in `ToArray` in LINQ to Objects.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerableHelpers.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            else
+            {
+                using (var en = source.GetEnumerator())
+                {
+                    if (await en.MoveNext(cancellationToken)
+                                .ConfigureAwait(false))
+                    {
+                        const int DefaultCapacity = 4;
+                        var arr = new T[DefaultCapacity];
```
bartdesmet: May or may not be a good candidate to evaluate `System.Buffers` in the future.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        internal enum AsyncIteratorState
+        {
+            New = 0,
+            Allocated = 1,
+            Iterating = 2,
+            Disposed = -1,
+        }
+
+        internal abstract class AsyncIterator<TSource> : IAsyncEnumerable<TSource>, IAsyncEnumerator<TSource>
```
bartdesmet: Great! Could become a candidate for external exposure for operator authors. We can do that later though; let's first battle test it internally, and also see whether we can align with future C# directions for async iterators.

Ix.NET/Source/System.Interactive.Async/AsyncEnumerableHelpers.cs
onovotny: I think we'll potentially hit these fairly often with the optimizations I've made to `ToAsyncEnumerable` and its use in various joins with Grouping and Lookup.

Ix.NET/Source/System.Interactive.Async/Concatenate.cs
```diff
@@ -80,53 +44,246 @@ public static IAsyncEnumerable<TSource> Concat<TSource>(params IAsyncEnumerable<
 
         private static IAsyncEnumerable<TSource> Concat_<TSource>(this IEnumerable<IAsyncEnumerable<TSource>> sources)
         {
-            return CreateEnumerable(
-                () =>
+            using (var e = sources.GetEnumerator())
+            {
+                IAsyncEnumerable<TSource> prev = null;
+                while (e.MoveNext())
```
bartdesmet: This seems too eager; it drains the `sources` by creating a chain of `Concat`s rather than deferring the top-level `IEnumerable<IAsyncEnumerable<T>>`'s enumerator's `MoveNext` call until the inner sequence has been drained. It also doesn't seem to defer the enumeration of the outer sequence until the first `MoveNextAsync` call.

This should be tested with `Enumerable.Repeat(AsyncEnumerable.Return(42)).Concat().Take(1)` to ensure it evaluates as lazily as possible.

onovotny: I'm not sure how to best reconcile this overload with the `ConcatAsyncNIterator` pattern. The pattern there is that it effectively constructs a linked list of previous `ConcatNAsyncIterator`'s. The one returned to the caller must be the last one in the chain as it reaches back to the first one.

The notes in there describe it like this:
> // Walk back through the chain of ConcatNIterators looking for the one
                // that has its _nextIndex equal to index.  If we don't find one, then it
                // must be prior to any of them, so call GetEnumerable on the previous
                // Concat2Iterator.  This avoids a deep recursive call chain.

I can certainly return the original logic for this method if it has better characteristics, but it seems odd that calling `Concat` multiple times is vastly different than passing them in all at once.

Thoughts?

onovotny: Thinking about it, I can probably write another specialized iterator to handle this case.

bartdesmet: The async iterator's state needs to be the combination of the outer (sync) enumerator and the currently running inner async enumerator, just like it was done before.

onovotny: Done.

Ix.NET/Source/System.Interactive.Async/ToAsyncEnumerable.cs
```diff
@@ -91,5 +79,237 @@ private static IEnumerable<TSource> ToEnumerable_<TSource>(IAsyncEnumerable<TSou
...
+            }
+
+            public Task<List<T>> ToListAsync(CancellationToken cancellationToken)
+            {
+                return Task.FromResult(source.ToList());
+            }
+
+            public Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
+            {
+                return Task.FromResult(source.Count());
```
danielcweber: Use `Count` instead of Count() since `source` is an `IList`.

Ix.NET/Source/System.Interactive.Async/ToAsyncEnumerable.cs
```diff
@@ -91,5 +79,237 @@ private static IEnumerable<TSource> ToEnumerable_<TSource>(IAsyncEnumerable<TSou
...
+            {
+                return Task.FromResult(source.ToArray());
+            }
+            public Task<List<T>> ToListAsync(CancellationToken cancellationToken)
+            {
+                return Task.FromResult(source.ToList());
+            }
+            public Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
+            {
+                return Task.FromResult(source.Count());
```
danielcweber: Use `Count` instead of `Count()` since source is an `ICollection`.

Ix.NET/Source/System.Interactive.Async/Buffer.cs
```diff
@@ -31,72 +31,103 @@ public static IAsyncEnumerable<IList<TSource>> Buffer<TSource>(this IAsyncEnumer
...
+                                }
+
+                                if (buffers.Count > 0 && buffers.Peek()
+                                                                .Count == count)
+                                {
+                                    current = buffers.Dequeue();
+                                    return true;
+                                }
+
+                                goto case AsyncIteratorState.Iterating; // loop
```
bartdesmet: We may be able to use regular loop constructs here; please consider if feasible.

Ix.NET/Source/System.Interactive.Async/Catch.cs
```diff
@@ -96,60 +52,174 @@ public static IAsyncEnumerable<TSource> Catch<TSource>(this IAsyncEnumerable<TSo
...
-                                    return false;
+                                    current = enumerator.Current;
+                                    return true;
                                 }
+                            }
+                            catch (TException ex)
+                            {
+                                var err = handler(ex)
+                                    .GetEnumerator();
+                                enumerator?.Dispose();
```
bartdesmet: Add a comment that this order is kept for compat with previous versions. Ideally, we'd have disposed the current enumerator prior to invoking the `handler`, but we should keep the original behavior.

Ix.NET/Source/System.Interactive.Async/DefaultIfEmpty.cs
```diff
@@ -17,53 +19,111 @@ public static IAsyncEnumerable<TSource> DefaultIfEmpty<TSource>(this IAsyncEnume
...
+                        if (await enumerator.MoveNext(cancellationToken)
+                                            .ConfigureAwait(false))
+                        {
+                            current = enumerator.Current;
+                            state = AsyncIteratorState.Iterating;
+                        }
+                        else
+                        {
+                            current = defaultValue;
+                            state = AsyncIteratorState.Disposed;
```
bartdesmet: Will the `enumerator` get disposed?

Ix.NET/Source/System.Interactive.Async/DefaultIfEmpty.cs
```diff
@@ -17,53 +19,111 @@ public static IAsyncEnumerable<TSource> DefaultIfEmpty<TSource>(this IAsyncEnume
...
+                var list = await source.ToList(cancellationToken).ConfigureAwait(false);
+                if (list.Count == 0)
+                {
+                    list.Add(defaultValue);
+                }
+
+                return list;
+            }
+
+            public async Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
```
bartdesmet: Do we have tests for *all* combinations of these optimized paths? We definitely didn't originally because we didn't have such optimizations and relied on the compositionality of operators as the inductive "proof" of correctness.

Ix.NET/Source/System.Interactive.Async/Distinct.cs
```diff
@@ -102,56 +95,375 @@ public static IAsyncEnumerable<TSource> DistinctUntilChanged<TSource>(this IAsyn
...
+            public async Task<List<TSource>> ToListAsync(CancellationToken cancellationToken)
+            {
+                var s = await FillSet(cancellationToken)
+                            .ConfigureAwait(false);
+                return s;
+            }
+
+            public async Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
+            {
+                return onlyIfCheap ? -1 : (await FillSet(cancellationToken)
```
bartdesmet: Do we need to allocate a list in order to count? This seems more expensive than the naïve non-optimized case where we'd just have the set allocated and the `Count` operator simply using a single `Int32`.

Ix.NET/Source/System.Interactive.Async/Distinct.cs
```diff
@@ -102,56 +95,375 @@ public static IAsyncEnumerable<TSource> DistinctUntilChanged<TSource>(this IAsyn
...
+            public async Task<List<TSource>> ToListAsync(CancellationToken cancellationToken)
+            {
+                var s = await FillSet(cancellationToken)
+                            .ConfigureAwait(false);
+                return s.ToList();
+            }
+
+            public async Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
+            {
+                return onlyIfCheap ? -1 : (await FillSet(cancellationToken)
```
bartdesmet: Same as above. If we decide to go for fast paths for operators like `ToList`, `ToArray`, and `Count` (rationale to be elaborated on to make a final decision), we should make sure they *really* result in an optimization.

Ix.NET/Source/System.Interactive.Async/Distinct.cs
```diff
@@ -102,56 +95,375 @@ public static IAsyncEnumerable<TSource> DistinctUntilChanged<TSource>(this IAsyn
...
+                            }
+                            if (!hasCurrentValue || !comparerEquals)
+                            {
+                                hasCurrentValue = true;
+                                currentValue = item;
+                                current = item;
+                                return true;
+                            }
+
+                            goto case AsyncIteratorState.Iterating; // loop
```
bartdesmet: Can a regular loop be used?

bartdesmet: Same for other places; please check across the whole PR.

Ix.NET/Source/System.Interactive.Async/Do.cs
```diff
@@ -72,52 +73,84 @@ public static IAsyncEnumerable<TSource> Do<TSource>(this IAsyncEnumerable<TSourc
...
-
-                    return CreateEnumerator(
-                        f,
-                        () => current,
-                        d.Dispose,
-                        e
-                    );
-                });
+                        }
+                        catch (OperationCanceledException)
```
bartdesmet: Add a note we do this for compatibility; I think we should have been more narrow here and added a filter to check whether the cancellation was due to the `cancellationToken` being set.

Ix.NET/Source/System.Interactive.Async/Except.cs
```diff
@@ -31,40 +31,105 @@ public static IAsyncEnumerable<TSource> Except<TSource>(this IAsyncEnumerable<TS
...
+                }
+
+                return false;
+            }
+
+            private async Task FillSet(CancellationToken cancellationToken)
+            {
+                var array = await second.ToArray(cancellationToken)
+                                        .ConfigureAwait(false);
+                for (var i = 0; i < array.Length; i++)
```
bartdesmet: NB: A regular `foreach` over an array will emit the same `for` loop construct, but is more concise on the eye.

Ix.NET/Source/System.Interactive.Async/Generate.cs
```diff
@@ -20,44 +21,69 @@ public static partial class AsyncEnumerable
...
+                        started = false;
+                        currentState = initialState;
 
+                        state = AsyncIteratorState.Iterating;
+                        goto case AsyncIteratorState.Iterating;
+
+                    case AsyncIteratorState.Iterating:
+                        if (started)
+                        {
+                            currentState = iterate(currentState);
```
bartdesmet: Is it ok for exceptions that happen here to come out of the `MoveNextCore` method and not through the `Task<bool>` it returns? This could be brittle and depend on how the caller of `MoveNextCore` is implemented, so merely making sure we're aware of this possible pitfall.

bartdesmet: Right now, it seems ok because `MoveNext` is implemented as an `async` method. A remark may be warranted there in order to prevent someone changing it to a non-`async` method and forgetting to deal with the case where `MoveNextCore` throws.

Ix.NET/Source/System.Interactive.Async/Do.cs
onovotny: Should I add that exception for the `cancellationToken` here now or has that ship sailed for compatibility?

Ix.NET/Source/System.Interactive.Async/DefaultIfEmpty.cs
onovotny: Not yet, but good point - we have a lot more tests we need to add.

Ix.NET/Source/System.Interactive.Async/Generate.cs
onovotny: Done.

Ix.NET/Source/System.Interactive.Async/Distinct.cs
WilkaH: The Immutable Collections package often tries to use [TryGetCount](https://github.com/dotnet/corefx/blob/d0dc5fc099946adc1035b34a8b1f6042eddb0c75/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableExtensions.cs#L38) to quickly get the count of a sequence, then falls back to a slower method if it can't do it quickly. e.g. [inside AddRange](https://github.com/dotnet/corefx/blob/21fd3b96ff8c470cb4f98c316821060c8ef25944/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Builder.cs#L250)

onovotny:
@bartdesmet The `goto case` loops have been replaced with `while` and `do` loops as appropriate. While in the iterator mindset, it was too easy to fall into the pit of `goto` ;)

onovotny:
BTW, we now have code coverage enabled in the command-line build and for CI. The CI build publishes to coveralls.io, like CoreFX does.

There doesn't yet seem to be any working code coverage tooling that works within VS.

https://coveralls.io/github/Reactive-Extensions/Rx.NET

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -0,0 +1,136 @@
...
+            internal AsyncIteratorState state = AsyncIteratorState.New;
+
+            protected AsyncIterator()
+            {
+                threadId = Environment.CurrentManagedThreadId;
+            }
+
+            public IAsyncEnumerator<TSource> GetEnumerator()
+            {
+                var enumerator = state == AsyncIteratorState.New && threadId == Environment.CurrentManagedThreadId ? this : Clone();
```
shiftkey: :lipistck: move the conditional terms down to the next line so this is a bit easier to read:

```csharp
 var enumerator = state == AsyncIteratorState.New && threadId == Environment.CurrentManagedThreadId
    ? this
    : Clone();
```

shiftkey: Or however we're formatting these in the codebase `*shrug*`

Ix.NET/Source/System.Interactive.Async/Concatenate.cs
```diff
@@ -80,53 +44,318 @@ public static IAsyncEnumerable<TSource> Concat<TSource>(params IAsyncEnumerable<
...
+        }
+
+        private sealed class Concat2AsyncIterator<TSource> : ConcatAsyncIterator<TSource>
+        {
+            private readonly IAsyncEnumerable<TSource> first;
+            private readonly IAsyncEnumerable<TSource> second;
+
+            internal Concat2AsyncIterator(IAsyncEnumerable<TSource> first, IAsyncEnumerable<TSource> second)
+            {
+                Debug.Assert(first != null && second != null);
```
shiftkey: I like doing this sort of thing for the internal contracts, should we encourage it more broadly?

Ix.NET/Source/System.Interactive.Async/Count.cs
```diff
@@ -17,6 +17,18 @@ public static Task<int> Count<TSource>(this IAsyncEnumerable<TSource> source, Ca
             if (source == null)
                 throw new ArgumentNullException(nameof(source));
 
+            var collectionoft = source as ICollection<TSource>;
```
shiftkey: :lipstick: just call this `collection`?

Ix.NET/Source/System.Interactive.Async/Create.cs
```diff
@@ -91,37 +65,55 @@ public IAsyncEnumerator<T> GetEnumerator()
...
             }
 
-            public Task<bool> MoveNext(CancellationToken cancellationToken)
+            public override AsyncIterator<T> Clone()
             {
-                if (_disposed)
-                    return TaskExt.False;
-
-                return _moveNext(cancellationToken);
+                throw new NotSupportedException("Iterator only");
```
shiftkey: Can we make this message a bit more verbose?

Ix.NET/Source/System.Interactive.Async/DefaultIfEmpty.cs
```diff
@@ -17,53 +19,114 @@ public static IAsyncEnumerable<TSource> DefaultIfEmpty<TSource>(this IAsyncEnume
...
+        {
+            private readonly IAsyncEnumerable<TSource> source;
+            private readonly TSource defaultValue;
+            private IAsyncEnumerator<TSource> enumerator;
+
+            public DefaultIfEmptyAsyncIterator(IAsyncEnumerable<TSource> source, TSource defaultValue)
+            {
+                this.source = source;
+                this.defaultValue = defaultValue;
+                Debug.Assert(source != null);
```
shiftkey: :lipstick: move this up before assigning the fields

Ix.NET/Source/System.Interactive.Async/ToAsyncEnumerable.cs
```diff
@@ -91,5 +79,242 @@ private static IEnumerable<TSource> ToEnumerable_<TSource>(IAsyncEnumerable<TSou
...
+        }
+
+        internal sealed class AsyncICollectionEnumerableAdapter<T> : AsyncIterator<T>, IIListProvider<T>, ICollection<T>
+        {
+            private readonly ICollection<T> source;
+            private IEnumerator<T> enumerator;
+            public AsyncICollectionEnumerableAdapter(ICollection<T> source)
+            {
+                Debug.Assert(source != null);
+                this.source = source;
```
shiftkey: :lipstick: whitespace before the assign statement

Ix.NET/Source/System.Interactive.Async/ToAsyncEnumerable.cs
```diff
@@ -91,5 +79,242 @@ private static IEnumerable<TSource> ToEnumerable_<TSource>(IAsyncEnumerable<TSou
...
+
+        internal sealed class AsyncIListEnumerableAdapter<T> : AsyncIterator<T>, IIListProvider<T>, IList<T>
+        {
+            private readonly IList<T> source;
+            private IEnumerator<T> enumerator;
+
+            public AsyncIListEnumerableAdapter(IList<T> source)
+            {
+                Debug.Assert(source != null);
+                this.source = source;
```
shiftkey: :lipstick: whitespace before

Ix.NET/Source/System.Interactive.Async/Expand.cs
```diff
@@ -19,59 +19,89 @@ public static IAsyncEnumerable<TSource> Expand<TSource>(this IAsyncEnumerable<TS
...
+            }
+
+            public override AsyncIterator<TSource> Clone()
+            {
+                return new ExpandAsyncIterator<TSource>(source, selector);
+            }
+
+            public override void Dispose()
+            {
+                if (enumerator != null)
```
shiftkey: :lipstick: could this sort of repetitive code be simplified:

```csharp
enumerator?.Dispose();
enumerator = null;
```

onovotny: I wasn't sure on this. Happy to make this change everywhere if desired.

Ix.NET/Source/System.Interactive.Async/First.cs
```diff
@@ -88,6 +88,12 @@ public static Task<TSource> FirstOrDefault<TSource>(this IAsyncEnumerable<TSourc
 
         private static async Task<TSource> First_<TSource>(IAsyncEnumerable<TSource> source, CancellationToken cancellationToken)
         {
+            var list = source as IList<TSource>;
+            if (list?.Count > 0)
```
shiftkey: No need for `?` here

Ix.NET/Source/System.Interactive.Async/Grouping.cs
```diff
@@ -238,98 +117,269 @@ public static partial class AsyncEnumerable
...
+            private readonly Func<TSource, TKey> keySelector;
+            private readonly Func<TKey, IAsyncEnumerable<TSource>, TResult> resultSelector;
+            private readonly IEqualityComparer<TKey> comparer;
+
+            Internal.Lookup<TKey, TSource> lookup;
+            IEnumerator<TResult> enumerator;
+
+            public GroupedResultAsyncEnumerable(IAsyncEnumerable<TSource> source, Func<TSource, TKey> keySelector, Func<TKey, IAsyncEnumerable<TSource>, TResult> resultSelector, IEqualityComparer<TKey> comparer)
+            {
+                if (source == null) throw new ArgumentNullException(nameof(source));
```
shiftkey: This is different to how we've used `Debug.Assert` elsewhere for these new iterators. Am I missing some context on this addition?

Ix.NET/Source/System.Interactive.Async/Grouping.cs
```diff
@@ -238,98 +117,269 @@ public static partial class AsyncEnumerable
...
-            public AsyncGrouping(TKey key, Func<CancellationToken, Task<bool>> iterateSource, IDisposable sourceDisposable)
+            private Internal.Lookup<TKey, TElement> lookup;
+            private IEnumerator<IGrouping<TKey, TElement>> enumerator;
+
+            public GroupedAsyncEnumerable(IAsyncEnumerable<TSource> source, Func<TSource, TKey> keySelector, Func<TSource, TElement> elementSelector, IEqualityComparer<TKey> comparer)
             {
-                this.iterateSource = iterateSource;
-                this.sourceDisposable = sourceDisposable;
-                Key = key;
+                if (source == null) throw new ArgumentNullException(nameof(source));
```
shiftkey: Another place where we can get away with `Debug.Assert` - guard clauses on public methods calling into this should protect us here...

Ix.NET/Source/System.Interactive.Async/Grouping.cs
```diff
@@ -238,98 +117,269 @@ public static partial class AsyncEnumerable
...
+                        Dispose();
+                        break;
+                }
+
+                return false;
+            }
+            
+            public async Task<IAsyncGrouping<TKey, TElement>[]> ToArrayAsync(CancellationToken cancellationToken)
+            {
+                IIListProvider<IAsyncGrouping<TKey, TElement>> lookup = await Internal.Lookup<TKey, TElement>.CreateAsync(source, keySelector, elementSelector, comparer, cancellationToken).ConfigureAwait(false);
```
shiftkey: :lipstick: variable name clashes with field - rename? (there's a few places later in this file too)

Ix.NET/Source/System.Interactive.Async/Expand.cs
shiftkey: Eh, minor thing. Let's punt on it for now.

Ix.NET/Source/System.Interactive.Async/Lookup.cs
```diff
@@ -377,5 +348,94 @@ private void Resize()
 
             _groupings = newGroupings;
         }
+
+        IAsyncEnumerator<IGrouping<TKey, TElement>> IAsyncEnumerable<IGrouping<TKey, TElement>>.GetEnumerator()
+        {
+            return this.ToAsyncEnumerable<IGrouping<TKey, TElement>>().GetEnumerator();
```
shiftkey: :lipstick: can drop the explicit `<IGrouping<TKey, TElement>>`

Ix.NET/Source/System.Interactive.Async/Reverse.cs
```diff
@@ -16,38 +19,104 @@ public static IAsyncEnumerable<TSource> Reverse<TSource>(this IAsyncEnumerable<T
...
+            }
+
+            public Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationToken)
+            {
+                if (onlyIfCheap)
+                {
+                    var listProv = source as IIListProvider<TSource>;
+                    if (listProv != null)
+                    {
+                        return listProv.GetCountAsync(onlyIfCheap: true, cancellationToken: cancellationToken);
```
shiftkey: :lipstick: drop the named parameters here?

shiftkey:
An open question about the use of `Debug.Assert` versus throwing an exception (some places we do one, or the other, or neither) and now I'm mostly picking up stylistic things while reviewing - I could probably open a PR to just do those changes directly, or just push to this branch.

I also kind of agree with @bartdesmet's comment on how much we can trust the current set of tests. I'll see if I can find time this week to dig into those and see where we can improve them.

Ix.NET/Source/System.Interactive.Async/Where.cs
```diff
@@ -19,35 +20,14 @@ public static IAsyncEnumerable<TSource> Where<TSource>(this IAsyncEnumerable<TSo
...
-                        };
-
-                    return CreateEnumerator(
-                        ct => f(cts.Token),
-                        () => e.Current,
-                        d.Dispose,
-                        e
-                    );
-                });
+            // TODO: Can we add array/list optimizations here, does it make sense?
```
shiftkey: This seems reasonable, but we still have to convert back to an `IAsyncEnumerable` here - that might mean it's easier to let it pass through:

```
var list = source as IList<TSource>;
if (list != null)
{
    return list.Where(predicate).ToAsyncEnumerable();
}
```

onovotny:
@shiftkey about the current state of the tests -- they're actually not too far off. That's why I hooked up the code coverage to verify current/new gaps. I'm going through the coverage report to fill new tests as needed to cover the new bits.

Ix.NET/Source/System.Interactive.Async/Reverse.cs
```diff
@@ -67,7 +67,7 @@ public Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancellationT
                     var listProv = source as IIListProvider<TSource>;
                     if (listProv != null)
                     {
-                        return listProv.GetCountAsync(onlyIfCheap, cancellationToken);
+                        return listProv.GetCountAsync(true, cancellationToken);
```
onovotny: shouldn't this be passing the `onlyIfCheap` parameter along instead of overriding it?

Ix.NET/Source/System.Interactive.Async/Where.cs
onovotny: So far, the only place that'd be true from our code is in paths that are already wrapped via calling `ToAsyncEnumerable` and are thus in an `AsyncIListEnumerableAdapter`. -- So the earlier if/else that tests for the `AsyncIterator<T>` base will pass. I don't think there's anyway to hit an `IList<T>` for our code, but it's possible someone else implements it somehow. That said, I'm ok with ignoring it for now.

Ix.NET/Source/System.Interactive.Async/Distinct.cs
onovotny: The implementation of this code has been fixed so that FillSet doesn't allocate a List -- only the set. Not sure why GitHub isn't showing this as outdated.

onovotny:
Okay @bartdesmet @shiftkey. Test coverage should be complete now...and is 1.5% higher than it was before ;) You can see in the coveralls report (or by running a local build) that we've now hit pretty much all of the new code and ensured the optimization paths all work.

Ix.NET/Source/System.Interactive.Async/Reverse.cs
shiftkey: Yeah, you're right.

shiftkey: Actually, this is wrapped in `if (onlyIfCheap)` so this is equivalent - there's a few places where we do this, so I'm going to leave it as-is.

Ix.NET/Source/System.Interactive.Async/DefaultIfEmpty.cs
onovotny: Following up to this note -- yes, we have full test coverage now.

PR #241, opened 2016-07-31
Lock free SerialDisposable

LeeCampbell:
Makes the internal implementation of `SerialDisposable` a lock free and wait free implementation.

The design is fairly simple. As the object can only be in one of two states (Active or Disposed), these two states are represented internally as two separate implementations. While active, the setting of the `Disposable` property just executes an atomic exchange of the old value for the new value. The old value is disposed.

When an instance of `SerialDisposable` is disposed, it migrates to its disposed state by performing an atomic exchange of the internal implementations.

While the goal of this is a performance improvement, it is in the realms of _marginal gains_. I don't expect a perceivable difference in performance for the Rx code base. However my theory is if it can run faster, with no other observable changes in the behavior or API, then it is a win. If this PR is accepted, then perhaps other areas of the Rx code base can be targeted too.

Performance tests across various x86 (mobile) processors (Atom, i3, i5 & i7) there is a performance improvement. With the notable exception of when running throughput test on an i7 with 2 threads there is a performance drop. i.e. of the 24 variations of tests this was the only variation that showed a reduction (~12%) in performance. All other variations (1 - 9 threads for i7 and 1-5 threads for i5, i3 and Atom processors) showed performance increase from 12% to 180%.

Single-threaded performance all showed between 81-99% improvement. Running two threads was the worst performance increase for for the i7, i5 and i3 processors. Generally best performance gains were seen when the number of threads matched the number of logical cores.

What we are really seeing is that the lock free implementation has much better single threaded performance and consistent contended performance in contrast to the existing implementation that has degrading performance as contention increases.

# Summary results
Measurements are Operations per second, where an operation is just assigning a `BooleanDisposable` to the `SerialDisposable.Disposable` property. e.g `24M` is 24,000,000 operations per second. The summaries below round down to the nearest million. The full test data is attached.

## I7-4702HQ@2.20GHZ	

	Threads		   1       2       3	   4	   5	   6	   7	   8	   9
	--------------------------------------------------------------------------------
	Existing	 29M	 24M	 19M	 20M	 20M   	 20M     18M	 12M	 13M
	Lockfree	 59M	 21M	 24M	 22M 	 24M	 28M	 29M	 31M	 29M
	--------------------------------------------------------------------------------
	Scaled		1.99	0.88	1.25	1.12	1.18	1.42	1.56	2.43	2.26

## I5-3317U@1.70GHZ

	Threads		   1       2       3	   4	   5
	------------------------------------------------
	Existing	 22M	 17M	 15M	 10M	 10M
	Lockfree	 43M	 22M	 20M	 26M	 28M	
	------------------------------------------------
	Scaled		1.93	1.27	1.31	2.39	2.80

## I3-2365M@1.40GHZ

	Threads		   1       2       3	   4	   5
	------------------------------------------------
	Existing	 13M	  9M	  7M	  6M	  5M
	Lockfree	 23M	 12M	 13M	 16M	 14M
	------------------------------------------------
	Scaled		1.81	1.36	1.94	2.64	2.58

## ATOMZ2760@1.80GHZ

	Threads		   1       2       3	   4	   5
	------------------------------------------------
	Existing	  7M	  3M	  3M	  3M	  3M
	Lockfree	 14M	  6M	  8M	  8M	  8M
	------------------------------------------------
	Scaled		1.91	1.98	2.16	2.60	2.63

[SerialDisposable Results.xlsx](https://github.com/Reactive-Extensions/Rx.NET/files/392337/SerialDisposable.Results.xlsx)

Test source code available here - https://github.com/LeeCampbell/RxPerfTests/tree/c37dc941d8cbfe381c85b5595a0b2d1da9381db6

### DIFF
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information. 
 
+using System.Threading;
+
 namespace System.Reactive.Disposables
 {
     /// <summary>
@@ -9,9 +11,7 @@ namespace System.Reactive.Disposables
     /// </summary>
     public sealed class SerialDisposable : ICancelable
     {
-        private readonly object _gate = new object();
-        private IDisposable _current;
-        private bool _disposed;
+        private ISerialCancelable _current = new ActiveSerialDisposable();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="T:System.Reactive.Disposables.SerialDisposable"/> class.
@@ -23,16 +23,7 @@ namespace System.Reactive.Disposables
         /// <summary>
         /// Gets a value that indicates whether the object is disposed.
         /// </summary>
-        public bool IsDisposed
-        {
-            get
-            {
-                lock (_gate)
-                {
-                    return _disposed;
-                }
-            }
-        }
+        public bool IsDisposed => _current.IsDisposed;
 
         /// <summary>
         /// Gets or sets the underlying disposable.
@@ -40,29 +31,8 @@ namespace System.Reactive.Disposables
         /// <remarks>If the SerialDisposable has already been disposed, assignment to this property causes immediate disposal of the given disposable object. Assigning this property disposes the previous disposable object.</remarks>
         public IDisposable Disposable
         {
-            get
-            {
-                return _current;
-            }
-
-            set
-            {
-                var shouldDispose = false;
-                var old = default(IDisposable);
-                lock (_gate)
-                {
-                    shouldDispose = _disposed;
-                    if (!shouldDispose)
-                    {
-                        old = _current;
-                        _current = value;
-                    }
-                }
-                if (old != null)
-                    old.Dispose();
-                if (shouldDispose && value != null)
-                    value.Dispose();
-            }
+            get { return _current.Disposable; }
+            set { _current.Disposable = value; }
         }
 
         /// <summary>
@@ -70,20 +40,66 @@ namespace System.Reactive.Disposables
         /// </summary>
         public void Dispose()
         {
-            var old = default(IDisposable);
+            var old = Interlocked.Exchange(ref _current, DisposedSerialDisposable.Instance);
+            old.Dispose();
+        }
 
-            lock (_gate)
+        /// <summary>
+        /// Private interface to allow substitution of two implementations based on the state of the serial disposable.
+        /// </summary>
+        private interface ISerialCancelable : ICancelable
+        {
+            IDisposable Disposable { get; set; }
+        }
+        
+        /// <summary>
+        /// Internal implementation of a <see cref="SerialDisposable"/> that has been disposed.
+        /// </summary>
+        /// <remarks>
+        /// Is a singleton implementation as it does not maintain any state.
+        /// </remarks>
+        private sealed class DisposedSerialDisposable : ISerialCancelable
+        {
+            public static readonly DisposedSerialDisposable Instance = new DisposedSerialDisposable();
+
+            private DisposedSerialDisposable()
+            { }
+
+            public bool IsDisposed => true;
+
+            public IDisposable Disposable
             {
-                if (!_disposed)
+                get { return null; }
+                set { value?.Dispose(); }
+            }
+
+            public void Dispose()
+            { }
+        }
+
+        /// <summary>
+        /// Internal implementation of a <see cref="SerialDisposable"/> that is still active.
+        /// </summary>
+        private sealed class ActiveSerialDisposable : ISerialCancelable
+        {
+            private IDisposable _disposable;
+
+            public bool IsDisposed => false;
+
+            public IDisposable Disposable
+            {
+                get { return _disposable; }
+                set
                 {
-                    _disposed = true;
-                    old = _current;
-                    _current = null;
+                    var old = Interlocked.Exchange(ref _disposable, value);
+                    old?.Dispose();
                 }
             }
 
-            if (old != null)
-                old.Dispose();
+            public void Dispose()
+            {
+                _disposable?.Dispose();
+            }
         }
     }
 }
--- a/Rx.NET/Source/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
+++ b/Rx.NET/Source/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
@@ -511,7 +511,7 @@ namespace ReactiveTests.Tests
             m.Dispose();
             Assert.True(m.IsDisposed);
             Assert.True(disp);
-            //Assert.Null(m.Disposable); // BREAKING CHANGE v2 > v1.x - Undefined behavior after disposal.
+            Assert.Null(m.Disposable);   // At v1.1 Behaviour after disposal was undefined. v3 can now enforce the breaking change behavior from 1.1->2.0
         }
 
         [Fact]


Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
-            var old = default(IDisposable);
+            private IDisposable _disposable;
+
+            public bool IsDisposed => false;
 
-            lock (_gate)
+            public IDisposable Disposable
             {
-                if (!_disposed)
+                get { return _disposable; }
```
akarnokd: Volatile read?

akarnokd:
Actually, you don't need this indirection and work on a single IDisposable [field directly](https://github.com/reactor/reactor-core-dotnet/blob/master/Reactor.Core/util/DisposableHelper.cs).

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
+                set
                 {
-                    _disposed = true;
-                    old = _current;
-                    _current = null;
+                    var old = Interlocked.Exchange(ref _disposable, value);
```
akarnokd: I think there is a race condition when the outer is disposed at the same time the active reference is still active and this exchange swap in a live Disposable that doesn't get disposed at the end.

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
+                    old?.Dispose();
                 }
             }
 
-            if (old != null)
-                old.Dispose();
+            public void Dispose()
+            {
+                _disposable?.Dispose();
```
akarnokd: Volatile read of `_disposable`?

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
         /// Gets or sets the underlying disposable.
         /// </summary>
         /// <remarks>If the SerialDisposable has already been disposed, assignment to this property causes immediate disposal of the given disposable object. Assigning this property disposes the previous disposable object.</remarks>
         public IDisposable Disposable
         {
-            get
-            {
-                return _current;
-            }
+            get { return _current.Disposable; }
```
akarnokd: Volatile read on `_current`?

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
+            set { _current.Disposable = value; }
```
akarnokd: Volatile read on `_current`?

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
```diff
@@ -23,67 +23,83 @@ public SerialDisposable()
...
-        {
-            get
-            {
-                lock (_gate)
-                {
-                    return _disposed;
-                }
-            }
-        }
+        public bool IsDisposed => _current.IsDisposed;
```
akarnokd: Volatile read on `_current`.

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
LeeCampbell: Thank for the review. I believe (but happy to be corrected) that as all writes to the field are via `Interlocked.Exchange` that these will enforce the fence. Multiple reads are safe. Reads in the presence of writes will be ordered due to the fence at write time?

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
LeeCampbell: Yes I see how this could happen.

LeeCampbell:
Regarding the indirection vs the single field; the idea was to remove the conditional branches and loops which in my early implementations either exposed race conditions or brought performance back to that of using a lock

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
akarnokd: Plain reads could be reordered by the compiler, JIT or the CPU in a way that breaks the expected visibility. For example, this plain read may be pulled out completely and no matter how the field changes on, you'll see the old value. If there is an exchange on a field, there must be a volatile read because these come and work in pairs. In addition, it may appear to work on x86 which is a strong TSO architecture but fail on weak architectures such as ARM. Yes there are cases where you can go relaxed but not in terms of `SerialDisposable`.

akarnokd:
My experience with `SerialDisposable` is that you rarely hammer it with multiple `Disposable`s concurrently, first of all because it will be non-deterministic who's got to stay at the end, but rather make sure a concurrent `dispose` call disposes the contained or about to be contained `Disposable`. Since this is practically a rare event, the cost of branch and CompareExchange is very low compared to the C# locks.

Also note that benchmarking locked structures for throughput can be misleading as locks tend to serialize the bursts of the concurrent sources. This is because most of the time, thread parking/unparking and arbitration has a high cost, sometimes several microseconds, during which time an individual thread can burst through their own work not competing on the lock before each action.

Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
LeeCampbell: Cheers. As soon as I posted that comment I thought non-x86 architectures may behave differently.
A quick update to my tests locally show no change in performance by adding the `Volatile.Read` commands as suggested.

akarnokd: On x86, basically all reads are volatile reads as the CPU doesn't reorder reads with each other so there is no actual cost increase in execution. Generally though, a compiler seeing an explicit volatile read may not cache a read that is actually safe to cache (like reading a readonly field) and you have to manually pull those out of the volatile regions.

LeeCampbell:
Thanks for helping with this. Due to the race condition you have pointed out (dispose outer while inner is being set), this PR seems dead-in-the-water anyway.

For what it is worth, my quick tests have shown C# locks to be pretty quick and branching to be fairly slow (compared to what I thought would be the costs). Note that the original code has both locks and numerous branches through out. The `set_Disposable` having 1 lock and 4 conditionals, and the `Dispose` method have 1 lock and 2 conditionals. Seeing this was what initially made me think can this be done better, and has been my pet project for several months.

With regards to benchmarking locked structures, yes I can see how the bursting on a single thread would happen. But this *could* happen in a highly contended system in production too, right?
With this in mind, are there any tools I should look to help validate the code and benchmark it appropriately? I have used [BenchmarkDotNET](https://github.com/PerfDotNet/BenchmarkDotNet) before, but I don't think it designed for multi-threaded throughput testing, so I stuck with my own test set up.

akarnokd:
I think the dominant benchmark for `SerialDisposable` is the single threaded one, setting elements one after another. I don't think .NET does lock elision so you can't get faster than using Interlocked and Volatile-based algorithms. I believe it is worth making these containers lock free. I hope you looked at the `DisposableHelper` I linked whose methods practically have the fewest instructions that make the operations still correct. If you could apply the same concept here, that's a similar huge win. (On a side note,  the overhead reduction of lock-free comes from the fact that it only incurs 1 CAS whereas the most thinnest lock has 2 CAS: enter, exit).

akarnokd:
FYI: #246.

Issue #227, opened 2016-07-17
Possible race condition in WithLatestFrom when setting the latest


akarnokd:
If I understand C# memory model correctly (should be very similar to Java's), there is a potential race when [reading _latest](https://github.com/Reactive-Extensions/Rx.NET/blob/1ad21f5a853f2b7be5f3a7e4a8d276e7d28f0d73/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs#L95) in the main 'rail' and [writing _latest](https://github.com/Reactive-Extensions/Rx.NET/blob/1ad21f5a853f2b7be5f3a7e4a8d276e7d28f0d73/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs#L143) it from the other 'rail'.

Since the `value` is not volatile here, if `hasLatest` has been set for a previous element, a partially setup/constructed object may be visible on the reader side. For example:

``` cs
class MyValue {
   int field;
}

MyValue v = new MyValue()
v.field = 5;
// other.onNext(v);
_parent._latest = v; // this is plain write
// <-- writer gets interrupted here
_parent._hasLatest = true;
```

Even if constructing `v` has a full barrier, it is possible `v.field = 5` gets reordered with `_parent._latest = v` and reading `_latest.field` on the other side may yield 0.

A solution would be to wrap `TSecond` into an object that is written via volatile write to the _latest field. Then, there is no need for `hasLatest` but check for `null` (or other special value) to see if there is one latest available or not.

akarnokd:
Related PR: #247.

danielcweber:
#247 was merged, closing.

PR #247, opened 2016-08-05
Fix WithLatestFrom potential memory visibility issues

akarnokd:
As I reported in #227, setting a volatile `hasLatest` doesn't ensure proper visibility of the latest value and one needs at least an ordered store for the whole element (by boxing it and writing the reference in with an ordered store).

**Cost of correctness**: one boxing and one unboxing of structure types.

**Possible tradeoff choice**: using `Volatile.Write` instead of `Interlocked.Exchange`; both are correct but the former may only become visible a bit delayed (once the CPU write buffer is emitted) whereas the latter is "immediately" visible to other cores.

### DIFF
--- a/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs
+++ b/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs
@@ -40,9 +40,10 @@ namespace System.Reactive.Linq.ObservableImpl
             }
 
             private object _gate;
-            private object _latestGate;
-            private volatile bool hasLatest;
+            private volatile bool _hasLatest;
             private TSecond _latest;
+
+            private object _latestGate;
 
             public IDisposable Run()
             {
@@ -89,7 +90,7 @@ namespace System.Reactive.Linq.ObservableImpl
 
                 public void OnNext(TFirst value)
                 {
-                    if (_parent.hasLatest)
+                    if (_parent._hasLatest)
                     {
 
                         TSecond latest;
@@ -156,9 +157,9 @@ namespace System.Reactive.Linq.ObservableImpl
                         _parent._latest = value;
                     }
 
-                    if (!_parent.hasLatest)
+                    if (!_parent._hasLatest)
                     {
-                        _parent.hasLatest = true;
+                        _parent._hasLatest = true;
                     }
                 }
             }
--- a/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs
+++ b/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs
@@ -5,6 +5,7 @@
 #if !NO_PERF
 using System;
 using System.Reactive.Disposables;
+using System.Threading;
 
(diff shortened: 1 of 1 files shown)


akarnokd:
```
=== TEST EXECUTION SUMMARY ===
   Tests.System.Reactive  Total: 3419, Errors: 0, Failed: 0, Skipped: 2, Time: 20.187s
```

Tests passed but then something failed.

bartdesmet:
Please rework to avoid the boxing. I'd prefer a plain old lock over boxing with arbitrary box lifetimes which solely depend on the incoming sequence's timing.

akarnokd:
@bartdesmet That may only be possible if I use locks; I don't know the cost of boxing but I'd assume its less than taking a lock.

bartdesmet:
@akarnkd - I'm not as much concerned about the cost of boxing than I am about the effects down the line. In the Bing service I'm working on now, we're very allergic to any boxing without prior knowledge about the eventual lifetime of the box: if it's known to be short-lived, it's ok (if the box can't be avoided at all); if it's known to be very long-lived (ideally infinitely long), it's ok to let it age to Gen-2 (if the box can't be avoided at all); if it's somewhere in the middle, we'll pay the cost during a GC down the line. In this particular case, we're dealing with lifetimes dependent on the velocity of the `TSecond` stream and the struct may be arbitrarily big (think the new `ValueTuple<...>` for C# 7.0 as a great source of N-ary composite objects).

My recommendation would be to use a `lock` (correctness above everything else), as we do in many places, and keep the costs in the control plane (`IDisposable`, cf. your other PR, thanks!) and on the input side (subjects) low by exploring lock free approaches. For high-performance scenarios, the best advice I'd have is to have many instances of single-threaded Rx executors with operators that don't do any locking at all. That's closer to what we do in Bing with single-threaded schedulers (N of which run in a process, where N is aligned with the number of cores) that merge events from many sources (a la what `Tx` does) onto a single event dispatch loop. It makes an assumption though about being able to control the threading and where the sources produce their messages (easier to do in a service where all the input comes from big pipes over the network). This would also work well for UI frameworks etc. that are single-threaded anyway.

akarnokd:
Okay. I'll change it to use a lock.

bartdesmet:
Cool, thanks! Do you still have issues with contribution agreement we have to look into?

akarnokd:
No, found another link and signed it (see cla-already-signed label).

akarnokd:
Updated the PR, not sure what takes so long for the CI (I get access denied is it may not even run yet).

Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/WithLatestFrom.cs
```diff
@@ -39,12 +40,14 @@ public _(WithLatestFrom<TFirst, TSecond, TResult> parent, IObserver<TResult> obs
             }
 
             private object _gate;
-            private volatile bool _hasLatest;
```
bartdesmet: Can you change the field name back to `_hasLatest` for consistency and to minimize the diff?

akarnokd:
Done.

shiftkey:
👍

onovotny:
@shiftkey @bartdesmet @mattpodwysocki is this good to merge?

Issue #194, opened 2016-06-18
Better support for async - Task based overloads for AsyncEnumerable


onovotny:
As mentioned in several Issues and PR's: #96, #170, #168, #144 and other threads.

We need to add async overloads to all of the query operators in `AsyncEnumerable` to support async callbacks, accumulators and the rest.

From an API design perspective, do we want to add overloads or create new `*Async` versions of the methods? Do the language integrated query keywords work with async such that we _need_ to maintain the current naming to work there?

onovotny:
We should probably wait until preview 4 of VS "15" for this based on @ljw1004's thoughts:
http://www.twitter.com/lwischik/status/755933011724476416

Seems like this will help perf here as it'll be an inner-loop. As `ValueTask` is in a public package, we can start exposing this before ppl can easily use it and once people use C# 7, it'll "just work."

ljw1004:
Agreed. If you're operating over a sequence, you should expose something like
```csharp
IAsyncEnumerable<U> SelectAsync<T,U>(this IEnumerable<T> seq, Func<T,ValueTask<U>> lambda)
```
with `ValueTask` not `Task`.

The reason is that presumably this lambda will be invoked once for each element in the sequence. It's typical for folks to write an async lambda which returns synchronously 99.9% of the time, and only shells out to async on the rare occasions that it needs to pull in more data. Using ValueTask will avoid one heap allocation per element on those 99.9% of times.


You should not offer two overloads
```csharp
IAsyncEnumerable<U> SelectAsync<T,U>(this IEnumerable<T> seq, Func<T,ValueTask<U>> lambda)
IAsyncEnumerable<U> SelectAsync<T,U>(this IEnumerable<T> seq, Func<T,Task<U>> lambda)
```
If you did, then someone who invokes `seq.SelectAsync(async (x) => {...; return x+1;})` in C#6 would get the Task overload, and in C#7 would get an ambiguity error.

onovotny:
@ljw1004 You bring up a good / different point about the surface area that we hadn't considered but probably should - async accumulators to `IEnumerable` that turn into `IAsyncEnumerable`.

Originally, I was thinking of 
```csharp
IAsyncEnumerable<U> SelectAsync<T,U>(this IAsyncEnumerable<T> seq, Func<T,ValueTask<U>> lambda)
```

But your use-case is an equally valid way to get an `IAsyncEnumerable` sequence from a "regular" `IEnumerable`

ljw1004:
[...]

PR #255, opened 2016-08-11
Add overload for AsyncEnumerable.Defer for delayed creation of inner enumerable.

danielcweber:
Adds an overload for AsyncEnumerable.Defer that takes a `Func<CancellationToken, Task<IAsyncEnumerable<T>>> to allow delayed creation of the inner IAsyncEnumerable.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/Defer.cs
+++ b/Ix.NET/Source/System.Interactive.Async/Defer.cs
@@ -2,9 +2,8 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information. 
 
-using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace System.Linq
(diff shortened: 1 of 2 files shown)


Ix.NET/Source/System.Interactive.Async/Defer.cs
```diff
@@ -20,5 +19,34 @@ public static IAsyncEnumerable<TSource> Defer<TSource>(Func<IAsyncEnumerable<TSo
...
+                {
+                    var baseEnumerator = default(IAsyncEnumerator<TSource>);
+
+                    return CreateEnumerator(
+                        async ct =>
+                        {
+                            if (baseEnumerator == null)
+                                baseEnumerator = (await asyncFactory(ct)).GetEnumerator();
+
+                            return await baseEnumerator.MoveNext(ct);
```
onovotny: ConfigureAwait(false)

Ix.NET/Source/System.Interactive.Async/Defer.cs
```diff
@@ -20,5 +19,34 @@ public static IAsyncEnumerable<TSource> Defer<TSource>(Func<IAsyncEnumerable<TSo
...
+            return CreateEnumerable(
+                () => 
+                {
+                    var baseEnumerator = default(IAsyncEnumerator<TSource>);
+
+                    return CreateEnumerator(
+                        async ct =>
+                        {
+                            if (baseEnumerator == null)
+                                baseEnumerator = (await asyncFactory(ct)).GetEnumerator();
```
onovotny: ConfigureAwait(false)

Ix.NET/Source/System.Interactive.Async/Defer.cs
danielcweber: Added.

Ix.NET/Source/System.Interactive.Async/Defer.cs
danielcweber: Added.

Ix.NET/Source/System.Interactive.Async/Defer.cs
```diff
@@ -20,5 +19,34 @@ public static IAsyncEnumerable<TSource> Defer<TSource>(Func<IAsyncEnumerable<TSo
                 () => factory()
                     .GetEnumerator());
         }
+
+        public static IAsyncEnumerable<TSource> Defer<TSource>(Func<CancellationToken, Task<IAsyncEnumerable<TSource>>> asyncFactory)
```
onovotny: I think this needs to wait until we implement `ValueTask` as part of C# 7

onovotny:
This overall is helpful, but is also part of adding async overloads to the larger surface area. I'd like to see this included as part of #194

danielcweber:
What's the deal with these seemingly random AppVeyor fails ?

onovotny:
I'm not sure what's going on....there appears to be a test in Rx that's intermittently hanging. Might be related to some GC as that's the only non-deterministic thing that I can think of.

I saw some weirdness iwth some Ix MBRO tests that I fixed by returning a null lifetime, I wonder if there's a similar issue with Rx. Need to dig into that. Tests always passed locally but sometimes failed on AppVeyor.

akarnokd:
Close & defer till C# 8?

Issue #242, opened 2016-07-31
MSDN page update


vbjay:
[MSDN Page ](https://msdn.microsoft.com/en-us/data/gg577610) still mentions older rx-\* nuget packages.

onovotny:
@martinwoodward @mattpodwysocki @bartdesmet, how do we update that page?

vbjay:
It would be nice if the specific packages were listed like this [page ]([url](http://getglimpse.com/Docs/#download))for glimpse.

EamonNerbonne:
The [RX.net repo's](https://github.com/Reactive-Extensions/Rx.NET) README.md doesn't list updated package names either.

How *do* you install rx, nowadays?

vbjay:
Yes it does.  https://github.com/Reactive-Extensions/Rx.NET#breaking-changes says https://www.nuget.org/packages/System.Reactive/

martinwoodward:
Let me go see about updating that page. In the meantime, would someone be able to make a first go at the page the equivalent page on our side?  I can then see about either getting the content updated to be a copy of the page we have here or ideally just make it redirect to GitHub for us.

martinwoodward:
Take it we would want this page updating too?  https://msdn.microsoft.com/en-us/data/gg577609

vbjay:
Yes please.

EamonNerbonne:
Ahh, to me those looked like namespaces at first (and subsequent) glances.

vbjay:
That is the whole point of their renaming.  The package now matches the
main namespace of the library.

akarnokd:
That page is pretty outdated but [this](https://docs.microsoft.com/en-us/previous-versions/dotnet/reactive-extensions/hh242986(v%3dvs.103)) looks like at least the assembly names are recent. Is this MSDN page within the control of this project even?

vbjay:
I do wish they would link to the GitHub repo.

akarnokd:
The site is no longer updated I'm afraid so we can't do much about that.

PR #258, opened 2016-08-12
Include links to nuget

EamonNerbonne:
Just in case I'm not the only one that was confused as I was in #242: link to nuget package page from package name.


EamonNerbonne:
If this kind of change is welcome and wanted: I think it might also be clearer to rephrase the "breaking change" section from the perspective of the present. I.e. name the section "Nuget packages" with a bold "breaking change" in the section body rather than title, and list the current package names before the packages they replace.

onovotny:
Thanks!

PR #263, opened 2016-08-17
Optimization for MoveNext

danielcweber:
Separate initialization logic from iteration logic. Avoids repeated patterns of code. Dispose the iterator when MoveNextCore returns false. Please review OrderedAsyncEnumerable on whether the changes preserve the semantics.


onovotny:
@bartdesmet @shiftkey thoughts about this split?

The only reason it is structured the way it is currently is that's how the pattern is done in CofeFX's System.Linq.

The only possible down-side to this, and it's a trade-off as with most optimizations, is that there's an extra await call, allocation (the await state machine), and method calls on the first move-next. The current design (without this PR) uses a `goto case`, which is an efficient jump in IL.

PR #274, opened 2016-08-30
Fix the logic of the Sample operator

bnayae:
When the source observable complete
the Sample immediately sending the last value
and not waiting for the periodic schedule event

This fix was sent before but I had some problem with the Git
so I had to close the request and do it again


danielcweber:
@bnayae I don't think the desired semantics of `Sample` is to emit the last value when the source completes. The overload that takes a sampler-observable definitely doesn't, and it would be a bug surprise if this overload would.

akarnokd:
Close?

PR #246, opened 2016-08-05
Lock free MultipleAssignmentDisposable

akarnokd:
This PR removes the locking from `MultipleAssignmentDisposable` and uses "plain" Volatile/Interlocked actions.

PR #275, opened 2016-09-02
Lockfree SerialDisposable

LeeCampbell:
After a review of #241, @akarnokd identified a race condition error in the implementation.

This PR follows his guidance to take from https://github.com/reactor/reactor-core-dotnet/blob/master/Reactor.Core/util/DisposableHelper.cs the pattern to implement a lock free `SerialDispoable`. It also would bring the `SerialDisposable` inline with the `MultipleAssignmentDisposable` changes from #246 

Initial tests indicate there is a large improvement in Single Threaded performance, but a reduction in multi threaded performance. Results and benchmark code to follow.

I would like to benchmark how this change would affect the Rx lib at a higher level, i.e. the performance impact on the classes that use `SerialDisposable`

### DIFF
--- a/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
+++ b/Rx.NET/Source/System.Reactive.Core/Reactive/Disposables/SerialDisposable.cs
@@ -46,7 +46,7 @@ namespace System.Reactive.Disposables
                 // Don't leak the DISPOSED sentinel
                 if (a == BooleanDisposable.True)
                 {
-                    a = DefaultDisposable.Instance;
+                    a = null;
                 }
                 return a;
             }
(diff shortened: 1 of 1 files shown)


akarnokd:
You might want to adjust the [test](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs#L491) or return null instead of the default disposable. Strangely, the MultipleAssignmentDisposable didn't have this null requirement on a disposed instance so not sure how strong the given requirement is.

LeeCampbell:
I will mod the code to keep the same behaviour. I am struggling to get the tests running locally, or the build scripts to fail (when tests fail). I will update https://github.com/Reactive-Extensions/Rx.NET/wiki/Building%20Testing%20and%20Debugging when I figure out how to play ball. However I think I will need to spin up a few VMs to figure out what is actually required to build, vs, what I happen to have installed ;-)

akarnokd:
This optimization could go in before the roadmap.

@LeeCampbell Do you have the time to resolve the conflict and repost these changes?

LeeCampbell:
Sure thing. I will re-run the benchmarks and include them here too (else it is a pretty poor PR!)

akarnokd:
To be replaced by #535 (commits were preserved).

onovotny:
Merged as part of #535

PR #293, opened 2016-11-10
Reduce number of allocations (closures, delegates) in TaskObservableExtensions.ToTask

danielcweber:
We use the ToTask-extension a lot in our code and found it rather expensive wrt. object allocations. Looking into the code, I found that by implementing a custom IObserver, creation of closures and delegates can be avoided. I think even without anonymous lambdas, the code is still straghtforward and easy to read.


shiftkey:
@danielcweber this looks great! :sparkles: Introducing more of these into the Rx side to reduce allocations seems like a good idea for the next minor release.

PR #313, opened 2017-01-09
When converting an IAsyncEnumerable to an IObservable, unwrap the AggregateException that may be thrown by IAsyncEnumerable.MoveNext.

danielcweber:
This may be considered a breaking change. When converting an IAsyncEnumerable into an IObservable, AggregateExceptions thrown by IAsyncEnumerator.MoveNext should be unwrapped before being passed to OnError. Justifications:

- A user of the resulting IObservable cannot anticipate any AggregateExceptions, much less the level of nesting when repeatedly converting.
- If ToObservable would be implemented using async/await, there would be no AggregateExceptions.
- Since the async/await-feature, AggregateExceptions have become less useful. In fact, async state machines also only rethrow the first exception in an AggregateException.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/ToObservable.cs
+++ b/Ix.NET/Source/System.Interactive.Async/ToObservable.cs
@@ -190,7 +190,7 @@ namespace System.Linq
                                          {
                                              if (t.IsFaulted)
                                              {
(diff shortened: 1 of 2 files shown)


akarnokd:
Close & defer till C# 8?

Issue #319, opened 2017-01-24
Cancellation in Ix.Async introduces subtle race conditions


danielcweber:
Currently, `AsyncIterator`, the base class of most implementations of `IAsyncEnumerable` [registers a callback](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs#L87) in its `MoveNext` method on the `CancellationToken` that was passed in.

Overrides of `Dispose` regularly dispose and nullify inner instances of `IAsyncEnumerable`, [e.g. in the TakeWhile operator](https://github.com/Reactive-Extensions/Rx.NET/blob/b84974db5d4940b88308b98723893cce9f10f386/Ix.NET/Source/System.Interactive.Async/Take.cs#L87).

This introduces race conditions. While `MoveNextCore` might observe the current iterator state to be ` AsyncIteratorState.Iterating`, it may be cancelled at the very next moment, throwing a `NullReferenceException` ([e.g. here](https://github.com/Reactive-Extensions/Rx.NET/blob/b84974db5d4940b88308b98723893cce9f10f386/Ix.NET/Source/System.Interactive.Async/Take.cs#L108)). I've been runnng into those repeatedly since upgrading to v3.

There might be conflicting views on this. One might state that, when you cancel an async iterator, all bets are off. I believe that the only thing that should be thrown are `TaskCanceledExceptions`, and such things as `NullReferenceExceptions` cannot be anticipated or handled well. As a result, you cannot write good async software that properly supports cancellation.

A possible workaround might be avoid the `CancellationToken` callback and check `IsCancellationRequested` in `MoveNext` before and after the call to `MoveNextCore`, disposing the iterator if necessary. There might, however, be code around that relies on the iterator being disposed immediately.

Any thoughts on this?

onovotny:
Do you have a repro of this? Are the iterator and `Dispose` being called on different threads?

The iterator itself is not thread-safe, and was never intended to be so far. That said, I agree we should never be throwing `NullReferenceException`'s. Once `Dispose` is called, `Current` will be `default(T)`, so essentially useless. In the case you highlighted, the fix could be just to do `enumerator?.Current` to protect against the condition you describe, but I'd like to understand how you got there if the iterator was only "touched" by a single thread?

>There might, however, be code around that relies on the iterator being disposed immediately.
I'm fairly sure there is....but I'd have to check.

danielcweber:
[...]

PR #320, opened 2017-01-25
Eat exceptions in the MoveNext that come as a result of a cancelation

onovotny:
Fix for #319

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
@@ -90,7 +90,7 @@ namespace System.Linq
                     try
                     {
                         // Short circuit and don't even call MoveNexCore
-                        cancellationToken.ThrowIfCancellationRequested();
+                        cts.Token.ThrowIfCancellationRequested();
 
                         var result = await MoveNextCore(cts.Token)
                                          .ConfigureAwait(false);
@@ -101,8 +101,16 @@ namespace System.Linq
                     }
                     catch
                     {
+                        // preserve this before calling Dispose, which will always set IsCancellationRequested
+                        var cxlRequested = cts.IsCancellationRequested;
+
                         currentIsInvalid = true;
                         Dispose();
+
+                        // throw a cxl if it was from an incoming request
+                        if(cxlRequested)
+                            cts.Token.ThrowIfCancellationRequested();
+
                         throw;
                     }
                 }


danielcweber:
I'm not convinced. This would swallow the `NullReferenceException`described in #319, but that would mean masking a bug. Moreover, as long as Dispose can be called while `MoveNextCore` is running, you cannot easily avoid those exceptions (unless every operator locks its `Dispose` override). They will be thrown anyway, which might also be a performance issue.

onovotny:
I'm not sure that there's any way to properly synchronize/clean up these iterators -- or that we really care about what happens after a cancel, long as Dispose is called to clean up the chain.

I'm not sure putting a lock around `MoveNextCore` and `Dispose` to prevent this overlap is a good idea either due to perf. Further, `Dispose` today acts like a Cancel, so it actually has to be able to be called while in `MoveNextCore` or that's a breaking change.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -101,8 +101,16 @@ public TSource Current
                     }
                     catch
                     {
+                        // capture here as Dispose will set
```
shiftkey: Is this comment missing something after `set`?

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -90,7 +90,7 @@ public TSource Current
                     try
                     {
                         // Short circuit and don't even call MoveNexCore
-                        cancellationToken.ThrowIfCancellationRequested();
+                        cts.Token.ThrowIfCancellationRequested();
```
shiftkey: Yep, this feels more correct

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
onovotny: I can clarify -- it's more that Dispose sets "IsCancelRequest" to true always since Dispose calls Cancel. So we need to capture the state before we call dispose.

shiftkey: Aha, so what about something like
```
// preserve this before calling Dispose, which will always set IsCancellationRequested
```

danielcweber:
What if `Dispose` would continue to act as a Cancel (which is IMO good and intended behaviour) , but a cancellation would not dispose the enumerator right away?

I had a look at some IAsyncEnumerable-consuming methods in Ix and my code. The usual pattern is to guard IAsyncEnumerator `e` by a using block, and passing any outer `CancellationToken` to `MoveNext`. This pattern does not dispose `e` when `ct` is cancelled, it does so only after `MoveNext` throws (or completes).

Second, there is a somewhat circular behavior that doesn't feel right. `MoveNext` registers on `ct` and calls `Dispose` on cancellation, but `Dispose` cancels its own `CancellationTokenSource`. The usage pattern described above is not reflected here.

I see that automatically disposing the enumerator on cancellation protects against wrong usage of the enumerator, i.e. not wrapping `e` in a `using` block etc. Do we really have to make such misleading guarantees at the risk of throwing strange `NullReferenceExceptions`? Does it solve dangerous potential memory leaks? I think the reality is worse: The avove approach still frees the user of taking care of the enumerator properly, but does so by hiding possible bugs in Ix.

I suggest not registering on `ct` in `MoveNext`. Instead `MoveNext` should check `ct` before and after `MoveNextCore` and catch `TaskCanceledExceptions`, disposing the instance only at safe points. Still, `Dispose` has to cancel its inner `CancellationTokenSource`.

danielcweber:
I checked into this since we're having lots of stray (unobserved task-)exceptions that we cannot really get rid of and is polluting our logs. We work a lot with cancellation so this is important to us.

For a test, I uncommented [this](https://github.com/Reactive-Extensions/Rx.NET/blob/develop/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs#L87) on the develop branch. The tests are still green. Should `MoveNextCore` throw a `TaskCancelledException`, this will be caught [here](https://github.com/Reactive-Extensions/Rx.NET/blob/develop/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs#L102) and the iterator will be disposed. So now we have a safe point at where to dispose the iterator and will never do so in beetween calls to `MoveNextCore`.

What will be broken by the proposed changed behaviour?
- When the local `cancellationToken` in `MoveNext` is cancelled, the field `cancellationTokenSource` will not be cancelled right away, but this is not observable to `MoveNextCore` - it only sees the combined token anyway.
- `MoveNextCore` overrides might not observe the cancellationToken and possibly run forever - which I would consider a bug. Disposing the iterator however will not help stopping `MoveNextCore` either.
- `MoveNextCore` might complete without throwing. A check on `cancellationToken` just before [this line](https://github.com/Reactive-Extensions/Rx.NET/blob/develop/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs#L98) can dispose the iterator at a safe point.

Please also note my points from above, I still think they are valid: If you consume your async iterator in a while loop, you usually have the enumerator in a `using` block, only disposing it when `MoveNext` throws (or returns `false`) but never in betwwen. We should reflect this pattern.

I'll happily provide a pull request.

onovotny:
I'm still not sure....I don't think that behavior is correct. What were the issues from my proposed PR, which fixes it a different way?

danielcweber:
No issues with the PR per se, it's only that eating all exceptions is generally bad style. Moreover, they will still be thrown, which means a stack walk every time and whatnot. And you have no way of circumventing these exceptions. If you will, they are not even exceptional if they happen as part of normal design. Exceptions must be exceptional.

I cannot say that the resulting behavior is correct or not - IMO, I do not expect an object to dispose itself when I cancel one of its methods. But consumers of Ix.Async might see it different, however, there's no test asserting this particular behavior.

danielcweber:
Opened a pull request to outline my idea and for further discussions. Sorry for nagging. We really cannot work around this issue in our own code alone.

danielcweber:
@onovotny Since C# 8's async iterators feature will deal with cancellation in a whole different way, this PR is obsolete and can be closed.

PR #325, opened 2017-02-10
EventLoopScheduler: set _nextItem to null on Tick

TobBrandt-Work:
This fixes a potential memory leak. 

If _queue is empty after the item is removed from it, then _nextItem is never overriden and therefore keeps a reference to item.
Until another action is scheduled with a future due time, the item and all its associated state is kept in memory.


bartdesmet:
@TobBrandt-Work, thanks for your contribution. We're looking into adding this to Rx 4.0 but there seems to be a merge conflict. If you have time to help to resolve this, feel free to do so. If you don't, that's fine too, and we'll have a look over here. The CI issue should have been resolved by now, so once the merge conflict goes away, I'd expect everything to turn green.

TobBrandt-Work:
I have merged in the changes from upstream.

onovotny:
@TobBrandt-Work Can you please rebase this off of `develop`? It should not be going into master.

Thanks

TobBrandt-Work:
I cannot change the source branch of the PR (at least I don't know how).

New PR: https://github.com/Reactive-Extensions/Rx.NET/pull/421

PR #329, opened 2017-02-21
fix removing elements other than top

astrofishgames:


akarnokd:
Honestly I don't know how PriorityQueue should behave or how it is working, but if this PR intends to fix a bug, there should be an unit test verifying the fix and also showing the bug in the current version.

danielcweber:
@astrofishgames Could you give us some insights of what this PR fixes?

danielcweber:
I reproduced the bug and rebased the fix from this PR onto the lastest master, closing this in favor of https://github.com/dotnet/reactive/pull/569/commits.

PR #340, opened 2017-03-31
WIP Fix #319

### LINKED ISSUE #319: Cancellation in Ix.Async introduces subtle race conditions

danielcweber:
Alternative to #320.

This fix will restore the behavior of `AnonymousAsyncEnumerator<T>` in Rx 2.x which did not observe the `CancellationToken` passed to `MoveNext` and subsequently did not dispose the enumerator on cancellation.

While #320 catches all exceptions occuring due to races (refer to #319 for details), this fix will hopefully prevent such races.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
@@ -84,7 +84,6 @@ namespace System.Linq
                     return false;
                 }
 
-                using (cancellationToken.Register(Dispose))
                 using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cancellationTokenSource.Token))
                 {
                     try
@@ -95,6 +94,7 @@ namespace System.Linq
                         var result = await MoveNextCore(cts.Token)
                                          .ConfigureAwait(false);
 
+                        cancellationToken.ThrowIfCancellationRequested();
                         currentIsInvalid = !result; // if move next is false, invalid otherwise valid
 
                         return result;


danielcweber:
Tried to rebase the changes onto the master branch for a possible 3.x release but the build system has already moved to 4.0 I guess.

onovotny:
Tagging @bartdesmet @shiftkey @mattpodwysocki As per the discussion in the issue, I'm not sure if this is the correct behavior or leads to other bugs/memory leaks. Is this or #320 a better fix?

danielcweber:
@onovotny Since C# 8's async iterators feature will deal with cancellation in a whole different way, this PR is obsolete and can be closed.

danielcweber:
Since C# 8's async iterators feature will deal with cancellation in a whole different way, this PR is obsolete and can be closed.

PR #343, opened 2017-04-13
Initial implementation of C# 7.0 task-like support for observables

bartdesmet:
This PR contains an initial implementation of task-like support for `IObservable<T>`. Given that we can't put the required `AsyncMethodBuilder` attribute on the interface in the BCL, we have to use an alternative type, so I'm introducing `ITaskObservable<T>` (obviously derived from `IObservable<T>`) which both complies with the task-like requirements and exposes the awaiter pattern. The reason an interface is used is to enable usage in a covariant manner on interfaces and delegate types.

Example:
```csharp
async ITaskObservable<int> RxAsync()
{ 
    var res = await Observable.Return(21).Delay(TimeSpan.FromSeconds(1)); 
    return res * 2; 
}
```


RxDave:
How about also refactoring all of the operators that return singleton observables to return `ITaskObservable<T>` instead? E.g., FirstAsync, LastAsync, Aggregates, etc.

I remember when I was teaching people how to use Rx a few years back (pre-async methods), one of the most common complaints was the ambiguity between an `IObservable<T>` that is known to only contain a single value vs. an instance that contains many values, and I never had a satisfactory explanation for this. Async methods kind of made it a moot point, but do you think it would still be worth making the change?

mikehaas763:
With the goal being that you can await on an observable in order to get the first item in the stream?

bartdesmet:
@RxDave - This is something we may consider. I'll give it some more thought.
@mikehaas763 - Await support for observables already exists in Rx; this adds support to return an observable from an asynchronous method, with the goal to make it easier to e.g. implement interfaces that have an observable-based return type and want to do away with observable/task converters when implementing such methods using `async`.

mikehaas763:
I see cool cool

PR #397, opened 2017-04-26
Reduction of producer/sink overheads

bartdesmet:
This mirrors some of the design points we had for the Reactor service here at Bing. With previous PRs, I've detached the `IObservable<T>` "plane" from the underlying observer "plane" by reducing the "parent" reference to only the things needed by the hot observers/sinks.

In this change, the `Run` method of producers has been changed to eliminate the need for a `setSink` callback, pushing some wiring to the base class. This reduces delegate allocations. Furthermore, the `IDisposable` objects referenced by a subscription have been reduced in terms of indirection levels (1 stable composite + 2 single assignment to 1 specialized + 1 single assignment).

More stuff may follow later after some more design thoughts over here. The event processing pipeline in Reactor has some different design points compared to Rx (in particular, allowing for traversal of the dynamically evolving runtime operator graph at any point of its execution to facility things beyond disposal, e.g. checkpoint and recovery), so some work is needed in converging the two.


Rx.NET/Source/Directory.build.props
```diff
@@ -6,7 +6,7 @@
     <MinClientVersion>2.12</MinClientVersion>
     <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
     <Authors>.NET Foundation and Contributors</Authors>
-    <UseFullSemVerForNuGet>true</UseFullSemVerForNuGet>
+    <UseFullSemVerForNuGet>false</UseFullSemVerForNuGet>
```
onovotny: Why disable this? VS 15.3 supports this as does VS 2015 with NuGet 3.6. This is also only for preview builds.

bartdesmet: Not sure; just synced back to `master` and messed up things.

onovotny: Hmm...this looks like a weird merge.

onovotny:
@bartdesmet, `master` shouldn't be merged into your dev branch, that should be `develop`. `master` is 3.x. `develop` is the 4.x stuff and has a lot of changes from `master`. Can you try a hard reset of your branch to before https://github.com/Reactive-Extensions/Rx.NET/pull/397/commits/b55957f50cbcfb444e8c58f318967349ec7cb48a and then merge/rebase off of `develop`?

bartdesmet:
Yeah have been using the same commands as everywhere else internally or .NET Core, coming from `master` rather than `develop`. Fixed now.

onovotny:
Cool. I've been keeping work in develop so that we could potentially easily do hotfixes from master. That said, I have no objection to merging this stuff into master if we went to develop from there instead.

PR #402, opened 2017-05-30
Switch to use Nerdbank.Gitversioning

onovotny:
This task supports two features we need:

1. different versions per project, Ix and Rx.
2. supports dotnet cli and removes the msbuild requirement.

### DIFF
--- a/Ix.NET/Source/Directory.build.props
+++ b/Ix.NET/Source/Directory.build.props
@@ -6,7 +6,6 @@
     <MinClientVersion>2.12</MinClientVersion>
     <GenerateDocumentationFile Condition=" '$(Configuration)' == 'Release' ">true</GenerateDocumentationFile>
     <Authors>.NET Foundation and Contributors</Authors>
-    <UseFullSemVerForNuGet>true</UseFullSemVerForNuGet>
     <PackageIconUrl>http://go.microsoft.com/fwlink/?LinkId=261274</PackageIconUrl>
     <PackageProjectUrl>http://go.microsoft.com/fwlink/?LinkId=261273</PackageProjectUrl>
     <PackageLicenseUrl>https://raw.githubusercontent.com/Reactive-Extensions/Rx.NET/master/Ix.NET/Source/license.txt</PackageLicenseUrl>
@@ -15,16 +14,9 @@
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)ReactiveX.snk</AssemblyOriginatorKeyFile>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <DefaultLanguage>en-US</DefaultLanguage>
-    <WriteVersionInfoToBuildLog>false</WriteVersionInfoToBuildLog>
     <IncludeSymbols>false</IncludeSymbols>
     <IsTestProject>$(MSBuildProjectName.Contains('Test'))</IsTestProject>
-    
-    <GetVersion Condition=" '$(NCrunch)' != '' ">false</GetVersion>
   </PropertyGroup>
-  
-  <ItemGroup>
-    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0011" PrivateAssets="All" /> 
-  </ItemGroup>
   
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(NCrunch)' == '' and '$(SourceLinkEnabled)' != 'false'">
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.0" PrivateAssets="All" /> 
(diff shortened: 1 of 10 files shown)


Ix.NET/Source/Directory.build.props
```diff
@@ -15,34 +14,32 @@
...
   <Target Name="GetPackagingOutputs" />
-   <PropertyGroup>
-    <GitVersionTaskVersion>4.0.0-beta0011</GitVersionTaskVersion>
+  
+  <PropertyGroup>
+    <NerdbankGitVersioningVersion>2.0.3-beta-g4d80666c50</NerdbankGitVersioningVersion>
   </PropertyGroup>
+  
+  <ItemGroup>
+    <PackageReference Include="Nerdbank.GitVersioning" Version="2.0.3-beta-g4d80666c50" PrivateAssets="all" />
```
AArnott: Did you mean to use `$(NerdbankGitVersioningVersion)` for the `Version` attribute here?

Ix.NET/Source/build-new.ps1
```diff
@@ -30,16 +30,17 @@ if (!(Test-Path .\nuget.exe)) {
 # get tools
 .\nuget.exe install -excludeversion SignClient -Version 0.7.0 -outputdirectory packages
 .\nuget.exe install -excludeversion JetBrains.dotCover.CommandLineTools -pre -outputdirectory packages
-.\nuget.exe install -excludeversion gitversion.commandline -pre -outputdirectory packages
+.\nuget.exe install -excludeversion Nerdbank.GitVersioning -Version 2.0.3-beta-g4d80666c50 -pre -outputdirectory packages
 .\nuget.exe install -excludeversion xunit.runner.console -pre -outputdirectory packages
 .\nuget.exe install -excludeversion ReportGenerator -outputdirectory packages
 #.\nuget.exe install -excludeversion coveralls.io -outputdirectory packages
 .\nuget.exe install -excludeversion coveralls.io.dotcover -outputdirectory packages
```
AArnott: tip: you can drop `-outputdirectory packages` on all these steps if you add the equivalent setting to your `nuget.config` file:
```xml
  <config>
    <add key="repositorypath" value="packages" />
  </config>
```

Ix.NET/Source/NuGet.Config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <add key="xUnit CI" value="https://www.myget.org/F/xunit/api/v3/index.json" />
+    <add key="NBGV CI" value="https://ci.appveyor.com/nuget/nerdbank-gitversioning" />
```
AArnott: I see you also added this to your root level nuget.config. Why do you also need to add it here?

Rx.NET/Source/NuGet.Config
```diff
@@ -2,8 +2,8 @@
 <configuration>
   <packageSources>
     <add key="xUnit CI" value="https://www.myget.org/F/xunit/api/v3/index.json" />
+    <add key="NBGV CI" value="https://ci.appveyor.com/nuget/nerdbank-gitversioning" />
```
AArnott: I see you also added this to your root level nuget.config. Why do you also need to add it here?

Rx.NET/Source/Directory.build.props
```diff
@@ -34,17 +29,21 @@
   <!-- Workaround -->
   <Target Name="GetPackagingOutputs" />
   
-  <!-- https://github.com/NuGet/Home/issues/4337 -->
- <PropertyGroup>
-    <GitVersionTaskVersion>4.0.0-beta0011</GitVersionTaskVersion>
+  <PropertyGroup>
+    <NerdbankGitVersioningVersion>2.0.3-beta-g4d80666c50</NerdbankGitVersioningVersion>
```
AArnott: Why not define this property (and perhaps even the `PackageReference` and your `FixUpVersion` target) in a top-level Directory.Build.props file that all projects share? I see you define this in multiple Directory.Build.props files.
You can have multiple directory levels of Directory.Build.props files by adding this line to all but the top-most one:
```xml
  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" Condition=" '$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))' != '' " />
```

Ix.NET/Source/Directory.build.props
onovotny: Can't, see https://github.com/NuGet/Home/issues/5295

Ix.NET/Source/NuGet.Config
onovotny: The build was failing otherwise since with AppVeyor the working directory is root and it is invoking the ps1 scripts.

Rx.NET/Source/NuGet.Config
onovotny: Shouldn't be needed, will clean all those nuget configs up eventually :)

Rx.NET/Source/Directory.build.props
onovotny: We are effectively trying to keep Ix/Rx separate despite living in the same repo, that's why...

PR #409, opened 2017-07-21
Make NuGet badges link to NuGet

j2ghz:
The NuGet version badges now link to NuGet instead of opening the image in new tab


onovotny:
Thanks!

PR #418, opened 2017-08-25
bring back standalone System.Reactive.Interfaces package

shiftkey:
This is the manual work to revive the `System.Reactive.Interfaces` package, as we've found scenarios where people wanted a reference assemblies with interfaces only. I've used the typeforwarders contents as a starting point and moved those across. 

 - [ ] actually create the package again
 - [ ] pin the version to `4.0.0`


shiftkey:
@onovotny the CI links are "access denied" for me - can you see them?

bartdesmet:
Thanks for giving this a go! If we decide we have to proceed this way - awaiting some analysis outcome on my end, as discussed offline - it will be good to ensure that the interfaces assembly's version can remain locked. That was the original design point, but ports of repositories and build systems lost it (IIRC in the Microsoft OpenTech days predating the .NET Foundation move), so we ended up with this stable assembly having its version number evolve (resulting in the need for version redirects even in places that ought to have been stable).

onovotny:
If only Assembly Neutral Interfaces were a thing...

onovotny:
Weird test failure:

https://ci.appveyor.com/project/dotnetfoundation/rx-net-la3iv/build/4.0.0-preview.2.build.33+g95d3864e13#L5798

Not even a test, process not shutting down cleanly:
```
System.ComponentModel.Win32Exception (0x80004005): Invalid window handle

   at MS.Win32.UnsafeNativeMethods.PostMessage(HandleRef hwnd, WindowMessage msg, IntPtr wparam, IntPtr lparam)

   at MS.Win32.ManagedWndProcTracker.HookUpDefWindowProc(IntPtr hwnd)

   at MS.Win32.ManagedWndProcTracker.OnAppDomainProcessExit()

   at MS.Win32.ManagedWndProcTracker.ManagedWndProcTrackerShutDownListener.OnShutDown(Object target, Object sender, EventArgs e)

   at MS.Internal.ShutDownListener.HandleShutDown(Object sender, EventArgs e)
```

Rx.NET/Source/src/System.Reactive.Interfaces/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+using System.Resources;
+using System.Runtime.InteropServices;
+using System.Security;
+
+[assembly: ComVisible(false)]
+[assembly: CLSCompliant(true)]
+[assembly: AllowPartiallyTrustedCallers]
+[assembly: NeutralResourcesLanguage("en-US")]
```
onovotny: Should this file exist? At the very least, seems like this should be a common/shared file with System.Reactive?

shiftkey:
Still seeing the same test errors: https://ci.appveyor.com/project/dotnetfoundation/rx-net-la3iv/build/4.0.0-preview.2.build.33+g95d3864e13#L5798

Also not sure why the CI badge isn't being applied to this PR to indicate that it's not ready to merge 🤔🤔🤔🤔.

shiftkey:
@onovotny two things to go:

 - I don't seem to be able to get into the AppVeyor portal to enable PR integration again - are you able to check that?
 - Anything else to add about the PR?

shiftkey:
I think the `System.Reactive` nuspec isn't quite right - that should have the specific version for `System.Reactive.Interfaces` embedded in:



I've probably overlooked something, but I don't think this is ready to merge currently.

onovotny:
@shiftkey will give this a look-over during the next week. We need to be using the latest msbuild and nerdbank.giversioning for the numbers to properly align (bug fixes in the tools).

shiftkey:
@onovotny yeah, I might follow up after this PR to tidy up our scripts a bit and see what can be simplified...

PR #419, opened 2017-08-27
Setting up a minimal configuration for Rx.Next work

bartdesmet:
No real content yet, but noodling around with Visual Studio 2017 15.3 Preview 7.1 to try to set up a minimal working solution and project layout to build .NET Core compatible product binaries with packages that have inter-dependencies, topping it all off with an attempt to have CI working.


bartdesmet:
Moving this work to https://github.com/Reactive-Extensions/Reactor/pull/2 instead.

PR #436, opened 2017-10-01
Add a new 'partition' operator to split a sequence based on a given predicate

spencercw:
Inspired by the equivalents from F# and RxJS, this is a new Linq operator which allows a sequence to be split in two based on some predicate. This is useful as an alternative to `Where` in case you are interested in both the true and false branches and want to process them independently.

Example usage:

```
var (evens, odds) = observable.Partition(x => x % 2 == 0);
evens.Subscribe(x => Console.WriteLine("Even: " + x));
odds.Subscribe(x => Console.WriteLine("Odd: " + x));
```

This is roughly equivalent to the below. The benefits being that the sequence is only evaluated once (so no need to convert the sequence into a connectable), and the predicate only needs to be evaluated once for each element.

```
var numbers = observable.Publish();
numbers.Where(x => x % 2 == 0).Subscribe(x => Console.WriteLine("Even: " + x));
numbers.Where(x => x % 2 != 0).Subscribe(x => Console.WriteLine("Odd: " + x));
numbers.Connect();
```

I think I need to do something to `Qbservable`, but I'm not sure what exactly. The HomoIcon tool seems to be a bit broken, at least on develop. The `Qbservable_Observable_Parity` test is currently failing because `Partition` is missing.

This is my first time trying to do anything in the Rx.NET codebase so there's a decent chance I've messed it up one way or another. Any feedback would be welcome.

### DIFF
--- a/Rx.NET/Source/src/Microsoft.Reactive.Testing/TestScheduler.cs
+++ b/Rx.NET/Source/src/Microsoft.Reactive.Testing/TestScheduler.cs
@@ -63,6 +63,64 @@ namespace Microsoft.Reactive.Testing
         }
 
         /// <summary>
+        /// Schedules the given sequence. Uses the specified virtual times to invoke the factory function, subscribe to the resulting sequence, and dispose the subscription.
+        /// </summary>
+        /// <typeparam name="T">The element type of the observable sequence being tested.</typeparam>
+        /// <param name="create">Factory method to create an observable sequence.</param>
+        /// <param name="created">Virtual time at which to invoke the factory to create an observable sequence.</param>
+        /// <param name="subscribed">Virtual time at which to subscribe to the created observable sequence.</param>
+        /// <param name="disposed">Virtual time at which to dispose the subscription.</param>
+        /// <returns>Observer with timestamped recordings of notification messages that were received during the virtual time window when the subscription to the source sequence was active.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="create"/> is null.</exception>
+        public ITestableObserver<T> Schedule<T>(Func<IObservable<T>> create, long created, long subscribed, long disposed)
+        {
+            if (create == null)
+                throw new ArgumentNullException(nameof(create));
+
+            var source = default(IObservable<T>);
+            var subscription = default(IDisposable);
+            var observer = CreateObserver<T>();
+
+            ScheduleAbsolute(default(object), created, (scheduler, state) => { source = create(); return Disposable.Empty; });
+            ScheduleAbsolute(default(object), subscribed, (scheduler, state) => { subscription = source.Subscribe(observer); return Disposable.Empty; });
+            ScheduleAbsolute(default(object), disposed, (scheduler, state) => { subscription.Dispose(); return Disposable.Empty; });
+
+            return observer;
+        }
+
+        /// <summary>
+        /// Schedules the given sequence. Uses the specified virtual time to dispose the subscription to the sequence obtained through the factory function.
+        /// Default virtual times are used for <see cref="ReactiveTest.Created">factory invocation</see> and <see cref="ReactiveTest.Subscribed">sequence subscription</see>.
+        /// </summary>
+        /// <typeparam name="T">The element type of the observable sequence being tested.</typeparam>
+        /// <param name="create">Factory method to create an observable sequence.</param>
+        /// <param name="disposed">Virtual time at which to dispose the subscription.</param>
+        /// <returns>Observer with timestamped recordings of notification messages that were received during the virtual time window when the subscription to the source sequence was active.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="create"/> is null.</exception>
+        public ITestableObserver<T> Schedule<T>(Func<IObservable<T>> create, long disposed)
+        {
+            if (create == null)
+                throw new ArgumentNullException(nameof(create));
+
+            return Schedule(create, ReactiveTest.Created, ReactiveTest.Subscribed, disposed);
+        }
+
+        /// <summary>
+        /// Schedules the given sequence. Uses default virtual times to <see cref="ReactiveTest.Created">invoke the factory function</see>, to <see cref="ReactiveTest.Subscribed">subscribe to the resulting sequence</see>, and to <see cref="ReactiveTest.Disposed">dispose the subscription</see>.
+        /// </summary>
+        /// <typeparam name="T">The element type of the observable sequence being tested.</typeparam>
+        /// <param name="create">Factory method to create an observable sequence.</param>
+        /// <returns>Observer with timestamped recordings of notification messages that were received during the virtual time window when the subscription to the source sequence was active.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="create"/> is null.</exception>
+        public ITestableObserver<T> Schedule<T>(Func<IObservable<T>> create)
+        {
+            if (create == null)
+                throw new ArgumentNullException(nameof(create));
+
+            return Schedule(create, ReactiveTest.Created, ReactiveTest.Subscribed, ReactiveTest.Disposed);
+        }
+
+        /// <summary>
         /// Starts the test scheduler and uses the specified virtual times to invoke the factory function, subscribe to the resulting sequence, and dispose the subscription.
         /// </summary>
         /// <typeparam name="T">The element type of the observable sequence being tested.</typeparam>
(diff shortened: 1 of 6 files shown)


AviAvni:
You can use GroupBy

spencercw:
@AviAvni Yes, the example is roughly equivalent to this:

```
var groups = observable.GroupBy(x => x % 2 == 0).Publish();
var evens = groups.Where(g => g.Key).Merge();
var odds = groups.Where(g => !g.Key).Merge();
evens.Subscribe(x => Console.WriteLine("Even: " + x));
odds.Subscribe(x => Console.WriteLine("Odd: " + x));
groups.Connect();
```

I think the meaning is much clearer with `Partition` and doesn't require messing about with connectables.

AviAvni:
You don't need the connect in the group you can subscribe to it and then subscribe to each observable and conditionally print what you want

spencercw:
I need to be able to return the separate sequences rather than subscribing to the `IGroupedObservable` and handling it inline. Here's an example. I'm struggling to come up with anything like as clean as this by using `GroupBy`.

```
public static (IObservable<string>, IObservable<double>) Run(IObservable<int> input)
{
    var (evens, odds) = input.Partition(x => x % 2 == 0);
    return (
        evens.Select(x => x.ToString()),
        odds.Select(x => x / 2.0));
}
```

RxDave:
`Publish` is generally fine for this kind of thing. So if your primary concern is that the predicate is being evaluated twice, then simply publish the predicate.

    public static (IObservable<string>, IObservable<double>) Run(IObservable<int> input)
    {
      var pairs = input.Select(x => (X: x, Div2: x % 2 == 0)).Publish();
      var evens = pairs.Where(p => p.Div2).Select(p => p.X);
      var odds = pairs.Where(p => !p.Div2).Select(p => p.X);
      return (
          evens.Select(x => x.ToString()),
          odds.Select(x => x / 2.0));
    }

spencercw:
Where does the `Connect` go? If I call `RefCount` then events will start flowing as soon as one of the output sequences is subscribed to, which can mean events bound for the second sequence can be missed. If I call `Connect` anywhere in that method then events start flowing immediately, and those will be lost because nothing has subscribed to either of the output sequences yet.

RxDave:
Thanks for noticing, I've corrected the example to account for the missing call to `Connect`!

You must flow the connectability to the caller, if it matters in your case. Otherwise, you can just call `Connect` before returning.

The problem is that ultimately when splitting a sequence into more than one sequence, you're implicitly changing a potentially _cold_ observable into a _hot_ observable. It's unavoidable, because the contract for `IObservable<T>.Subscribe` only accepts a single `IObserver<T>`. Once you split the sequence, which call to `Subscribe` should control the connectability of the source observable? There's no way to answer that in general.

Your proposed `Partition` operator suffers from the same issue. Take for example your original code:

    var (evens, odds) = observable.Partition(x => x % 2 == 0);
    evens.Subscribe(x => Console.WriteLine("Even: " + x));
    odds.Subscribe(x => Console.WriteLine("Odd: " + x));

When is the subscription to _observable_ made in your example? If it happens before **Partition** returns, then _evens_ and _odds_ may miss values. Otherwise, which of the two output sequences controls the subscription to the source? Both?

The temperature isn't clear here. If you use `Publish` instead, then you control the temperature.

spencercw:
The subscription to the source happens once both outputs have a subscription. In my head this makes sense because if you're only interested in one of the outputs you would just use `Where`.

Maybe it's just me but it seems a bit odd to have to resort to `Publish` in order to do what essentially amounts to an if-then-else.

RxDave:
You may still run into problems if you require both. For example, it will break this scenario: 

    evens.Subscribe(x => Console.WriteLine("Even: " + x));
    odds.Subscribe(x => Console.WriteLine("Odd: " + x));
    evensPlus1.Subscribe(x => Console.WriteLine("Even: " + x + 1));

When _odds_ subscribes, if a connection is made to the source observable, then _evensPlus1_ may miss notifications.

The solution of course is to `Publish` again! But if you just use `Publish` in the first place, then you have full control, without the overhead of multiple calls to `Publish`.

Of course, an operator like `Partition` might be useful is some scenarios, but I don't think it makes sense as a standard query operator. Or at least, if it's going to be a standard query operator, then it must expose the connectability of the source to the caller, either via an _out_ parameter or another item in the tuple, as in my example.

I don't see it as a simple if-then-else operation. It's primarily a temperature conversion operation with some additional semantics. In your case, it's filter semantics, but perhaps the same case could be made for projection or aggregation as well.

spencercw:
Fair enough, you got me there. It's probably best to just leave it in that case because it doesn't provide enough value on its own if you still have to deal with the connection semantics anyway.

RxDave:
To elaborate on my last point, I don't see it as being about splitting an observable into multiple sequences. It's primarily a temperature conversion operation, with additional filter semantics. In essence, what you're doing is taking a _cold_ observable and applying multiple filters to yield multiple _hot_ observables.

Your `Partition` example is just a special case of this generalization. A predicate, by definition, has two sides (true and false). So rather than passing in multiple positive predicates, all of which would yield their own sequences, you're passing in a single predicate to `Partition` and it's generating output observables from both sides of the predicate (true and false), mostly to save on the cost of evaluating two predicates on each item rather than just one.

Either way, it's a temperature conversion with filter semantics rolled in. If you don't expose the connectability in your API, then you're severely limiting the usefulness of your operator.

Also consider the problem of applying multiple projections over a _cold_ sequence to yield multiple _hot_ sequences. Something like this perhaps?

    IEnumerable<IObservable<R>> PublishMany<T,R>(this IO<T>, params Func<T,R>[] projections)

I see your `Partition` example as being a special-case of the "filter" version of `PublishMany`.

    IEnumerable<IObservable<T>> PublishMany<T>(this IO<T>, params Func<T,bool>[] filters)

Perhaps there could be some use for these operators?

spencercw:
Alright lets throw some stuff together and see if anything sticks. I'm struggling with the terminology a little. Assuming the input sequence is cold, does `Publish` (before calling `Connect`) return a hot or a cold sequence?

Presumably your `PublishMany` method needs to return something like either `IConnectableObservable<IObservable<R>>` or `IEnumerable<IConnectableObservable<R>>`? Assuming for the moment it's the latter because it's easier to work with, something like the below could be nice. If we have overloads for differing numbers of parameters then the projections can have different return types.

The projection won't necessarily return the same number of elements as the source, so I've changed it to return an `IObservable` (similar to `SelectMany`). I've also changed it to take in a sequence rather than each element which may or may not be a good idea.

```
public static (IConnectableObservable<R1>, IConnectableObservable<R2>)
    PublishMany<T, R1, R2>(
    this IObservable<T> o,
    Func<IObservable<T>, IObservable<R1>> p1,
    Func<IObservable<T>, IObservable<R2>> p2)
{
    // ...
}

public static (IConnectableObservable<string>, IConnectableObservable<double>)
    Run(IObservable<int> input)
{
    return input.
        Select(x => (Value: x, Even: x % 2 == 0)).
        PublishMany(
            o => o.Where(x => x.Even).Select(x => x.Value.ToString()),
            o => o.Where(x => !x.Even).Select(x => x.Value / 2.0));
}
```

Does this make any kind of sense?

spencercw:
Whoops, your edit wasn't showing when I replied. Maybe something like this, although it does mean there's no way of getting at the connectable to disconnect/dispose it.

```
public static (IObservable<R1>, IObservable<R2>)
    PublishMany<T, R1, R2>(
    this IObservable<T> o,
    out Action connect,
    Func<IObservable<T>, IObservable<R1>> p1,
    Func<IObservable<T>, IObservable<R2>> p2)
{
    // ...
}

public static (IObservable<string>, IObservable<double>)
    Run(IObservable<int> input, out Action connect)
{
    return input.
        Select(x => (Value: x, Even: x % 2 == 0)).
        PublishMany(
            out connect,
            o => o.Where(x => x.Even).Select(x => x.Value.ToString()),
            o => o.Where(x => !x.Even).Select(x => x.Value / 2.0));
}
```

RxDave:
> although it does mean there's no way of getting at the connectable to disconnect/dispose it.

Another good point. It's exactly why these kinds of operators typically don't make much sense. Vanilla `Publish` often works best (when breaking out of a query is necessary).

spencercw:
I suppose we could probably return something like `IConnectable<IObservable<R1>, IObservable<R2>>` but it's getting a little contrived and I'm not sure it adds enough value to make it worthwhile.

bartdesmet:
The best way to deal with this at the moment is to indeed use `Publish`, and return a tuple of all the sequences derived from it, together with a `Func<IDisposable>` to connect and disconnect. The caller will have to be made aware of the sequences being "arctic cold", namely that even calling `Subscribe` on them won't do anything until a call is made to connect matters. This is very similar to `EnableRaisingEvents` on various types such a `FileSystemWatcher`. Something like `PublishMany` could have some value, but leads to requiring N-ary overloads and API bloat that may not be worth it.

akarnokd:
I think this is too special to be part of the core Rx.NET feature set so I suggest this remains outside of it.

danielcweber:
I also think that the implications and semantics are too hard to get over and thus too special for this library.

Issue #400, opened 2017-05-17
Deadlock in LocalScheduler


tcberry:
Having updated to Rx.NET 3.0.0 I am seeing a deadlock occur in LocalScheduler due to the use of two locks, which are taken in different orders depending on the code path.

I am seeing one thead stuck in SystemClockChanged, having taken a lock on _gate and now waiting for a lock on s_gate. Another thread has entered ScheduleShortTermWork via EvaluateLongTermQueue, which has a lock on _gate, and is now waiting for a lock on s_gate.

Having reversed the order of locks taken in SystemClockChanged and rebuilt System.Reactive.Core locally, I no longer see this issue.

rosala:
We are using version 3.1.1 and we are experiencing the same deadlock as described by tcberry. 
See the attached snapshot:
![deadlock in localscheduler](https://user-images.githubusercontent.com/5101621/30542200-ab033a72-9c7e-11e7-84e6-e286874700fa.PNG)

lscpike:
I've just discovered this issue myself (after going through the v3 package rename pain) and was dismayed to find this 9 month old bug report and a 4.0 milestone. I'm not sure how more people aren't screaming about this. Maybe they can't figure out why their applications are freezing or randomly stopping working. 

Any chance this can be raised in priority for a 3.1.2 release? The fix is tiny as shown by @tcberry.

bartdesmet:
The PR for this issue has been merged; we'll look to include it in the next release.

PR #465, opened 2018-02-20
Reverse order of locks taken in LocalScheduler.SystemClockChanged. Po…

tcberryoxinst:
This change should fix issue #400.

My employer has been shipping a build of Rx 3.0 with this fix for almost a year, without any issues.


bartdesmet:
Thanks for catching and fixing this issue.

lscpike:
Thanks @bartdesmet. Is is possible to do a patch release for 3.1.1 (3.1.2) please?

Issue #420, opened 2017-08-28
Add Prepend and Append to Rx


bartdesmet:
To be on par with LINQ to Objects, we should add these.

mattpodwysocki:
@bartdesmet let's go even further to rename `StartWith` to `Prepend`, deprecating `StartWith`

onovotny:
Moving to vNext so we can get 4.0 out as per our discussion.

quinmars:
Is this up for grabs? This looks like a nice task to get hands on.

akarnokd:
@quinmars I'd think you can start working on this. Perhaps put it into the `System.Reactive.Observable.Aliases.QueryLanguage` so that you can have fun with its `Qbservable` generated counterpart :)

quinmars:
Remains the question if we should choose the n-ary approach or not. Personally I'd would prefer to exactly mirror the `IEnumerable`/`IQueryable` counter parts (not variadic) and keep `StartsWith()` as it is. Once/if the corefx team decides to add an [`AppendMany()` or `PrependMany()`](https://github.com/dotnet/corefx/issues/14119) we can consider to replace `StartsWith()` with `PrependMany()`.

quinmars:
Since #567 is merged, I think that issue can be closed.

PR #474, opened 2018-04-21
Update HomoIcon tool to reflect new repository structure and manual code changes

quinmars:
Originally, I started to work on issue #420 but I than realized that the HomoIcon tool does not work in its current form. The paths are not setup right and the formerly generated files seem to be hand optimzed (many #ifdefs are removed, the type notation in the docs was changed etc.).

This PR tries to fix those problems. The paths are valid again. Several #ifdefs are removed. From a code point of view it is almost the same (some ``using`s differ and some trailing white space was removed). While I tried to reflect the changes in the documentation as well, there are some type notations that differ (and it is only the notation that differs not the type). I have no idea if there is an easy way to fix those:

```diff
-     /// <seealso cref="Observable.ToEventPattern{TEventArgs}(IObservable{EventPattern{TEventArgs}})" />
+     /// <seealso cref="M:System.Reactive.Linq.Observable.ToEventPattern``1(System.IObservable{System.Reactive.EventPattern{``0}})" />
```
@bartdesmet, maybe you can tell which tool you used.
But I don't think it is such important, since the generated xml file won't differ.

Note: I haven't commit the newly generated files yet.


quinmars:
I committed now the newly generated files. That makes it probably easier to compare the result.

akarnokd:
Could you rebase this onto a fresh master?

quinmars:
Unfortunately, I haven't renamed the `develop` branch in the beginning. (It's my first PR). I will post a new PR, which will be rebased onto the new master.

Issue #478, opened 2018-04-30
Increase test coverage and the amount of platforms used during continuous integration validation


ghuntley:
We need to make it more accessible and safer for folks to contribute to this project. I would love to do a performance pass but not before basics such as increasing test coverage happen.

So this is a help wanted? Can you help out? Will you. 

Deliverables:
- Determine current coverage
- Increase test coverage where valuable and needed.
- Execute validation tests on Android & iOS platforms using  https://github.com/xamarin/Essentials/tree/master/DeviceTests

danielcweber:
I can check code coverage and evaluate what's missing.

ghuntley:
🏆 Make it so.

danielcweber:
The code coverage is just awesome. We have a coverage of 98% for the .netcore 2.0 run of the the tests. I did some updates though and will try to get rid of the nasty file-linking approach that is currently in place to enable the same unit tests on multiple platforms. I should be able to add a test runner for Xamarin.Android, every other platform is out of my scope unfortunately.

danielcweber:
A test runner has been added (#484). Removing file linking would not work unfortunately because xUnit would just not discover tests in referenced assemblies, so the new Android app will use file linking as well. This is a potential source for omissions. Maybe the files could be linked by wildcards?

danielcweber:
@onovotny Can we quickly and easily re-enable coveralls reporting by changing [this](https://github.com/dotnet/reactive/blob/f94b0062606c1f69be5c56594e016110c68f6209/Rx.NET/Source/build-new.ps1#L101) ?

onovotny:
I think it just needs `COVERALLS_REPO_TOKEN` set...

onovotny:
Okay, coverage is showing up now for CI builds:

https://coveralls.io/github/dotnet/reactive

akarnokd:
Looks like the source is not linked properly yet: https://coveralls.io/builds/17127097/source?filename=src%2FSystem.Reactive%2FLinq%2FObservable%2FAsObservable.cs

onovotny:
should be fixed now

akarnokd:
Thanks, it works now!

Btw, I've updated a PR and the [build](https://dotnetfoundation.visualstudio.com/Rx.NET/_build?buildId=909) says: `Blank or invalid Coveralls Repo Token.`

I'm assuming this is to avoid leaking the token but then how can a PR be assessed of its coverage?

onovotny:
[...]

PR #484, opened 2018-05-03
Add Xamarin.Android test runner

danielcweber:
This covers a part of #478.


akarnokd:
This could use some "update" button.

danielcweber:
I guess with everything becoming .netstandard, there is not really a point in this anymore (like there was when we still had Xamarin-PCLs and stuff like that). Closing.

PR #486, opened 2018-05-03
4.x: Add the RetryWhen operator

akarnokd:
This PR adds the `Observable.RetryWhen` extension method and operator. One can use the handler to have the `Exception` of the main source turned into a regular value and reacted to via an `IObservable` flow. This allows performing timed and/or retries by manipulating those errors via standard Rx operators.


akarnokd:
The build failed for some reason (or I can't see the reason on the VSTS dashboard).

onovotny:
I don't see anything either, which is weird. touching the PR should trigger another build.

akarnokd:
We are getting somewhere:

```
[xUnit.net 00:00:31.1080629] ReactiveTests.Tests.QbservableTest.Qbservable_Observable_Parity [FAIL]
       Missing Qbservable operator: RetryWhen
       Expected: True
       Actual:   False
       Stack Trace:
            at ReactiveTests.Tests.QbservableTest.Qbservable_Observable_Parity()
```

I can see a `Qbservable.Generated.cs` file. How is it regenerated? Do I have to manually add a partial class like `Qbservable.Joins.cs`?

onovotny:
I'll need to take a look and remember :)

akarnokd:
There is this [HomoIcon](https://github.com/dotnet/reactive/blob/master/Rx.NET/tools/HomoIcon/HomoIcon.sln) program that should generate those files but it hasn't been updated to support the new project structuring.

PR #492, opened 2018-05-05
4.x: Fix the capitalization of the directory Rx.Net -> Rx.NET

akarnokd:
Fixes the capitalization of the directory Rx.Net to Rx.NET so that `System.Reactive.Interfaces.csproj` is in the correct directory under case-sensitive OSes and in the Git repository.

image

Under Windows, I've experienced that VS (or the GitHub plugin) sometimes wants to recreate this file or just shows up as changed on its own.


akarnokd:
Closing as the directory is no longer there / fixed by some other change:

image

PR #493, opened 2018-05-05
Review producers and sinks

danielcweber:
The main objective of this PR is to strengthen the concept of the Sink class. Before, it would wrap, yet
still internally expose an observer. Inheriting classes would access it directly and the pattern of disposing the Sink after calling OnCompleted/OnError on the wrapped observer would be on them and repeat throughout the code. This commit encapsulates the observer and exposes Forward*-methods, taking care of disposing at the right places. The result is a lot of saved duplicated code. Moreover, we find that almost all classes inheriting from Sink also implement an IObserver of some kind by themselves, so we establish a base class Sink<TSource, TTarget>. The concept of a Sink is now more obvious, i.e. it serves as the logic between the source-stream and the emitted target stream. At last, we introduce IdentitySink which will just relay events, again, this will save a lot of duplicated code.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
@@ -83,7 +83,7 @@ namespace System.Reactive
         protected abstract IDisposable Run(IObserver<TSource> observer);
     }
 
-    internal abstract class Producer<TSource, TSink> : IProducer<TSource>
+    internal abstract class Producer<TTarget, TSink> : IProducer<TTarget>
         where TSink : IDisposable
     {
         /// <summary>
@@ -91,7 +91,7 @@ namespace System.Reactive
         /// </summary>
         /// <param name="observer">Observer to send notifications on. The implementation of a producer must ensure the correct message grammar on the observer.</param>
         /// <returns>IDisposable to cancel the subscription. This causes the underlying sink to be notified of unsubscription, causing it to prevent further messages from being sent to the observer.</returns>
-        public IDisposable Subscribe(IObserver<TSource> observer)
+        public IDisposable Subscribe(IObserver<TTarget> observer)
         {
             if (observer == null)
                 throw new ArgumentNullException(nameof(observer));
@@ -99,7 +99,7 @@ namespace System.Reactive
             return SubscribeRaw(observer, enableSafeguard: true);
         }
 
-        public IDisposable SubscribeRaw(IObserver<TSource> observer, bool enableSafeguard)
+        public IDisposable SubscribeRaw(IObserver<TTarget> observer, bool enableSafeguard)
         {
             var subscription = new SubscriptionDisposable();
 
@@ -109,7 +109,7 @@ namespace System.Reactive
             //
             if (enableSafeguard)
             {
-                observer = SafeObserver<TSource>.Create(observer, subscription);
+                observer = SafeObserver<TTarget>.Create(observer, subscription);
             }
 
             var sink = CreateSink(observer, subscription.Inner);
@@ -148,7 +148,7 @@ namespace System.Reactive
         /// <param name="sink">The sink object.</param>
         protected abstract IDisposable Run(TSink sink);
 
-        protected abstract TSink CreateSink(IObserver<TSource> observer, IDisposable cancel);
+        protected abstract TSink CreateSink(IObserver<TTarget> observer, IDisposable cancel);
     }
 
     internal sealed class SubscriptionDisposable : ICancelable
--- a/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
@@ -6,55 +6,95 @@ using System.Threading;
 
 namespace System.Reactive
 {
-    /// <summary>
-    /// Base class for implementation of query operators, providing a lightweight sink that can be disposed to mute the outgoing observer.
-    /// </summary>
-    /// <typeparam name="TSource">Type of the resulting sequence's elements.</typeparam>
-    /// <remarks>Implementations of sinks are responsible to enforce the message grammar on the associated observer. Upon sending a terminal message, a pairing Dispose call should be made to trigger cancellation of related resources and to mute the outgoing observer.</remarks>
-    internal abstract class Sink<TSource> : IDisposable
+    internal abstract class Sink<TTarget> : IDisposable
     {
-        protected internal volatile IObserver<TSource> _observer;
         private IDisposable _cancel;
+        private volatile IObserver<TTarget> _observer;
 
-        public Sink(IObserver<TSource> observer, IDisposable cancel)
+        protected Sink(IObserver<TTarget> observer, IDisposable cancel)
         {
             _observer = observer;
             _cancel = cancel;
         }
 
-        public virtual void Dispose()
+        public void Dispose()
         {
-            _observer = NopObserver<TSource>.Instance;
+            Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            _observer = NopObserver<TTarget>.Instance;
 
             Interlocked.Exchange(ref _cancel, null)?.Dispose();
         }
 
-        public IObserver<TSource> GetForwarder() => new _(this);
-
-        private sealed class _ : IObserver<TSource>
+        protected void ForwardOnNext(TTarget value)
         {
-            private readonly Sink<TSource> _forward;
+            _observer.OnNext(value);
+        }
 
-            public _(Sink<TSource> forward)
+        protected void ForwardOnCompleted()
+        {
+            _observer.OnCompleted();
+            Dispose();
+        }
+
+        protected void ForwardOnError(Exception error)
+        {
+            _observer.OnError(error);
+            Dispose();
+        }
+    }
+
+    /// <summary>
+    /// Base class for implementation of query operators, providing a lightweight sink that can be disposed to mute the outgoing observer.
+    /// </summary>
+    /// <typeparam name="TTarget">Type of the resulting sequence's elements.</typeparam>
+    /// <typeparam name="TSource"></typeparam>
+    /// <remarks>Implementations of sinks are responsible to enforce the message grammar on the associated observer. Upon sending a terminal message, a pairing Dispose call should be made to trigger cancellation of related resources and to mute the outgoing observer.</remarks>
+    internal abstract class Sink<TSource, TTarget> : Sink<TTarget>, IObserver<TSource>
+    {
+        protected Sink(IObserver<TTarget> observer, IDisposable cancel) : base(observer, cancel)
+        {
+        }
+
+        public abstract void OnNext(TSource value);
+
+        public virtual void OnError(Exception error)
+        {
+            ForwardOnError(error);
+        }
+
+        public virtual void OnCompleted()
+        {
+            ForwardOnCompleted();
+        }
+
+        public IObserver<TTarget> GetForwarder() => new _(this);
+
+        private sealed class _ : IObserver<TTarget>
+        {
+            private readonly Sink<TSource, TTarget> _forward;
+
+            public _(Sink<TSource, TTarget> forward)
             {
                 _forward = forward;
             }
 
-            public void OnNext(TSource value)
+            public void OnNext(TTarget value)
             {
-                _forward._observer.OnNext(value);
+                _forward.ForwardOnNext(value);
             }
 
             public void OnError(Exception error)
             {
-                _forward._observer.OnError(error);
-                _forward.Dispose();
+                _forward.ForwardOnError(error);
             }
 
             public void OnCompleted()
             {
-                _forward._observer.OnCompleted();
-                _forward.Dispose();
+                _forward.ForwardOnCompleted();
             }
         }
     }
--- a/Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
@@ -9,7 +9,7 @@ using System.Threading;
 
 namespace System.Reactive
 {
-    internal abstract class TailRecursiveSink<TSource> : Sink<TSource>, IObserver<TSource>
+    internal abstract class TailRecursiveSink<TSource> : IdentitySink<TSource>
     {
         public TailRecursiveSink(IObserver<TSource> observer, IDisposable cancel)
             : base(observer, cancel)
@@ -92,8 +92,7 @@ namespace System.Reactive
                         catch (Exception ex)
                         {
                             currentEnumerator.Dispose();
-                            _observer.OnError(ex);
-                            base.Dispose();
+                            ForwardOnError(ex);
                             Volatile.Write(ref _isDisposed, true);
                             continue;
                         }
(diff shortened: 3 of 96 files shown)


Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -6,55 +6,100 @@
...
+
     /// <summary>
     /// Base class for implementation of query operators, providing a lightweight sink that can be disposed to mute the outgoing observer.
     /// </summary>
-    /// <typeparam name="TSource">Type of the resulting sequence's elements.</typeparam>
+    /// <typeparam name="TTarget">Type of the resulting sequence's elements.</typeparam>
+    /// <typeparam name="TSource"></typeparam>
     /// <remarks>Implementations of sinks are responsible to enforce the message grammar on the associated observer. Upon sending a terminal message, a pairing Dispose call should be made to trigger cancellation of related resources and to mute the outgoing observer.</remarks>
-    internal abstract class Sink<TSource> : IDisposable
+    internal abstract class Sink<TTarget, TSource> : IObserver<TSource>, IDisposable
```
akarnokd: I'd be more comfortable with the arguments ordered `TSource, TTarget`.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -6,55 +6,100 @@
...
+        {
+            ForwardOnError(error);
+        }
+
+        public virtual void OnCompleted()
+        {
+            ForwardOnCompleted();
+        }
+
+        protected virtual void Dispose(bool disposing)
```
akarnokd: I'm not that familiar with C# but saw this pattern mentioned several times. Is this `Dispose(bool)` called by the system on finalization?

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -6,55 +6,100 @@
...
+        protected void ForwardOnError(Exception error)
+        {
+            _observer.OnError(error);
+            Dispose();
+        }
+
+        public IObserver<TTarget> GetForwarder() => new _(this);
 
-        private sealed class _ : IObserver<TSource>
+        private sealed class _ : IObserver<TTarget>
```
akarnokd: I haven't looked around much. Why is this wrapper needed? I'd assume so that `Sink.Dispose()` is not accessible to whomever receives the `_`.

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
```diff
@@ -188,28 +187,21 @@ private bool TryGetEnumerator(IEnumerable<IObservable<TSource>> sources, out IEn
                 // enumerating to find the next observable sequence. Therefore,
                 // we feed those errors directly to the observer.
                 //
-                _observer.OnError(exception);
-                base.Dispose();
+                ForwardOnError(exception);
```
akarnokd: I'd think the allocation [L157](https://github.com/dotnet/reactive/pull/493/files#diff-82a4030a0cc96eb062a6dead43b8f90fR157) is not really necessary. It could be replaced by simply:

```cs
_subscription.Disposable = next.SubscribeSafe(this);
```

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace System.Reactive
 {
-    internal abstract class TailRecursiveSink<TSource> : Sink<TSource>, IObserver<TSource>
+    internal abstract class TailRecursiveSink<TSource> : Sink<TSource>
```
akarnokd: This class is odd to my eyes. It is supposed to implement a trampolining mechanism where only one inner source should be running at a time but don't want to get into an infinite recursion when trying to continue with the next one.

Given that it operates out of a synchronous/blocking `IEnumerable`, I can't see why the stacks are even necessary. Also it enables some transformations on top of those which I don't know why they are provided.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: Yes, I was aimimg for that, it's only that way I could easily do a find/replace of the pattern Sink<...>, IObserver<...>. ReSharper may do that with one click maybe. I'll have a look.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: It's not since Sink doesn't have a finalizer. It's just a pattern that IMHO should be followed if we want to have a clean, state-of-the-art codebase.

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
danielcweber: Not in the scope of this PR, at least not for now. I am sure there is a lot to gain in allocation efficiency though.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: It's needed by the GetForwarder method and that's needed in one place only. I'd rather get rid of it but it's ok for now.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: Has been done.

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace System.Reactive.Disposables
+{
+    public abstract class AssignmentDisposable
+    {
+        /// <summary>
+        /// Gets or sets the underlying disposable. After disposal, the result of getting this property is undefined.
+        /// </summary>
+        protected IDisposable Get(ref IDisposable fieldRef)
+        {
+            var current = fieldRef;
```
akarnokd: Does `ref` preserve the volatileness of `fieldRef`? I'd call `Volatile.Read(ref fieldRef)` to be sure.

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            value?.Dispose();
+        }
+
+        protected bool GetIsDisposed(ref IDisposable fieldRef)
+        {
+            // We use a sentinel value to indicate we've been disposed. This sentinel never leaks
+            // to the outside world (see the Disposable property getter), so no-one can ever assign
+            // this value to us manually.
+            return fieldRef == BooleanDisposable.True;
```
akarnokd: `Volatile.Read(ref fieldRef)`?

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        protected void Dispose(ref IDisposable fieldRef)
+        {
+            Interlocked.Exchange(ref fieldRef, BooleanDisposable.True)?.Dispose();
```
akarnokd: Since Rx is calling `Dispose` all over the place and even possibly multiple times, I'd bias this so that if `fieldRef` is already `BooleanDisposable.True`, don't exchange:

```cs
if (Volatile.Read(ref fieldRef) != BooleanDisposable.True) {
    Interlocked.Exchange(ref fieldRef, BooleanDisposable.True)?.Dispose();
}
```

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
danielcweber: It's just a copy of what was there in SingleAssignmentDisposable. The old code produced warnings in this regard as well.

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
danielcweber: You see both patterns throughout the code, sometimes with Volatile.Read, sometimes without. Won't the volatile field just do?

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
danielcweber: Could be done.

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
akarnokd: https://stackoverflow.com/a/425150/61158

Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs
akarnokd: https://stackoverflow.com/a/425150/61158

danielcweber: The AssignmentDisposable shall be removed then. You may have to fetch and force-update your branch in some minutes, I'll do a rebase.

akarnokd: No need for removal, just use `Volatile.Read` as I suggested.

danielcweber: May the field then be marked as not volatile to get rid of the warnings ?

akarnokd: Yes, in [Reactive4.NET](https://github.com/akarnokd/Reactive4.NET/blob/master/Reactive4.NET/utils/SequentialDisposable.cs#L12) I didn't specify fields as `volatile` and had a [helper](https://github.com/akarnokd/Reactive4.NET/blob/master/Reactive4.NET/utils/DisposableHelper.cs#L29) static class work with `ref`s.

danielcweber: Awesome, that base helper static class does exactly that. Could you please review https://github.com/dotnet/reactive/pull/493/commits/3fc620da902592df3d408a0f46dc9cb88643c9c2 then.

akarnokd: Yes, much better.

danielcweber: Thanks

akarnokd:
The most non-trivial thing when writing an operator externally is this: https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Internal/Producer.cs#L52

If you don't want to include Rx.NET, there is no way to know a `Subscribe` should trampoline. The `Producer`s are internal so you can't externally depend on this feature either.

danielcweber:
We'll get there, I'm sure.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -1,60 +1,96 @@
...
-        public virtual void Dispose()
+        public void Dispose()
         {
-            _observer = NopObserver<TSource>.Instance;
+            Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            _observer = NopObserver<TTarget>.Instance;
```
akarnokd: You don't really need full barrier volatile write here due to the full barrier of `Exchange`. A `Volatile.Write(ref _observer, NopObserver<TTarget>.Instance)` should suffice.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -1,60 +1,96 @@
...
-    /// <remarks>Implementations of sinks are responsible to enforce the message grammar on the associated observer. Upon sending a terminal message, a pairing Dispose call should be made to trigger cancellation of related resources and to mute the outgoing observer.</remarks>
-    internal abstract class Sink<TSource> : IDisposable
+    internal abstract class Sink<TTarget> : IDisposable
     {
-        protected internal volatile IObserver<TSource> _observer;
         private IDisposable _cancel;
+        private volatile IObserver<TTarget> _observer;
 
-        public Sink(IObserver<TSource> observer, IDisposable cancel)
+        protected Sink(IObserver<TTarget> observer, IDisposable cancel)
```
akarnokd: Swap the assignment so that `_observer = ...` releases the plain `_cancel` field before the constructor returns.

Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
```diff
@@ -83,51 +83,78 @@ private IDisposable Run(IScheduler _, State x)
...
+                if (!sinkIsInternal)
+                    safeObserverDisposable.Disposable2 = runDisposable;
+            }
+            
             if (CurrentThreadScheduler.IsScheduleRequired)
             {
-                var state = new State { sink = sink, inner = subscription.Inner };
+                var state = new State { sink = sink, inner = runDisposable };
 
                 CurrentThreadScheduler.Instance.Schedule(state, Run);
```
akarnokd: I think this `BasicProducer.Run` could be replaced with a static delegate so that only one state object is allocated and not any delegate (I assume assigning a method to a delegate type actually instantiates some objects in C# right?):

```cs
       private struct State
        {
            public BasicProducer<TSource> parent;
            public SingleAssignmentDisposable subscription;
            public IObserver<TSource> observer;
        }

        static readonly Func<IScheduler, State, IDisposable> RunRun = (_, x) =>
        {
            x.subscription.Disposable = x.parent.Run(x.observer);
            return Disposable.Empty;
        };

//...
           if (CurrentThreadScheduler.IsScheduleRequired)
            {
                var state = new State { parent = this, subscription = subscription, observer = observer };
                CurrentThreadScheduler.Instance.Schedule(state, RunRun);
            }

```

danielcweber: Yes. I'm actually puzzled that making Run static won't make the compiler cache the delegate...

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: what's the advantage ?

akarnokd: Correct publication of the `IDisposable` reference received with respect to the memory model.

danielcweber: What's the incorrect behaviour that could happen?

akarnokd: Potential early publication of the referenced `IDisposable` before its state has become visible to other threads. Plain reference writes can be tricky in concurrent settings.

Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
danielcweber: Inlining Run as anonymous lambda that's not dependent on closure will do the trick. Interesting, I had thought that delegate caching will work on static methods as well.

danielcweber: I can implement all your suggestions, you may as well file a pull request to get the credit for the commit.

akarnokd: I'm working on `TailRecursiveSink` at the moment and I have these already changed in my version.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
danielcweber: What about the other accesses to _observer? Should be Volatile.Read then, right?

akarnokd: `_observer` is declared `volatile` so you get it for free. I tend to use `Volatile.Read` to clearly indicate the barriers in the code.

PR #494, opened 2018-05-05
4.x: Example of a new internal structure for writing operators

akarnokd:
This PR demonstrates with a minimal set of components of the viability to have an internal protocol that makes the boomeranging of `IDisposable`s and the need for introducing scheduling of synchronous generators obsolete in general.

This allows the use of the public `IObservable`/`IObserver` and the standard API to remain the same while enabling more aggressive optimizations internally. The drawback is that:

  - there is at least one type check per composed operator to detect if its source is a traditional `IObservable` or not, and
  - there is at least one type check per `Subscribe` call to detect if the consumer is a traditional `IObserver`.


Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/InternalRange.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            this.start = start;
+            this.end = start + count;
+        }
+
+        public override void Subscribe(IInternalObserver<int> observer)
+        {
+            var d = new BooleanDisposable();
+
+            observer.OnSubscribe(d);
```
danielcweber: Does every implementor of Subscribe have to call OnSubscribe by himself? Can this be internalized?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        public void OnCompleted()
+        {
+            Volatile.Write(ref upstream, BooleanDisposable.True);
+            downstream.OnCompleted();
+        }
+
+        public void OnError(Exception error)
+        {
+            Volatile.Write(ref upstream, BooleanDisposable.True);
```
danielcweber: Why wouldn't we want to dispose the upstream on an error, like Sink does today?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
```diff
@@ -0,0 +1,85 @@
...
+            var d = Volatile.Read(ref upstream);
+            if (d != BooleanDisposable.True)
+            {
+                Interlocked.Exchange(ref upstream, BooleanDisposable.True)?.Dispose();
+            }
+        }
+
+        public void OnCompleted()
+        {
+            Volatile.Write(ref upstream, BooleanDisposable.True);
```
danielcweber: Why wouldn't we want to dispose the upstream on completion, like Sink does today?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
```diff
@@ -0,0 +1,85 @@
...
+    internal interface IInternalObserver<in T> : IObserver<T>, IDisposable
+    {
+        /// <summary>
+        /// Called by the InternalSourceObservable to push an IDisposable before
+        /// any other OnXXX methods are called in order to provide the means
+        /// to cancel the flow both synchronously and asynchronously.
+        /// </summary>
+        /// <param name="upstream">The IDisposable pushed by the source InternalSourceObservable that
+        /// allows cancelling a flow.</param>
+        void OnSubscribe(IDisposable upstream);
```
danielcweber: Today, the equivalent of this is passing the upstream-disposable to the Sink on creation. Any reason why we would want to do it that way, and deal with a mutable field?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/InternalRange.cs
akarnokd: No. Some operators would have to call it before they do other preparations. Some operators do this from their own consumer's `OnSubscribe` (see `InternalTake`).

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
akarnokd: Cleaning up on a terminal event becomes the responsibility of the upstream operator.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
akarnokd: Upstream is considered cancelled at this point. I wonder how many cycles go into waste that every Sink.OnComplete disposes itself and its upstream over and over.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
akarnokd: 1) Allows the possibility to go allocation-free on this path. Sink usually gets handed a `SingleAssignmentDisposable` on every stage.
2) If the IDisposable is pushed before any `onNext` event, the upstream can be cancelled from within `onNext` and not wait for the usual `SingleAssignmentDisposable` to receive an actual `IDisposable`, which can only happen if the two stages are asynchronous/trampolined relative to each other. Not going async/trampolined saves allocation, time and indirection.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
danielcweber: Don't know. Protocol might dictate that the subscription be disposed, even after the sequence has ended.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/InternalTake.cs
```diff
@@ -0,0 +1,83 @@
...
+            public void OnNext(T value)
+            {
+                var r = remaining;
+                if (r != 0)
+                {
+                    remaining = --r;
+                    downstream.OnNext(value);
+                    if (r == 0)
+                    {
+                        downstream.OnCompleted();
```
danielcweber: Shouldn't that be followed by disposing the upstream subscription?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
akarnokd: That's the responsibility of the upstream operator. Typically either:

```cs
    downstream.OnComplete();
    resources.Dispose();
```

or 

```cs
    resources.Dispose();
    downstream.OnComplete();
```

If there are no resources in an operator, this call simply does not happen, such as with InternalRange. No reason to dispose the `BooleanDisposable` once the operator is about to stop signalling events anyway.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
danielcweber: Still puzzled. What if we subscribe to a source that doesn't know about OnSubscribe and therefore can't call it?

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/InternalTake.cs
akarnokd: More like before this, but yes.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/IInternalObserver.cs
akarnokd: This is internal and you don't subscribe to regular `IObservable`s with this but only to `IInternalObservable`s.

Rx.NET/Source/src/System.Reactive/Linq/InternalObservable/InternalTake.cs
danielcweber: But in line 18, we are subscribing to an ordinary IObservable. What will enforce that we will receive an IDisposable at all?

danielcweber:
Is it done that way in RxJava?

akarnokd:
This is a proven structure with RxJava, minus that wrapper stuff.

danielcweber:
Alright. What's a bit dangerous IMO is that this might actually be a bit much to swallow if done at once. If I see correctly, this PR would have to following advantages:

- Save the allocation of one SingleAssignmentDisposable for every Subscribe in the chain
- Introducing clearly defined concepts IInternalObservable and IInternalObserver to actually make the implementation of efficient and fast operators more simple.
- Save some computation time by avoiding unnecessary dispose-calls ("Cleaning up on a terminal event becomes the responsibility of the upstream operator.")

Add to the above list if I missed out on something.

Here's my two cents: Allocation efficiency is highly desirable and it's also desirable to either explain the existing Producer/Sink structure appropriately or replace them good by clear new concepts. The presented approach is good, however, I am still puzzled about when the call to OnSubscribe should happen, especially since you stated that "some operators would have to call it before they do other preparations. Some operators do this from their own consumer's OnSubscribe". Would operator implementation really become simpler for a casual Rx-user?

Besides, I feel that the concepts you describe are already there. Producer today is your InternalObservable, Sink today is your InternalObserver. There are some shortcuts as well (e.g. to avoid allocating a safe wrapper when subscribing to a Producer), just as you have in your approach. The number of SingleAssignmentDisposable is indeed a drawback, but it might be worth taking a look into the existing code whether it can be mitigated.

About your approach to avoid unnecessary dispose-calls, this is definitely a change in paradigm and while the existing code is maybe a bit too careful and conservative, existing user code might rely on exactly that behaviour. We don't want to break stuff at this stage.

It would maybe be more advisable to carefully rearrange the existing Producers/Sinks. That way, we will be able to reason about the consequences of the changes, and we have continuous testing.

akarnokd:
Thanks for the feedback.

> I am still puzzled about when the call to OnSubscribe should happen, especially since you stated that "some operators would have to call it before they do other preparations.

This requires a bit of a shift in thinking about how the subscription is established across operators. The rule is, call it once before anything tries to call `OnNext`/`OnError`/`OnCompleted`. If you can read Java code, [here are a lot of them](https://github.com/ReactiveX/RxJava/tree/2.x/src/main/java/io/reactivex/internal/operators/observable) for example.

> Would operator implementation really become simpler for a casual Rx-user?

I can't say for others, but I've implemented operators for the early 0.x - 1.x era of RxJava that used a very similar protocol as `IObservable`/`IObserver` has been this time. I felt a relief when going the proposed route as it freed me of the burden of those `SingleAssignmentDisposable`s everywhere and made the control flow more transparent in my eyes. We didn't have `Sink`s and `Producer`s matching Rx.NET though.

> I feel that the concepts you describe are already there

In concept, yes, but with the burden of the often mentioned boomeranging: the upstream's `IDisposable` may arrive later than the operator starts emitting `OnNext` items, so in order to stop the upstream, the `IDisposable` has to be swung around in a `SingleAssignmentDisposable` to be thread-safely disposed eventually.

> It would maybe be more advisable to carefully rearrange the existing Producers/Sinks

I've tried that first and got caught up with `TailRecursiveSink`. I might take another stab at it when your PR #493 cleans things up a bit. Note though that due to the intertwining, there might be no "small changes" to get it done.

glopesdev:
I fully agree with @danielcweber that any changes should be carefully weighted against the current implementation. From my point of view, I see this as a chance to fully understand all the design decisions that went into making the current design.

When dealing with reentrant code there are a lot of subtle implications that can break user scenarios, even down to compiler optimizations and system-dependent behavior.

To be honest, this is what I would really like to see before changing anything:
1. more documentation of the current implementation, explaining exactly why each line is there and what guarantees it is trying to enforce
2. make informed decisions of which guarantees we want to preserve, and which ones we might want to relax
3. expose a documented operator implementation API (maybe it is simply making Producer/Sink public)
4. write documented tests for that implementation API

akarnokd:
After spending almost 2 months with Rx.NET's code, I came to the conclusion that there is a certain scheduling expectation from operators and thus this internal structure would break that expectation. Mainly the `CurrentThreadScheduler.IsScheduleRequired` can cause oddities. Besides, the changes to the `Producer`/`Sink` are almost having the same benefits as this.

PR #500, opened 2018-05-11
[WIP] Reduce allocations in schedulers

danielcweber:
Some allocations of closures and delegates can be avoided, which should be very effective in code that uses Schedulers a lot. Note that this PR increases the language version used to C# 7.1 and takes a reference to System.ValueTuple for every target except .netstandard 2.


Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
```diff
@@ -18,6 +19,7 @@
     <None Update="Platforms\Windows\Strings_WindowsThreading.resx" CustomToolNamespace="System.Reactive" Generator="ResXFileCodeGenerator" LastGenOutput="Strings_WindowsThreading.Designer.cs" />      
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
+    <PackageReference Include="System.ValueTuple" Version="4.4.0" />
```
akarnokd: I'm a bit confused. Has this become a standard or is this just a library? I can remember this was mentioned in some C# what's new/upcoming.

danielcweber: It's a library that is needed for most targets to enable use of C# value tuples.

danielcweber: It's a feature of C# 7.

Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Async.cs
```diff
@@ -328,37 +328,37 @@ public static IDisposable ScheduleAsync(this IScheduler scheduler, DateTimeOffse
...
                 throw new ArgumentNullException(nameof(action));
 
-            return ScheduleAsync_(scheduler, default(object), dueTime, (self, o, ct) => action(self, ct));
+            return ScheduleAsync_(scheduler, action, dueTime, (self, closureAction, ct) => closureAction(self, ct));
         }
 
         private static IDisposable ScheduleAsync_<TState>(IScheduler scheduler, TState state, Func<IScheduler, TState, CancellationToken, Task> action)
         {
-            return scheduler.Schedule(state, (self, s) => InvokeAsync(self, s, action));
+            return scheduler.Schedule((state, action), (self, t) => InvokeAsync(self, t.state, t.action));
```
akarnokd: Does it have to be tuples? Can't you just define a new struct to hold both parameters and avoid the hassle with changing the language version and depending on yet another external library?

Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
```diff
@@ -22,7 +22,14 @@ public static IDisposable Schedule(this IScheduler scheduler, Action action)
...
                 throw new ArgumentNullException(nameof(action));
 
-            return scheduler.Schedule(action, Invoke);
+            // Surprisingly, passing the method group of Invoke will create a fresh
+            // delegate each an every time, although it's static, while an anonymous
+            // lambda without the need of a closure will be cached.
+            // Once Roslyn supports caching delegates for method groups,
+            // the anonymous lambda can be replaced by the method group again. Until then,
+            // to avoid the repetition of code, the call to Invoke is left intact.
+            // Watch https://github.com/dotnet/roslyn/issues/5835
```
akarnokd: As a general note, I'd think Rx 4.0 should run on wide variety of JITs and runtimes, most of which won't be advanced enough to cache things for you. Wouldn't it be safer to do these by hand so all of the targets get the optimal code?

Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Async.cs
danielcweber: C# 7.1 aims to avoid exactly that, having to create structs everytime you need such tuples. Besides, the language support for tuples is really nice, the syntax is very low-noise. What exactly do you mean by 'hassle with changing the language version' ?

Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
danielcweber: This is a missing compiler feature, not a runtime feature. Doing this by hand? Really?

Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Async.cs
akarnokd: Don't certain C# language versions require minimum .NET Framework, .NET Standard or .NET Core versions to run? https://en.wikipedia.org/wiki/C_Sharp_(programming_language)#Versions

danielcweber: Even if so, where's the break from C# 7.0 to 7.1? I didn't even pull it up to 7.3 right away, which would probably be fine as well.

PR #504, opened 2018-05-15
4.x: Improve the logic of AsyncLock

akarnokd:
This PR improves the `AsyncLock` class' logic:

- Avoid the cost of allocating a `queue` upfront in case there is no actual concurrency involved.
- Clearing that queue is more efficient by setting the reference to `null`.
- Enqueueing an `Action` when the "lock" is not held should not go through the queue because it can be executed immediately.
- Changed the lock to `this` (as `queue` can now be `null`). I've seen other places with `guard = new object()` but that's an allocation and `this` is available. Not sure what the policy is given that `AsyncLock` is public. If it was `internal` or have such copy, I'm sure such shortcuts could be safely taken.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
@@ -11,6 +11,7 @@ namespace System.Reactive.Concurrency
     /// </summary>
     public sealed class AsyncLock : IDisposable
     {
+        private object guard = new object();
         private Queue<Action> queue;
         private bool isAcquired = false;
         private bool hasFaulted = false;
@@ -28,7 +29,7 @@ namespace System.Reactive.Concurrency
                 throw new ArgumentNullException(nameof(action));
 
             // allow one thread to update the state
-            lock (this)
+            lock (guard)
             {
                 // if a previous action crashed, ignore any future actions
                 if (hasFaulted)
@@ -67,7 +68,7 @@ namespace System.Reactive.Concurrency
                 catch
                 {
                     // the execution failed, terminate this AsyncLock
-                    lock (this)
+                    lock (guard)
                     {
                         // throw away the queue
                         queue = null;
@@ -78,7 +79,7 @@ namespace System.Reactive.Concurrency
                 }
 
                 // execution succeeded, let's see if more work has to be done
-                lock (this)
+                lock (guard)
                 {
                     var q = queue;
                     // either there is no queue yet or we run out of work
@@ -101,7 +102,7 @@ namespace System.Reactive.Concurrency
         /// </summary>
         public void Dispose()
         {
-            lock (this)
+            lock (guard)
             {
                 queue = null;
                 hasFaulted = true;
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
@@ -11,7 +11,7 @@ namespace System.Reactive.Concurrency
     /// </summary>
     public sealed class AsyncLock : IDisposable
     {
-        private readonly Queue<Action> queue = new Queue<Action>();
+        private Queue<Action> queue;
         private bool isAcquired = false;
         private bool hasFaulted = false;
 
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
```diff
@@ -27,50 +27,72 @@ public void Wait(Action action)
             if (action == null)
                 throw new ArgumentNullException(nameof(action));
 
-            var isOwner = false;
-            lock (queue)
+            // allow one thread to update the state
+            lock (this)
```
danielcweber: Never ever lock on "this"! Code using the AsyncLock in a private field might use it as a lock as well and we're gonna run into some serious trouble.

danielcweber:
Some small review, see it there. Otherwise, I think work on AsyncLock is greatly appreciated. I had a look into this before and was considering an overload that takes an Action<T> and a state of T, to save more allocations of closures throughout the code. What do you think?

Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
```diff
@@ -27,50 +27,72 @@ public void Wait(Action action)
...
+                        hasFaulted = true;
                     }
+                    throw;
+                }
 
-                    try
+                // execution succeeded, let's see if more work has to be done
+                lock (this)
+                {
+                    var q = queue;
```
danielcweber: We're in a lock, why the local variable?

Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
akarnokd: That's why I noted it in the PR text. Adding a `guard = new object()` resolves this with the obvious added cost of allocation. If there was an internal version of this lock and the discipline to not lock on it by Rx code, that would be better. I'll update the PR to use a guard.

akarnokd:
> I had a look into this before and was considering an overload that takes an Action and a state of T, to save more allocations of closures throughout the code. What do you think?

The best would be not to use `AsyncLock` at all. It is a general purpose trampolining mechanism which could be inlined in a completely lock-free fashion most of the time.

Rx.NET/Source/src/System.Reactive/Concurrency/AsyncLock.cs
akarnokd: A habit from lock-free programming: touch fields as less as possible around atomic accesses.

danielcweber: I don't think there's much to gain, it just makes the code less readable.

danielcweber:
How complex would the inline be? Should it really be repeated throughout the code?

akarnokd:
Something along the lines of #499, but the general structure is something like this:

```cs
void Trampoline<TState>(ref int wip, ConcurrentQueue<TState> queue, 
        TState item, Action<TState> genericBody) 
{
    if (Interlocked.CompareExchange(ref wip, 1, 0) == 0) {
        genericBody(item);
        if (Interlocked.Decrement(ref wip) == 0) {
            return;
        }
    }
    else 
    {
        queue.Enqueue(item);
        if (Interlocked.Increment(ref wip) != 1) {
           return;
        }
    }
    var missed = 1;
    for (;;) {
        if (queue.TryDequeue(out var v)) {
            genericBody(v);
            continue;
        }

        missed = Interlocked.Add(ref wip, -missed);
        if (missed == 0) {
            break;
        }
    }
}
```

where `TState` should contain all relevant information so that `genericBody` could be a static `Action` and not a per-invocation allocated delegate. Note though that the contents of `for(;;)` can be something quite complicated beyond reading the contents of a queue: see [this](https://github.com/akarnokd/reactive-extensions/blob/master/reactive-extensions/observable/MergeMany.cs#L163) and [this](https://github.com/akarnokd/reactive-extensions/blob/master/reactive-extensions/observable/MergeMany.cs#L282).

danielcweber:
Ok, but you wouldn't inline that throughout the whole codebase, would you?

akarnokd:
I did in RxJava a lot because there are no `struct`s or `ref`s to support a generic function like the example above. The second reason is that one can never know if the JIT will actually inline the generic function or not, making it less efficient due to `TState` and `Action<TState>` having different types at different locations resulting in a so-called megamorphic dispatch.

> would you?

Yes I would specialize such trampolining for each operator I upgrade.

danielcweber:
>Yes I would specialize such trampolining for each operator I upgrade.

Why does it need to be specialized for every operator ?

akarnokd:
If we want optimized operators, they have to be adapted to the special circumstances.

danielcweber:
Let's discuss then if you really see the need to inline trampolining for some operator. AsyncLock may not be the prettiest thing but I think with a overload that takes `Action<T>` and state `T` we can get rid of a lot of closures and delegate allocations.

akarnokd:
`AsyncLock` is a public API class, right? Changing its API surface would break users outside the library. You should be introducing a new class with the desired properties.

danielcweber:
You can always add an overload.

danielcweber:
>dequeueing a struct is more involved as the whole outer struct has to be defaulted to cleanup all possible references.

I'm afraid I can't follow.

danielcweber:
Are we going for [this](https://gist.github.com/akarnokd/bb498ac53d498c75880064d30fd4b4f5) ?

akarnokd:
Not in this PR.

PR #506, opened 2018-05-16
4.x: Improve AutoDetachObserver memory footprint

akarnokd:
This PR inlines the `SingleAssignmentDisposable` behavior into the `AutoDetachObserver` to save on an allocation and the indirection.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/AutoDetachObserver.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/AutoDetachObserver.cs
@@ -3,13 +3,15 @@
 // See the LICENSE file in the project root for more information. 
 
 using System.Reactive.Disposables;
+using System.Threading;
 
 namespace System.Reactive
 {
     internal sealed class AutoDetachObserver<T> : ObserverBase<T>
     {
         private readonly IObserver<T> _observer;
-        private readonly SingleAssignmentDisposable _disposable = new SingleAssignmentDisposable();
+
+        private IDisposable disposable;
 
         public AutoDetachObserver(IObserver<T> observer)
         {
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Internal/AutoDetachObserver.cs
```diff
@@ -18,7 +20,17 @@ public AutoDetachObserver(IObserver<T> observer)
 
         public IDisposable Disposable
         {
-            set { _disposable.Disposable = value; }
+            set
+            {
+                if (Interlocked.CompareExchange(ref disposable, value, null) != null)
```
danielcweber: I wonder whether we can refactor that reoccuring logic into extensions that take refs, taking the idea from [AssignmentDisposable](https://github.com/danielcweber/Rx.NET/blob/3ed55448d0d80da5833673cea19ed9ab68c0cfe3/Rx.NET/Source/src/System.Reactive/Disposables/AssignmentDisposable.cs) a bit further.

danielcweber: The Setter-extensions could be called "SetSingle", "SetMultiple" etc....

akarnokd: Yes, I do this in my [repo](https://github.com/akarnokd/reactive-extensions/blob/master/reactive-extensions/tools/DisposableHelper.cs) but this is dotnet/reactive. 

The problem with such PRs is that if you need an utility for multiple other PRs as well, either it has to be PRd separately and then merged before anything in new PRs can use it.

danielcweber: >Yes, I do this in my repo but this is dotnet/reactive.

Awesome!

>...then merged before anything in new PRs can use it.

Let's hope any of our work will be merged eventually. Until then, we could agree on some basic improvements (like the import of ValueTuples, those extensions etc.) on a separate branch and rebase all our PRs on that branch.

PR #517, opened 2018-05-18
Reduce allocations for anonymous observables

danielcweber:
Almost every use of AnonymousObservable captures some variables in a closure and creates a delegate. This PR works around that by passing state.

Note: This is based on https://github.com/dotnet/reactive/pull/515 and https://github.com/dotnet/reactive/pull/518. These should be merged before this PR.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
@@ -35,19 +35,24 @@ namespace System.Reactive.Concurrency
             if (scheduler == null)
                 throw new ArgumentNullException(nameof(scheduler));
 
-            return new AnonymousObservable<TSource>(observer =>
-            {
-                var m = new SingleAssignmentDisposable();
-                var d = new SerialDisposable();
-                d.Disposable = m;
-
-                m.Disposable = scheduler.Schedule(() =>
+            return AnonymousObservable<TSource>.CreateStateful(
+                (observer, closureTuple1) =>
                 {
-                    d.Disposable = new ScheduledDisposable(scheduler, source.SubscribeSafe(observer));
-                });
+                    var m = new SingleAssignmentDisposable();
+                    var d = new SerialDisposable();
 
-                return d;
-            });
+                    d.Disposable = m;
+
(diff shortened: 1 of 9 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
```diff
@@ -35,19 +35,24 @@ public static IObservable<TSource> SubscribeOn<TSource>(IObservable<TSource> sou
...
                 throw new ArgumentNullException(nameof(scheduler));
 
-            return new AnonymousObservable<TSource>(observer =>
-            {
-                var m = new SingleAssignmentDisposable();
-                var d = new SerialDisposable();
-                d.Disposable = m;
-
-                m.Disposable = scheduler.Schedule(() =>
+            return AnonymousObservable<TSource>.CreateStateful(
```
akarnokd: How about not using a delegate either but extending some base class?

danielcweber: All this can be considered, but it would make review and merging at this point even more complicated. Instead I tried to keep replacements straightforward and easy to review.

danielcweber: I guess in the end, there will be no justification to use AnonymousObservable from inside Rx any more.

akarnokd: It is possible that this or some other class will be re-optimized in the future, depending on how the project leads decide upon critical components, which would lead to wasted "small step changes" on our parts. Just say'n.

akarnokd: `AnonymousX` in Rx were a workaround because C# still doesn't support anonymous inner classes like Java does and the devs wasn't apparently very keen on just defining new dedicated classes.

danielcweber: >Just say'n.

Don't really know what that's supposed to mean again. Try to encourage people for a change.

danielcweber:
Overridden by #549 if revised and approved.

danielcweber:
Obsoleted by #549.

PR #518, opened 2018-05-18
Add scheduler extension for actions

danielcweber:
This will allow to pass state to a scheduled action to avoid closure allocation and allow for delegate caching.

Note this is based on changes from https://github.com/dotnet/reactive/pull/515 which should be merged before this PR.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
@@ -26,6 +26,30 @@ namespace System.Reactive.Concurrency
         }
 
         /// <summary>
+        /// Schedules an action to be executed.
+        /// </summary>
+        /// <param name="scheduler">Scheduler to execute the action on.</param>
+        /// <param name="action">Action to execute.</param>
+        /// <param name="state">A state object to be passed to <paramref name="action"/>.</param>
+        /// <returns>The disposable object used to cancel the scheduled action (best effort).</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="scheduler"/> or <paramref name="action"/> is <c>null</c>.</exception>
+        internal static IDisposable Schedule<TState>(this IScheduler scheduler, Action<TState> action, TState state)
+        {
+            if (scheduler == null)
+                throw new ArgumentNullException(nameof(scheduler));
+            if (action == null)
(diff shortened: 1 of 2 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
```diff
@@ -25,6 +25,30 @@ public static IDisposable Schedule(this IScheduler scheduler, Action action)
...
 
+        /// <summary>
+        /// Schedules an action to be executed.
+        /// </summary>
+        /// <param name="scheduler">Scheduler to execute the action on.</param>
+        /// <param name="action">Action to execute.</param>
+        /// <param name="state">A state object to be passed to <paramref name="action"/>.</param>
+        /// <returns>The disposable object used to cancel the scheduled action (best effort).</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="scheduler"/> or <paramref name="action"/> is <c>null</c>.</exception>
+        public static IDisposable Schedule<TState>(this IScheduler scheduler, Action<TState> action, TState state)
```
akarnokd: Unit test? 

Also the typical signature is `state, action`.

danielcweber: Can do.

danielcweber: Changing the signature will create ambiguity with another overload. I will, for the moment, leave the signature as it is but make it internal so it can change without hassle.

Yeah69:
In all of my projects (private and in the company) I also add a "MinimalSchedule" extension method which even dismisses the state parameter competely (as generic parameter for the action and parameter for the extension method) and uses Unit.Default instead. Its handy in all cases where the state isn't relevant and isn't used.

danielcweber:
Could you post it?

Yeah69:
@danielcweber yeah, sure. I made a gist for you: https://gist.github.com/Yeah69/41f6c3bcfcfde7cbe53ae178a58bfbbc

A little bonus is the async version ;)
I see that you declare the extension method internal. That way no one outside of this project can benefit of this extension methods.

danielcweber:
Yes, the signature is up to debate because of possible ambiguity, therefore it is declared internal.

PR #521, opened 2018-05-23
Merge interfaces back into Sys.Reactive

onovotny:
There's no added value for this presently.


akarnokd:
It was somewhat useful to my extension library to not depend on the whole rx codebase when it only builds on the extra interfaces like ISubject and IScheduler.

onovotny:
@akarnokd from a consumer perspective, it adds additional complexity, which we were trying to reduce originally. We split it because of internal msft issues, which aren't an issue anymore. Are there any blockers to referencing just System.Reactive?

akarnokd:
Fair enough. I'll check the effects of going full System.Reactive later today.

PR #527, opened 2018-05-24
4.x: Refactor Linq operator unit tests into separate files/classes

akarnokd:
This PR refactors the various operator unit tests into their separate files and allows:

- locating tests of an operator by name instead of a category,
- running tests of one operator instead of a group of operators.


akarnokd:
Now the fun part. Finding out which is that 1 test that didn't run twice:

image

image

onovotny:
@akarnokd that could be due to the removal of some of the TFM's. We removed netcoreapp1.1, so that could account for one test depending on what the base version is.

akarnokd:
I was verifying periodically if the test runner still has the original 3534 test listed and it still shows me that:

image

>that could be due to the removal of some of the TFM's. 

This shows 7010 as well:

image

danielcweber:
Can you build it locally through the script, it should open a coverage report at the end, if it didn't decrease, it'll be good to merge.

akarnokd:
Looks like there were changes on the master earlier and [I compared against another PR](https://github.com/dotnet/reactive/pull/527#issuecomment-391870107) probably pointing further back in time.

Issue #270, opened 2016-08-24
Failed to start monitoring system clock changes


ButchersBoy:
Occasionally see exception message "Failed to start monitoring system clock changes" and application termination on desktop PCs making heavy utilization of RX.

I really don't want to write this up again; full explanation is here:

https://dragablz.net/2015/06/09/dealing-with-rx-exception-failed-to-start-monitoring-system-clock-changes/

In summary `PeriodicTimerSystemClockMonitor` may occasionally "time out" and throw the error.

It seems with RX v3 the work around described (and successfully running in our production instances) will no longer be possible.  

So can the code in `PeriodicTimerSystemClockMonitor`  be re-assessed?

onovotny:
At the very least, it seems like we need to provide a mechanism that exposes the `GetService<T>` to consumers to enable overrides. The mechanism shouldn't require any overrides -- perhaps a `RegisterService<T>` call instead?

onovotny:
The ability to set `PlatformEnglightenmentProvider.Current` has been restored with an `Obsolete` attribute attached. It is a strong goal of vNext to remove this mechanism all together, so I'm keeping this issue open to track the underlying need.

ButchersBoy:
Thank you.

baltie:
Would love to see the underlying issue fixed, it is a big problem that using Observable.Timer under heavy load causes the whole service to crash because of the InvalidOperationException being thrown in PeriodicTimerSystemClockMonitor.NewTimer after 100 retries.

shiftkey:
@baltie do you think the changes mentioned in [this blog post](https://dragablz.net/2015/06/09/dealing-with-rx-exception-failed-to-start-monitoring-system-clock-changes/) are sufficient? 

The only reservation I have is this comment from @ButchersBoy in the above post:

> The solution requires a message pump (which apparently is why RX doesn’t use this method by default). 

We are currently talking about the scope for v4, and changing the default behaviour of the internals (especially around timing) might be possible.

onovotny:
Moving to vNext so we can get 4.0 out asap.

akarnokd:
The `PeriodicTimerSystemClockMonitor` is pretty convoluted and possibly inconsistent/racy:

- It restarts the only timer upon a new event listener arrives to `SystemClockChanged`.
- It cancels the only timer if any of the listener gets removed.
- Has field `_lastTime`  accessed by `NewTimer` and `TimerChanged` possibly concurrently, overwriting each other's values.
[...]

PR #528, opened 2018-05-25
4.x: Fix PeriodicTimerSystemClockMonitor concurrency & failure behavior

akarnokd:
This PR fixes the `PeriodicTimerSystemClockMonitor` to:

- avoid restarting the timer when a new listener joins,
- avoid stopping the timer when one listener leaves (and there could be others still listening),
- avoid the data races between the `NewTimer` and the `TimerChanged` which involves using unix time milliseconds for the absolute time that can be atomically manipulated via `Volatile`/`Interlocked`,
- do not fail if the `SYNC_MAXRETRIES` has been reached but start backing off and sleeping a bit before retrying.

Fixes: #270


danielcweber:
>avoid restarting the timer when a new listener joins

This probably is not really what happens since there is `AddRef` and `Release` in `SystemClock`, although it should definitely be handled by the `PeriodicTimerSystemClockMonitor`.

Rx.NET/Source/src/System.Reactive/Internal/SystemClock.Default.cs
```diff
@@ -72,38 +90,58 @@ public PeriodicTimerSystemClockMonitor(TimeSpan period)
...
-                _lastTime = SystemClock.UtcNow;
+                var now = SystemClock.UtcNow.ToUnixTimeMilliseconds();
+                Interlocked.Exchange(ref _lastTimeUnixMillis, now);
+
                 _timer.Disposable = ConcurrencyAbstractionLayer.Current.StartPeriodicTimer(TimeChanged, _period);
-            } while (Math.Abs((SystemClock.UtcNow - _lastTime).TotalMilliseconds) > SYNC_MAXDELTA && ++n < SYNC_MAXRETRIES);
 
-            if (n >= SYNC_MAXRETRIES)
-                throw new InvalidOperationException(Strings_Core.FAILED_CLOCK_MONITORING);
+                if (Math.Abs(SystemClock.UtcNow.ToUnixTimeMilliseconds() - now) <= SYNC_MAXDELTA)
```
danielcweber: Why does a new timer have to be created, just because the previous one took to long to create? Can't we deal with that otherwise?

akarnokd: This is my understanding of the original logic:

The logic saves the `Now` in the `_lastTime` field and when the scheduled task runs, it reads it out to determine how much time has passed since the start of the timer and the first tick of the timer. If, for some reason the start itself gets delayed and runs eventually, you may end up with a clock-drift notification right away. Instead, the previous timer is cancelled and a new round is attempted. I'd think when the drift does actually happen this very first time, the timer action will restart the timer anyway so my guess is that this just makes the syncing happen earlier.

akarnokd:
> This probably is not really what happens since there is AddRef and Release in SystemClock

Indeed, interesting. Kind of hard to determine the intention here since the class itself is public and the notification is designed with an event: both implying the feature could be used from the outside and perhaps on multiple occasions. I'd say the safety should be added because of it.

danielcweber:
>I'd say the safety should be added because of it.

Definitely, but can we get rid of `AddRef` and `Release` then?

akarnokd:
I don't think so. If removed, it would do this all the time:

```cs
s_serviceSystemClockChanged.Value.SystemClockChanged += OnSystemClockChanged;
```

potentially registering that method over and over resulting in duplicate notifications.

Rx.NET/Source/src/System.Reactive/Internal/SystemClock.Default.cs
```diff
@@ -63,7 +78,10 @@ public PeriodicTimerSystemClockMonitor(TimeSpan period)
         {
             add
             {
-                NewTimer();
+                if (Interlocked.Increment(ref _timerActive) == 1)
+                {
+                    NewTimer();
+                }
 
                 _systemClockChanged += value;
```
danielcweber: This isn't thread safe, if we deal with Interlocked above, this all should go into a lock then.

danielcweber:
I think an oddity here is that the code could perfectly well use Rx itself...FromEvent, RefCount, all that stuff.

Rx.NET/Source/src/System.Reactive/Internal/SystemClock.Default.cs
akarnokd: I though += on a delegate is generates as an atomic operation by the compiler.

akarnokd:
I think there would be a chance of circular dependency when schedulers come into play.

Rx.NET/Source/src/System.Reactive/Internal/SystemClock.Default.cs
danielcweber: Nope, this is just non atomic read, change and write operation.

akarnokd: I haven't changed that line so if that's true, you've been running with an unsafe add for a long time now. Or, the compiler is smart enough to produce += as an atomic exchange: 

https://stackoverflow.com/a/3522556/61158

danielcweber: The compiler would do magic if the event add/remove code wasn't explicitly implemented. I think the unsafe add is harmless since that class only has only one consumer anyway. That being said, I agree that all of that looks horribly broken.

danielcweber: It's really odd, what this code does (creating a resource on first subscribe etc) is solvable easily by Rx itself! It should be redone with Rx.

PR #536, opened 2018-05-26
4.x: Add the RepeatWhen operator

akarnokd:
This PR adds the `RepeatWhen` operator that repeats a source observable sequence when it completes normally and then a secondary observable signals an item. It is the sibling of the `RetryWhen` operator from #486. This operator allows special/timed repetition of a source observable.

The operator shares its serialization helper with `RetryWhen` which has been refactored to support both operators. A missing license header was also added.


akarnokd:
Reworked onto a fresh master.

akarnokd:
Can this be merged now?

PR #538, opened 2018-05-26
Fix compile warning (CS1711) caused by class inline documentation

quinmars:
Fix the following compile warning:

      2>Runtime\CompilerServices\TaskObservableMethodBuilder.cs(206,30): warning CS1711: XML comment has a typeparam tag for 'T', but there is no type parameter by that name [D:\a\1\s\Rx.NET\Source\src\System.Reactive\System.Reactive.csproj]


onovotny:
Thanks!

PR #539, opened 2018-05-26
4.x: Inline operation instantiation into Observable, remove indirection types

akarnokd:
This PR inlines the operator instantiations into `Observable`, removing the interface and indirection calls. I've also fixed a couple of typos in the XML documentation.


danielcweber:
Added the do-not-merge label since this is to be discussed wrt. code that relies on IQueryLanguage by reflection.

akarnokd:
Sure, otherwise let's not forget the typos in the xml doc if this gets discared.

akarnokd:
I don't think this could be reasonably rebased so let's just defer it until #540.

PR #542, opened 2018-05-28
Fix compiler directive in ImmutableListTest

danielcweber:
The code should be tested when System.Reactive is visible to the test assembly which is exactly when it is signed (and not vice versa). The SIGNED-directive is thus added to the Uwp-DeviceRunner project.


danielcweber:
@akarnokd

danielcweber:
Because the Android-Testrunner  cannot be signed (https://github.com/dotnet/reactive/pull/484)

akarnokd:
Still, why not run this particular test always?

danielcweber:
Because ImmutableList is internal and must be declared visible to the test assembly which, in turn, must be signed.

danielcweber:
See InternalsVisibleTo.cs

akarnokd:
I leave this up to you, I have practically no undestanding of the underlying infrastructure.

PR #545, opened 2018-05-28
4.x: Dedicated Amb implementation for arrays and enumerables

akarnokd:
This PR adds dedicated, lock-free implementation for `Amb(params IObservable<T>[] sources)` and `Amb(this IEnumerable<IObservable<T>> sources)`.

The original internal implementation was chaining the two-argument `Amb`s which was both wasteful and possibly prone to stack overflow due to the nesting.


Rx.NET/Source/src/System.Reactive/Linq/Observable/AmbMany.cs
```diff
@@ -0,0 +1,233 @@
...
+            this.sources = sources;
+        }
+
+        protected override IDisposable Run(IObserver<T> observer)
+        {
+            var sourcesEnumerable = this.sources;
+            var n = 0;
+            var sources = new IObservable<T>[8];
+
+            try
```
danielcweber: Just use ToList or ToArray here.

akarnokd: This basically avoids the final copy in `ToArray` to the actual size as the internal expanding array will have slack otherwise. We can simply use `n` to limit the evaluation of the array here.

danielcweber: No please, slack is just fine, this is not the place to save bytes, we're possibly re-allocating and copying here more often, the Linq-internal ToArray does the trick just fine with a good resize-strategy.

akarnokd: Here is that extra copy which is avoided in the PR:

https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Collections/Generic/EnumerableHelpers.Linq.cs#L105

akarnokd: Or this:

https://github.com/dotnet/corefx/blob/master/src/Common/src/System/Collections/Generic/LargeArrayBuilder.cs#L329

danielcweber: I don't see how that copies more often than the PR

Rx.NET/Source/src/System.Reactive/Linq/Observable/AmbMany.cs
```diff
@@ -0,0 +1,224 @@
...
+            var o = new InnerObserver[n];
+            for (int i = 0; i < n; i++)
+            {
+                o[i] = new InnerObserver(this, i);
+            }
+            observers = o;
+            Volatile.Write(ref winner, -1);
+        }
+
+        internal static IDisposable Create(IObserver<T> observer, IObservable<T>[] sources, int n)
```
danielcweber: n is pretty much redundant, right?

Rx.NET/Source/src/System.Reactive/Linq/Observable/AmbMany.cs
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            var parent = new AmbCoordinator<T>(observer, n);
+
+            parent.Subscribe(sources, n);
+
+            return parent;
+        }
+
+        internal void Subscribe(IObservable<T>[] sources, int n)
```
danielcweber: Use observers.Length for n

Rx.NET/Source/src/System.Reactive/Linq/Observable/AmbMany.cs
```diff
@@ -0,0 +1,224 @@
...
+                {
+                    break;
+                }
+                inner.OnSubscribe(sources[i].Subscribe(inner));
+            }
+        }
+
+        public void Dispose()
+        {
+            var o = observers;
```
danielcweber: Can you not use those intermediate values, observers is readonly so this might be optimized by JIT, and o.Length is constant since o is an array.

akarnokd: I don't know the exact rules but even if they are readonly fields, it is possible the compiler would re-read the fields around those `Volatile` accesses all the time. This way, there is practically no chance the compiler inlines the read of the field back into all those places thus causing the aforementioned re-reads.

danielcweber: I am pretty sure it won't since it's all constant anyway. Anyway, it's just a lot of noise.

akarnokd: It is part of the optimized logic which also prevents bad decisions on the JIT's part, which I came to not expect much of right now.

danielcweber: What makes you think the JIT makes bad decisions?

akarnokd: Looks like I can't convince you about this. I'll update the PR.

danielcweber: I haven't done a lot of these micro-optimizations, and there might be good reasons to do this, but I have learned not to try to outsmart the JIT, because it usually does a pretty good job, and there is always a tradeoff wrt. readability, which is one of my concerns also. Plus, we want to get a lot of people involved in this, and not everybody can write that kind of code.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AmbMany.cs
```diff
@@ -0,0 +1,224 @@
...
+            var sourcesEnumerable = this.sources;
+            var sources = default(IObservable<T>[]);
+
+            try
+            {
+                sources = sourcesEnumerable.ToArray();
+            }
+            catch (Exception ex)
+            {
+                observer.OnError(ex);
```
danielcweber: Won't this be handled by SafeObserver?

akarnokd: I don't think there is a try-catch guarding the `IProducer.Run` method. It would likely crash out of the Subscribe method and count as a badly implemented `IObservable`:

https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Internal/Producer.cs#L59

danielcweber: If that's an issue, we should discuss that separately for all operators and put it in a central place because no other operator guards the Run like this.

akarnokd: They do: https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Linq/Observable/Defer.cs#L43

This nothing new and is there in operators that call delegates in their `Subscribe`/`Run` methods. Nothing to worry about.

danielcweber: ok, because your're enumerating? Fine for now I guess, but if it's a pattern, it should be....well, a pattern.

PR #546, opened 2018-05-28
4.x: Reimplement Timeout(TimeSpan) with lock-free methods

akarnokd:
This PR reimplements the `Timeout.Relative` operator with a lock-free set of methods. The atomic index change by one ensures that the main source won; when the index atomically changes to `long.MaxValue` the timeout won. Emissions after these transitions are guaranteed to be thread-safe thus no lock is needed.

The PR also inlined the two `SingleAssignmentDisposable`s and the `StableCompositeDisposable` return. However, to avoid a cyclic `Dispose` due to the `cancel` input saved in `Sink`, I have exposed just the clearing of the downstream observable and otherwise the `Dispose(bool)` disposes the inlined `IDisposable`s.

If this style is acceptable, the other variants will be updated to lock-free as well in a separate PR.


danielcweber:
Can we first have these Disposable-Helper-methods with SetSingle, SetSerial, SetMultiple, GetIsDisposed, Dispose taking those ref-parameters? I am currently working on this, it will reduce a lot of repetitive code, especially those volatile fields everywhere and lots of Interlocked stuff. This will be a lot cleaner afterwards and more intention revealing wrt. the disposable fields.

akarnokd:
Yes, I was planning on doing that. We can have just the helper merged, then a separate PR can go over all classes and change to them.

danielcweber:
Ok, I am working on the helper currently.

danielcweber:
Helper was merged.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
```diff
@@ -46,100 +49,82 @@ public _(Relative parent, IObserver<TSource> observer, IDisposable cancel)
...
+
+                        Disposable.SetSingle(ref _otherDisposable, d);
                     }
+                }
 
-                    if (onNextWins)
+                public override void OnNext(TSource value)
+                {
+                    var idx = Volatile.Read(ref _index);
+                    if (idx != long.MaxValue && Interlocked.CompareExchange(ref _index, idx + 1, idx) == idx)
```
danielcweber: How many messages would it have to produce per second for how many years to actually break this?

akarnokd: If you could produce 1 message per nanosecond, this would overflow in 500+ years.

Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -24,11 +24,16 @@ public void Dispose()
 
         protected virtual void Dispose(bool disposing)
         {
-            _observer = NopObserver<TTarget>.Instance;
+            ClearObserver();
 
             Interlocked.Exchange(ref _cancel, null)?.Dispose();
         }
 
+        protected void ClearObserver()
```
danielcweber: Why divide this?

akarnokd: Because _cancel holds a `SingleAssignmentDisposable` which holds `this` in itself and would call `Dispose` again on the implementing instance. In fact, any newer `Run` implementation returning `this` is prone to this wasteful call.

danielcweber: That's an issue I have been trying to wrap my head around for some time now. Can we defer this change for later, going with the wasteful call for now, because it would add more complexity to the Sink and I am thinking around ways to lower complexity of Sink.

PR #548, opened 2018-05-29
Avoid repeating code patterns in the Disposables-namespace.

danielcweber:


Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
...
+            {
+                // If it is the disposed instance, dispose the value.
+                if (old == BooleanDisposable.True)
+                {
+                    value?.Dispose();
+                    return;
+                }
+
+                // Atomically swap in the new value and get back the old.
+                var b = Interlocked.CompareExchange(ref fieldRef, value, old);
```
akarnokd: The problem with this sentinel technique is that if there are a lots of concurrent `SetMultiple`, threads may keep retrying on ever so state values. Having `old` read fresh inside the loop can change the window to a favorable thread-thread interaction.

http://cs.oswego.edu/pipermail/concurrency-interest/2016-September/015339.html

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
...
+                    ? DefaultDisposable.Instance
+                    : null
+                : current;
+        }
+
+        /// <summary>
+        /// Gets or sets the underlying disposable. After disposal, the result of getting this property is undefined.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown if the resource has already been assigned to.</exception>
+        internal static void SetSingle(ref IDisposable fieldRef, IDisposable value)
```
akarnokd: I found it sometimes valuable to have this method return true on success and false on an already disposed state.

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
...
+            if (old == null)
+                return;
+
+            if (old != BooleanDisposable.True)
+                throw new InvalidOperationException(Strings_Core.DISPOSABLE_ALREADY_ASSIGNED);
+
+            value?.Dispose();
+        }
+
+        internal static void SetMultiple(ref IDisposable fieldRef, IDisposable value)
```
akarnokd: Return `bool` true for success, `false` for disposed state?

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
...
+                {
+                    return;
+                }
+
+                // Otherwise, make the old reference the current and retry.
+                old = b;
+            }
+        }
+
+        internal static void SetSerial(ref IDisposable fieldRef, IDisposable value)
```
akarnokd: Return `bool` true for success, `false` for disposed state?

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
...
+        {
+            // We use a sentinel value to indicate we've been disposed. This sentinel never leaks
+            // to the outside world (see the Disposable property getter), so no-one can ever assign
+            // this value to us manually.
+            return Volatile.Read(ref fieldRef) == BooleanDisposable.True;
+        }
+
+        internal static void Dispose(ref IDisposable fieldRef)
+        {
+            Interlocked.Exchange(ref fieldRef, BooleanDisposable.True)?.Dispose();
```
akarnokd: I'd advise bias this against multiple dispose calls by doing a volatile read on `fieldRef` and checking if not already disposed. `Dispose` calls fly around pretty frequently in Rx.NET and I'm not sure generally those are at most 1-2 per target.

Also returning `bool` `true` if the current thread managed to dispose the contents could be valuable.

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,98 @@ public static IDisposable Create(Action dispose)
 
             return new AnonymousDisposable(dispose);
         }
+
+        /// <summary>
+        /// Gets or sets the underlying disposable. After disposal, the result of getting this property is undefined.
+        /// </summary>
+        internal static IDisposable Get(ref IDisposable fieldRef, bool returnDefaultWhenDisposed)
```
akarnokd: I'd split this into two, one guarding against the leak, the others not. Saving a branch on cases where there is no worry about leaking the reference would be beneficial.

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
danielcweber: It's currently just the existing stuff copy and pasted. I would not like to change the actual implementation in this PR.

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
danielcweber: Can be done.

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
danielcweber: I don't know what is more performant. A volatile read isn't free either.

danielcweber:
@akarnokd Good to go like this?

Rx.NET/Source/src/System.Reactive/Disposables/Disposable.cs
```diff
@@ -27,5 +29,111 @@ public static IDisposable Create(Action dispose)
 
             return new AnonymousDisposable(dispose);
         }
+
+        /// <summary>
+        /// Gets or sets the underlying disposable. After disposal, the result of getting this property is undefined.
+        /// </summary>
+        internal static IDisposable GetValue(ref IDisposable fieldRef)
+        {
+            var current = Volatile.Read(ref fieldRef);
```
akarnokd: Just return this.

danielcweber: ?

akarnokd: Returning null when the contents are disposes is misleading. Maybe there is no need for this method after all since I can just do a `Volatile.Read(ref field)` when I need the exact result.

danielcweber: I want the not-leaking-the-setinel part there

PR #551, opened 2018-05-29
4.x: Rework & fix SkipUntil with lock-free methods

akarnokd:
This PR reimplements the `SkipUntil` operator to use lock-free methods and also fixes a potential race condition with the original setup when both observables signal `OnError` concurrently.


Rx.NET/Source/src/System.Reactive/Linq/Observable/SkipUntil.cs
```diff
@@ -24,102 +25,141 @@ public SkipUntil(IObservable<TSource> source, IObservable<TOther> other)
...
+            }
 
-                public void OnNext(TSource value)
+            public override void OnNext(TSource value)
+            {
+                if (_forward)
                 {
-                    if (_forward)
-                        _parent.ForwardOnNext(value);
+                    if (Interlocked.CompareExchange(ref _halfSerializer, 1, 0) == 0)
```
danielcweber: Would Interlocked.Increment work as well?

akarnokd: Interlocked has to force a new value so it is more expensive in contended cases whereas CAS is generally cheaper, especially on the weaker platforms where increment may be actually implemented as a CAS loop. Here, there is no need to loop.

danielcweber:
Merging, I would really like to see that halfserializer-pattern generalized (static methods etc)

akarnokd:
I'll post a PR with additional toolsets.

Rx.NET/Source/src/System.Reactive/Linq/Observable/SkipUntil.cs
```diff
@@ -24,102 +25,141 @@ public SkipUntil(IObservable<TSource> source, IObservable<TOther> other)
...
-                var sourceSubscription = parent._source.SubscribeSafe(sourceObserver);
-                
-                sourceObserver.Disposable = sourceSubscription;
-                otherObserver.Disposable = otherSubscription;
-
-                return StableCompositeDisposable.Create(
-                    sourceSubscription,
-                    otherSubscription
-                );
+                Disposable.TrySetSingle(ref _otherDisposable, parent._other.Subscribe(new OtherObserver(this)));
```
danielcweber: Sorry, I missed those before merging, this will for now but it should probably be SetSingle (i.e. throw when already set).

PR #552, opened 2018-05-30
Remove a wrapper for AsyncLock in the corresponding test.

danielcweber:
It is unclear why that was useful once...suggestions, somebody ?


akarnokd:
No idea, maybe bypass some old platform bug?

akarnokd:
https://github.com/danielcweber/Rx.NET/blame/fef06a9a8685b1564f71fdb599042a54389e6c4c/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Concurrency/AsyncLockTest.cs#L97

Based on the dates, I'd guess this is an ancient remnant from the closed source era. So most likely Bart is the culprit :)

danielcweber:
Probably

PR #554, opened 2018-05-30
Enhance AsyncLock to support passing state.

danielcweber:


danielcweber:
Up to discussion: Should the public API be enhanced and the new internal method be made public?

akarnokd:
I'm not convinced this is much of a win. the `object` state requires `struct`s to be boxed which is equivalent with having a capture of a struct inside a freshly allocated delegate.

danielcweber:
Right, structs would need to be boxed, but I found most of the time, if something would need to be passed to AsyncLock, it's a class. Besides, even if boxing would need to occur, it's still one allocation less than closure+delegate.

danielcweber:
So what's your proposal for a trampoline with state? Duplicate AsyncLock with a generic type parameter?

akarnokd:
Yes, don't penalize the stateless user places.

danielcweber:
That would work if all places in a class would only ever pass the same type of state. This might be the case in Rx but it would severely limit the usefulness of AsyncLock. Hence the object parameter.

akarnokd:
Why would it be public? Support Rx primarily, also T can be object.

danielcweber:
Will revisit after #558 is merged.

PR #556, opened 2018-05-30
Use the helper methods for IDisposable-fields, avoid repeating code patterns

danielcweber:


Rx.NET/Source/src/System.Reactive/Disposables/ContextDisposable.cs
```diff
@@ -28,7 +28,7 @@ public ContextDisposable(SynchronizationContext context, IDisposable disposable)
                 throw new ArgumentNullException(nameof(disposable));
 
             Context = context;
-            _disposable = disposable;
+            Disposable.TrySetSingle(ref _disposable, disposable);
```
danielcweber: Is TrySetSingle required here?

Rx.NET/Source/src/System.Reactive/Disposables/ScheduledDisposable.cs
```diff
@@ -28,7 +28,7 @@ public ScheduledDisposable(IScheduler scheduler, IDisposable disposable)
                 throw new ArgumentNullException(nameof(disposable));
 
             Scheduler = scheduler;
-            _disposable = disposable;
+            Disposables.Disposable.TrySetSingle(ref _disposable, disposable);
```
danielcweber: Is TrySetSingle required here?

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
```diff
@@ -172,15 +170,8 @@ void DisposeAll()
...
-            var d = Volatile.Read(ref currentSubscription);
-            if (d != BooleanDisposable.True)
-            {
-                d?.Dispose();
-                if (Interlocked.CompareExchange(ref currentSubscription, null, d) == d)
-                {
-                    Drain();
-                }
-            }
+            if (Disposable.TrySetSerial(ref currentSubscription, null))
```
danielcweber: Does this still capture the original semantics?

danielcweber:
@akarnokd I added some notes for further review, thx.

Rx.NET/Source/src/System.Reactive/Disposables/ContextDisposable.cs
akarnokd: This is in the constructor so nothing else should be able to call `ContextDisposable.Dispose()` at this point.

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
akarnokd: There is a loop in `TrySetSerial` where a concurrent dispose failing the CAS will have it repeat the loop, detect that it needs to dispose `d` and then quits. This logic relies on the fact that if the single CAS fails, it can be only due to cancellation and there is nothing to do further.

Rx.NET/Source/src/System.Reactive/Disposables/ScheduledDisposable.cs
akarnokd: Same as above, this is in the constructor where it can't be disposed yet.

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
```diff
@@ -131,7 +129,8 @@ void Drain()
                             else
                             {
                                 var sad = new SingleAssignmentDisposable();
-                                if (Interlocked.CompareExchange(ref currentSubscription, sad, null) == null)
+
+                                if (Disposable.TrySetSingle(ref currentSubscription, sad) == TrySetSingleResult.Success)
```
danielcweber: @akarnokd This was added.

PR #558, opened 2018-05-31
4.x: Use Schedule calls with state

akarnokd:
This PR changes most of the no state argument `Schedule` calls into the state-aware `Schedule` calls. This should help avoid instantiating a new delegate because of the variable capture.


Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
```diff
@@ -56,9 +56,11 @@ protected override IDisposable SubscribeCore(IObserver<TSource> observer)
                 var d = new SerialDisposable();
                 d.Disposable = m;
 
-                m.Disposable = scheduler.Schedule(() =>
+                m.Disposable = scheduler.Schedule((scheduler, source, observer, d),
```
danielcweber: If you are using this overload, the scheduler will be passed. Convention throughout the rest of the code is to name it 'self' (now '_'). You can then omit it from the state-tuple.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
```diff
@@ -414,7 +414,7 @@ private void DrainQueue(ICancelable cancel)
                         if (shouldWait)
                         {
                             var timer = new ManualResetEventSlim();
-                            _scheduler.Schedule(waitTime, () => { timer.Set(); });
+                            _scheduler.Schedule(timer, waitTime, (_, state) => { state.Set(); return Disposable.Empty; });
```
danielcweber: Consider using the extension I recently added. You may omit the scheduler parameter (_), won't have to return Disposable.Empty everywhere.

Also, please give the inner variable a revealing name. I used to go for 'closureXYZ', e.g. closureTimer in this case. Just 'state' is not revealing.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FromEvent.cs
```diff
@@ -284,12 +284,19 @@ public IDisposable Connect(IObserver<TArgs> observer)
                         {
                             if (--_count == 0)
                             {
-                                _parent._scheduler.Schedule(_removeHandler.Dispose);
+                                _parent._scheduler.Schedule(_removeHandler, (_, self) => RemoveHandlerDispose(self));
```
danielcweber: 'self', throughout calls to Schedule, has bee 'traditionally' used to denote the executing scheduler. If 'this' is passed as state, I like using @this as a name.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Return.cs
```diff
@@ -33,13 +34,14 @@ public _(TResult value, IObserver<TResult> observer, IDisposable cancel)
 
             public IDisposable Run(IScheduler scheduler)
             {
-                return scheduler.Schedule(Invoke);
+                return scheduler.Schedule(this, (_, self) => self.Invoke());
             }
 
-            private void Invoke()
+            private IDisposable Invoke()
```
danielcweber: As above, consider the new extension and don't repeat returning Disposable.Empty.

Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
akarnokd: This was an odd case, wrapping the outer `scheduler`. I'm not sure if the inner scheduler would be the correct one in this situation.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
akarnokd: The problem with that overload is that the parameters are after the lambda, which I find harder to read.

Naming it like `closureTimer` makes little sense as these are not captured values by the lambda like before, they are now arguments.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FromEvent.cs
akarnokd: I'll inline the method call back here.

Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
danielcweber: It is generally just the executing one.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
danielcweber: Changing the signature creates ambigouities unfortunately.

If not 'closureTimer', at least something more revealing than 'state'.

akarnokd: I'll rename these. You could introduce an internal extension under a different name but with the same ordering of arguments.

danielcweber: I can rename it in a subsequent PR and let ReSharper do the reordering. Would you then use the Action<TState>-extension ?

akarnokd: I find the ordering of the parameters distracting with those Action overloads. I'd prefer this ordering and a renamed internal schedule method.

danielcweber: Ok I will file a PR with a renamed method and then we can get the Disposable.Empty out of that?

akarnokd: Sure.

Issue #382, opened 2017-04-14
ObservableTimeTest.Delay_DateTimeOffset_DefaultScheduler stuck in CI


bartdesmet:
See https://ci.appveyor.com/project/dotnetfoundation/rx-net-la3iv/build/4.0.0-MiscSubjectImprovements.180.build.281

```
[00:59:28] [xUnit.net 00:54:11.7505664] Tests.System.Reactive: [Long Running Test] 'ReactiveTests.Tests.ObservableTimeTest.Delay_DateTimeOffset_DefaultScheduler', Elapsed: 00:53:51
```

akarnokd:
There could be a bug in the operator that under some circumstances stops handling the delayed signals and the test just hangs waiting for the operator to complete. When the test is looped with practically a zero shift, it tends to hang after several hundred rounds on my machine:

```cs
for (int i = 0; i < 10000; i++)
{
    Assert.True(Observable.Return(1)
        .Delay(DateTimeOffset.UtcNow/* + TimeSpan.FromMilliseconds(1)*/)
        .ToEnumerable()
        .SequenceEqual(new[] { 1 }));
}
```

akarnokd:
Fix posted in #560.

PR #560, opened 2018-05-31
4.x: Fix Delay hanging due to wrong disposable-chain management

akarnokd:
This PR fixes a bug in `Delay` that can lead to an unwanted cancellation of the drain loop. In addition, the `IStopwatch.Elapsed` should be consistently read while holding the same lock.

The bug is that the `_cancelable` can be reset by the `IDisposable` of the very first task starting the drain loop while that drain loop already set the `IDisposable` for the next round. The first assignment then cancels the second run's cancellation token which then throws an `ObjectDisposedException` and the loop quits for good. Any queued items and terminal signals will then never be emitted.

The solution is to conditionally set that first `IDisposable` if the target is still null, otherwise the subsequent runs have set a newer one and nothing has to be done. The disposability is still intact in both cases.

To be able to do this, the original subscription containers had to be replaced by fields so that the new atomic `Disposable.TryX` methods can be employed.

In the pre-patch version, running the `DelayTest.Delay_DateTimeOffset_DefaultScheduler` with a loop a million times and with no practical time shift (FromMilliseconds(0)) triggered the bug within the first few hundred rounds. With the patch, the same test succeeded all million rounds. I did not change the test because it would take a minute to complete and also code coverage tends to choke on such high-volume tests in my experience.

Fixes: #382

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
@@ -36,7 +36,7 @@ namespace System.Reactive.Linq.ObservableImpl
             internal abstract class S : _
             {
                 protected readonly object _gate = new object();
-                protected readonly SerialDisposable _cancelable = new SerialDisposable();
+                protected IDisposable _cancelable;
 
                 protected readonly IScheduler _scheduler;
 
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
```diff
@@ -74,22 +74,31 @@ public override IDisposable Run(TParent parent)
...
-                    var sourceSubscription = new SingleAssignmentDisposable();
-                    _sourceSubscription = sourceSubscription;
-                    sourceSubscription.Disposable = parent._source.SubscribeSafe(this);
+                    Disposable.SetSingle(ref _sourceSubscription, parent._source.SubscribeSafe(this));
 
-                    return StableCompositeDisposable.Create(_sourceSubscription, _cancelable);
+                    return this;
+                }
+
+                protected override void Dispose(bool disposing)
```
danielcweber: Move those two disposables and Dispose into the base class?

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
```diff
@@ -521,7 +541,10 @@ public L(Absolute parent, IObserver<TSource> observer, IDisposable cancel)
 
                 protected override void RunCore(Absolute parent)
                 {
-                    _cancelable.Disposable = parent._scheduler.Schedule(parent._dueTime, Start);
+                    // ScheduleDrain might have already set a newer disposable
+                    // using TrySetSerial would cancel it, stopping the emission
+                    // and hang the consumer
+                    Disposable.TrySetSingle(ref _cancelable, parent._scheduler.Schedule(parent._dueTime, Start));
```
danielcweber: _cancelable was as SerialDisposable before, even if this works, consider TrySetSerial throughout to reveal intention more consistently.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
akarnokd: The class is ugly, didn't want to accidentally break it.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
akarnokd: That was the bug: _cancelable was SerialDisposable and SerialDisposable always forcefully overwrites and disposes the previous IDisposable it contains.

danielcweber: Ah, alright.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
danielcweber: Should be good for now, might be an optimization PR for later.

PR #561, opened 2018-05-31
4.x: Inline disposability into ScheduledItem

akarnokd:
This PR turns the abstract `ScheduledItem` into an `IDisposable` which allows direct cancellation when returned from a `Schedule()` method. The `CurrentThreadScheduler` had to wrap it before but I don't see any reason to not have it support dispose directly.

Also since this is a public API change, the API approval check required the expansion of the txt file with the altered signature.


danielcweber:
Thought about this before, Even if this is not breaking, I don't know if we are going for public surface changes right now per roadmap.

@onovotny @ghuntley Could you please weigh in?

danielcweber:
I'm thinking of merging this. First, there already were API surface changes from 4.0, second, we can't progress without touching the API and we're not retiring anything. However, could you please implement IDisposable explicitly to avoid confusion about Cancel/Dispose, I think this is good then.

akarnokd:
> However, could you please implement IDisposable explicitly to avoid confusion about Cancel/Dispose, I think this is good then.

I don't follow. I've added `IDisposable` and `public void Dispose()` but can't remove `Cancel` as it is the part of the API surface.

PR #562, opened 2018-05-31
The approved API of Qbservable.RetryWhen() has the wrong return type.

quinmars:
It should return an `IQbservable` instead of an `IObservable`.


ryanwersal:
Is the issue that `PublicApiGenerator` is producing an incorrect interpretation of the public API surface? If so, I'll have to dig into that library and find out why...

quinmars:
@ryanwersal, no worries, the generator works correct. It's the source that is wrong. To fix that I already have a pending PR. ;)

ryanwersal:
@quinmars Ah, good. I unfortunately don't know the API much so I couldn't tell myself - thanks for the confirmation! I should have a PR submitted tonight that will fix the NRE when the tests fail and output some sort of reasonable diff in the build logs.

quinmars:
@ryanwersal very nice! I really appreciate the API approval tests.

danielcweber:
>To fix that I already have a pending PR. ;)

@quinmars That's #537? Is it still a work in progress? To make this build you would have to manually fix the code in Qbservable.Generated as well, otherwise the build will remain broken.

quinmars:
No, it will build. I just fixed a wrong test. Of course the test will fail, but that's fine, that's the job of unit tests. But I can integrate it in #537 if you prefer.

danielcweber:
Sorry, didn't mean to close this PR...

quinmars:
Nevermind, I already integrated into PR #537 and wil push it in some minutes.

danielcweber:
>No, it will build. I just fixed a wrong test. Of course the test will fail, but that's fine, that's the job of unit tests. 

Right, what I mean, the CI will fail, so it won't allow me to merge your PR...

PR #563, opened 2018-06-01
4.x: Add ExceptionHelper & HalfSerializer lock-free tools

akarnokd:
This PR adds two lock-free helper class to work with partial signal serialization and atomic exception management.

#### HalfSerializer

This is an efficient, lock-free serialization when the `OnNext` are still coming in sequentially but there could be a concurrent `OnError` or `OnCompleted` happening, for example, with `TakeUntil` or `SkipUntil`. These methods establish the correct sequential behavior.

#### ExceptionHelper

Hosts a shared termination indication exception (similar to how `BooleanDisposable.True` is used for indicating a disposed `IDisposable` field).

In addition, it offers methods for atomically setting one Exception, swapping in a the terminal exception and getting back the last exception and the ability to aggregate exceptions until a terminal exception is swapped in. This latter feature can come in handy with future `SelectMany` and `Concat` operators that want to delay errors from the inner sequences and keep processing them, then finally emitting all the exceptions gathered as an `AggregateException`.


danielcweber:
Pretty awesome concept.

PR #565, opened 2018-06-01
Add a test that asserts the current behaviour of SkipUntil.

danielcweber:
Current behaviour can be considered buggy, but a change is breaking.

### DIFF
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/SkipUntilTest.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/SkipUntilTest.cs
@@ -418,6 +418,31 @@ namespace ReactiveTests.Tests
                 OnCompleted<int>(200)
             );
         }
+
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/SkipUntilTest.cs
```diff
@@ -418,6 +418,27 @@ public void SkipUntil_Immediate()
...
+            var r = scheduler.CreateHotObservable(
+                OnCompleted(250, 1)
+            );
+
+            var res = scheduler.Start(() =>
+                l.SkipUntil(r)
+            );
+
+            res.Messages.AssertEqual(
+                );
```
akarnokd: Perhaps indicate with a commented-out code what the expected outcome should be?

danielcweber: done.

PR #569, opened 2018-06-04
PriorityQueue fix

danielcweber:
Repro of a bug in PriorityQueue and @astrofishgames fix as in https://github.com/dotnet/reactive/pull/329.


danielcweber:
If that was clear at all...for the repro, I just assumed the queue is supposed to keep an ordered list and rearrange it on the fly...

danielcweber:
@akarnokd Two basic tests were added.

PR #571, opened 2018-06-04
Update API approvals to report a small diff on failure

ryanwersal:
Currently, the VSTS CI runs of the build that contain API approval failures are poorly communicated due to ApprovalTests appearing to favor UI oriented diffing tools. I spent a bit of time investigating why it wasn't falling back to the Xunit2 "reporter" but, even if it used that correctly, it only points out a character position of where it deviated which still wasn't very useful.

Instead, I looked at adding a new reporter that would produce a minimal diff should the tests fail. This is what I have implemented in this PR.

Example output from build script:
```
      Stack Trace:
           at ApprovalTests.Approvers.FileApprover.Fail()
           at ApprovalTests.Core.Approver.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
           at ApprovalTests.Approvals.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
           at ApprovalTests.Approvals.Verify(IApprovalWriter writer, IApprovalNamer namer, IApprovalFailureReporter reporter)
           at ApprovalTests.Approvals.Verify(IApprovalWriter writer)
           at ApprovalTests.Approvals.Verify(String text)
        tests\Tests.System.Reactive.ApiApprovals\Api\ApiApprovalTests.cs(26,0): at ReactiveTests.Tests.Api.ApiApprovalTests.Core()
      Output:
        -         public bool Equals(System.Reactive.Unit other) { }
        +         public bool Equals(System.Reactive.Unit foobar) { }
```

~~I'm marking this as a WIP as I can't seem to find a good way to tell we're running the tests from within the Powershell build script. This is needed so I can make it always use the `DiffPlexReporter` when in those circumstances but fall back to the GUI diff tools if running from within an IDE. It is also possible that we just leave it as-is and document how to swap out the reporter for local dev/debugging (I can certainly look to updating the wiki with some documentation). I'd appreciate any input on a direction to pursue here.~~

### DIFF
--- a/Rx.NET/Source/tests/Tests.System.Reactive.ApiApprovals/Api/ApiApprovalTests.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive.ApiApprovals/Api/ApiApprovalTests.cs
@@ -9,13 +9,23 @@ using System;
 using System.Linq;
 using System.Reflection;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace ReactiveTests.Tests.Api
 {
+#if DEBUG
     [UseReporter(typeof(DiffReporter))]
+#else
+    [UseReporter(typeof(DiffPlexReporter))]
+#endif
     [IgnoreLineEndings(true)]
     public class ApiApprovalTests
     {
+        public ApiApprovalTests(ITestOutputHelper output)
+        {
+            DiffPlexReporter.INSTANCE.Output = output;
+        }
+
         [Fact]
         public void Core()
         {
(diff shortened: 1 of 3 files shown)


danielcweber:
I could do without the graphical diff. An error in the approval tests should rarely happen if PRs are carefully crafted, and if so, text should be perfectly fine.

ryanwersal:
@danielcweber Sorry about that - it is ready for review and rebased against latest master.

danielcweber:
LGTM. Last minute idea: Based on compiler directive #DEBUG, switch between `DiffReporter` and `DiffPlexReporter`, for easy graphical output on local VS, would be the best of both worlds. What do you think?

ryanwersal:
@danielcweber  Done. PR is squashed and rebased. 

I admit I was a bit overly focused on trying to get the diff viewer (gui vs text) handled within the ApprovalTests context... Thanks for the feedback and review!

PR #572, opened 2018-06-05
Use HalfSerializer extensions

danielcweber:
The new HalfSerializer extensions should be used where applicable to reduce code repetition. The extensions have been changed to work on ISink instead of IObserver, a new base interface for Sinks.


danielcweber:
The alternative would have been to have the extensions keep working on IObserver, have only Sink<T> call those extensions and thus introduce a rather obscure zoo of Sinks, e.g. HalfSerializedSink<T>, HalfSerializedSink<TSource, TTarget>, HalfSerializedIdentitySink....

danielcweber:
It would be cool if RetryWhen could make use of the HalfSerializer now and somewho fit into the Producer/Sink scheme.

PR #574, opened 2018-06-05
4.x: Make RefCountDisposable lock-free 

akarnokd:
This PR changes the `RefCountDisposable` to use lock-free algorithms which are based on atomic conditional state changes: changes in the number of child disposables and whether the main disposable can be disposed.

It works by using bit 31 in the `_count` as the indicator for disposing the main underlying disposable once all child disposables have been disposed. Bits 30-0 are then the number of active child disposables. Since the field is still `int` this allows atomic CAS operations on this composite bit-structure.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Disposables/RefCountDisposable.cs
+++ b/Rx.NET/Source/src/System.Reactive/Disposables/RefCountDisposable.cs
@@ -72,6 +72,9 @@ namespace System.Reactive.Disposables
                     return Disposable.Empty;
                 }
 
+                // Should not overflow the bits 0..30
+                System.Diagnostics.Debug.Assert((cnt & 0x7FFFFFFF) < int.MaxValue);
+
                 // Increment the active count by one, works because the increment
                 // won't affect bit 31
                 var u = Interlocked.CompareExchange(ref _count, cnt + 1, cnt);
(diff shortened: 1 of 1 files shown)


danielcweber:
Nice.

Rx.NET/Source/src/System.Reactive/Disposables/RefCountDisposable.cs
```diff
@@ -56,20 +58,31 @@ public RefCountDisposable(IDisposable disposable, bool throwWhenDisposed)
...
                     return Disposable.Empty;
                 }
-                else
+
+                // Should not overflow the bits 0..30
+                System.Diagnostics.Debug.Assert((cnt & 0x7FFFFFFF) < int.MaxValue);
+
+                // Increment the active count by one, works because the increment
+                // won't affect bit 31
+                var u = Interlocked.CompareExchange(ref _count, cnt + 1, cnt);
```
danielcweber: The bahaviour is excatly as in the locked version before, after int.MaxValue it will wrap and lead to strange behaviour (actually, the new version will act as if disposed). We should take the opportunity and not just Debug.Assert but throw something.

danielcweber: Debug.Assert will not be compiled into Release builds. 2 billion child subscribers are possible, tested this morning.

akarnokd: We could expand to `long`, allowing far more child disposables.

danielcweber: I don't really think anybody will hit that number, let's for now just make sure it throws instead of wrapping.

PR #575, opened 2018-06-06
Rework subscription in Producer such that all Sinks set the upstream disposable by themselves.

danielcweber:
Sinks will no longer be passed an IDisposable on creation. Previously, the passed IDisposable would (through a SingleAssignmentDisposable) represent the upstream subscription (or whatever Producer.Run would return). Now, every Sink is responsible for setting the upstream by itself.


Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
```diff
@@ -24,17 +24,17 @@ public TailRecursiveSink(IObserver<TSource> observer, IDisposable cancel)
...
-                return Disposable.Empty;
+                return;
 
             stack = new Stack<IEnumerator<IObservable<TSource>>>();
             stack.Push(current);
 
             Drain();
 
-            return new RecursiveSinkDisposable(this);
+            SetUpstream(new RecursiveSinkDisposable(this));
```
akarnokd: Probably this allocation can now be eliminated by overriding `Dispose(bool)` and inlining that `DisposeAll` method there. Btw there are some operators that do not have an upstream so for them, this is not really needed anymore.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
```diff
@@ -266,29 +266,29 @@ public TimeHopping(IObservable<TSource> source, TimeSpan timeSpan, IScheduler sc
...
-                public IDisposable Run(TimeHopping parent)
+                public void Run(TimeHopping parent)
                 {
                     _list = new List<TSource>();
 
                     var d = parent._scheduler.SchedulePeriodic(parent._timeSpan, Tick);
                     var s = parent._source.SubscribeSafe(this);
 
-                    return StableCompositeDisposable.Create(d, s);
+                    SetUpstream(StableCompositeDisposable.Create(d, s));
```
akarnokd: With the new structure, we can eliminate most `StableCompositeDisposable` creation now.

Rx.NET/Source/src/System.Reactive/Linq/Observable/CombineLatest.Generated.tt
```diff
@@ -98,7 +98,7 @@ for (var j = 1; j <= i; j++)
 }
 #>
 
-                return StableCompositeDisposable.Create(subscriptions);
+                Set(StableCompositeDisposable.Create(subscriptions));
```
akarnokd: Did you mean `SetUpstream`?

Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
```diff
@@ -41,22 +41,19 @@ internal sealed class _ : IdentitySink<TSource>
...
                     _scheduler = parent._scheduler;
                 }
 
-                public IDisposable Run(IObservable<TSource> source)
+                public void Run(IObservable<TSource> source)
                 {
-
                     CreateTimer(0L);
 
                     Disposable.SetSingle(ref _mainDisposable, source.SubscribeSafe(this));
```
akarnokd: `_mainDisposable` could be `_upstream` now so this can be just `SetUpstream` and no need for a separate field.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
```diff
@@ -42,10 +42,12 @@ public IDisposable Run(Using<TSource, TResource> parent)
                 }
                 catch (Exception exception)
                 {
-                    return StableCompositeDisposable.Create(Observable.Throw<TSource>(exception).SubscribeSafe(this), disposable);
+                    SetUpstream(StableCompositeDisposable.Create(Observable.Throw<TSource>(exception).SubscribeSafe(this), disposable));
```
akarnokd: This is not the fault of this PR but why create a sequence just to signal OnError(exception) practically immediately?

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.Generated.tt
```diff
@@ -110,7 +110,7 @@ for (var j = 1; j <= i; j++)
 #>
                 });
 
-                return StableCompositeDisposable.Create(disposables);
+                Set(StableCompositeDisposable.Create(disposables));
```
akarnokd: Did you mean `SetUpstream`?

Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
```diff
@@ -56,5 +56,8 @@
     <None Update="Linq\Observable\Zip.Generated.tt" Generator="TextTemplatingFileGenerator" LastGenOutput="Zip.Generated.cs" />
     <Compile Update="Linq\Observable\Zip.Generated.cs" DesignTime="True" AutoGen="True" DependentUpon="Zip.Generated.tt" />   
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
akarnokd: ???

Rx.NET/Source/src/System.Reactive/Linq/Observable/CombineLatest.Generated.tt
danielcweber: Yes...

Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
danielcweber: Hmm...but _mainDisposable must be disposed without disposing the entire sink (L83).

Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
danielcweber: Maybe in earlier code there was a scheduler involved.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.Generated.tt
danielcweber: Yes.

Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
danielcweber: Will be reverted.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
danielcweber: Sounds good, in a follow-up PR then.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
akarnokd: Yeah, right, forgot you only added SetUpstream, and the field is still inaccessible from subclasses. Maybe worth considering so that a field can be saved if there's already one in the base Sink.

danielcweber: The upstream could have SerialDisposable-semantics, that way it should be possible to save the _otherDisposable field as well. I would not like to expose otherwise private fields.

danielcweber: SetUpstream would then call SetSerial.

danielcweber: Why wouldn't it work for Timeout? It just replaces the main-upstream by the other-upstream. No combining.

Rx.NET/Source/src/System.Reactive/Internal/TailRecursiveSink.cs
danielcweber: Not so trivial, the Sink would have to be protected against disposal before Run has run. Can do it in a follow up.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
akarnokd: The return of `SubscribeSafe` is racing with the rest of the operator; a similar bug was fixed in #560.

danielcweber: You're sure you don't refer to _timerDisposable? Maybe I'll work it into a subsequent PR we can then talk about, the idea seems nice and I think it can work. If it has races, you could show me what I'm missing.

PR #579, opened 2018-06-08
4.x: Fix timed delay hangs with dotCover and the DefaultScheduler

akarnokd:
For some reason, dotCover would hang in various timed test which use the `DefaultScheduler`. I don't know why, but removing that empty `try-catch` along with the atomic dispose management solves the issue. 

In addition, the `Virtual_ThreadSafety()` was running extremely slow in its original form (more than 5 minutes). The reduced time amount and a more close thread-synchronization should be faster now while preserving the behavior under test.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/ConcurrencyAbstractionLayerImpl.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/ConcurrencyAbstractionLayerImpl.cs
@@ -161,16 +161,7 @@ namespace System.Reactive.Concurrency
                 _state = state;
                 _action = action;
 
-                // Don't want the spin wait in Tick to get stuck if this thread gets aborted.
-                try { }
-                finally
-                {
-                    //
-                    // Rooting of the timer happens through the Timer's state
-                    // which is the current instance and has a field to store the Timer instance.
-                    //
-                    Disposable.SetSingle(ref _timer, new System.Threading.Timer(_ => Tick(_), this, dueTime, TimeSpan.FromMilliseconds(System.Threading.Timeout.Infinite)));
-                }
+                Disposable.SetSingle(ref _timer, new System.Threading.Timer(_ => Tick(_), this, dueTime, TimeSpan.FromMilliseconds(System.Threading.Timeout.Infinite)));
             }
 
             private static void Tick(object state)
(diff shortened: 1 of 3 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/ConcurrencyAbstractionLayerImpl.cs
```diff
@@ -183,23 +174,16 @@ private static void Tick(object state)
...
-
             public void Dispose()
             {
-                var timer = _timer;
-                if (timer != TimerStubs.Never)
+                if (Disposable.TryDispose(ref _timer))
                 {
                     _action = Stubs<object>.Ignore;
-                    _timer = TimerStubs.Never;
                     _state = null;
```
danielcweber: I see a race here where _state could be set null but the timer will execute, passing the wrong state (it doesn't help that _action is set to Stubs.Ignore before, that might be rearranged). Maybe timer._action should only be called when TryDispose was true, setting _action and _state by itself (or by some DisposeCore method)?

danielcweber:
Do you know anything about the issue or bug here? Seems that the SpinWait would loop forever?

akarnokd:
I don't know, I can't switch to debug while running with coverage enabled. Maybe there is a way to trigger a thread dump externally?

danielcweber:
I could give it a try.

danielcweber:
Do you know what process will eventually host our code? Everything I debug doesn't run anything form System.Reactive.

danielcweber:
I can't reproduce the hanging tests under coverage. `Virtual_ThreadSafety()` running slow is understandable, it takes a lot of loops and `AdvanceBy` to actually trigger the virtual scheduler.

akarnokd:
Were you running dotCover or microsoft's own coverage tool of VS Enterprise?

danielcweber:
DotCover within VS

akarnokd:
There could be a probabilistic signal loss from the timer.

You could try and add a loop around the body of `Delay_TimeSpan_DefaultScheduler` for example, usually 1.000 - 10.000 iterations. 

```cs
        [Fact]
        public void Delay_TimeSpan_DefaultScheduler()
        {
            for (int i = 0;  i < 10000; i++)    
                Assert.True(Observable.Return(1).Delay(TimeSpan.FromMilliseconds(1))
                   .ToEnumerable().SequenceEqual(new[] { 1 }));
        }
```

danielcweber:
Can't really reproduce it but the fix is straightforward and the SpinWait-solution is wacky enough that this should be merged.

PR #580, opened 2018-06-09
4.x: Inline Disposable.Create

akarnokd:
This PR inlines the `Disposable.Create(Action)` invocations, saving on an allocation and some indirection.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
@@ -35,47 +35,47 @@ namespace System.Reactive.Linq.ObservableImpl
             {
                 private readonly Func<TFirst, TSecond, TResult> _resultSelector;
 
-                FirstObserver firstObserver;
-                IDisposable firstDisposable;
+                private readonly object _gate;
 
-                SecondObserver secondObserver;
-                IDisposable secondDisposable;
+                private readonly FirstObserver _firstObserver;
+                private IDisposable _firstDisposable;
+
+                private readonly SecondObserver _secondObserver;
+                private IDisposable _secondDisposable;
 
                 public _(Func<TFirst, TSecond, TResult> resultSelector, IObserver<TResult> observer)
                     : base(observer)
                 {
+                    _gate = new object();
+
+                    _firstObserver = new FirstObserver(this);
+                    _secondObserver = new SecondObserver(this);
+
+                    _firstObserver.Other = _secondObserver;
+                    _secondObserver.Other = _firstObserver;
+
                     _resultSelector = resultSelector;
                 }
 
-                private object _gate;
-
                 public void Run(IObservable<TFirst> first, IObservable<TSecond> second)
                 {
-                    _gate = new object();
-
-                    firstObserver = new FirstObserver(this);
-                    secondObserver = new SecondObserver(this);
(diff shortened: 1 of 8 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
```diff
@@ -47,32 +53,43 @@ public void Run(IObservable<TFirst> first, IObservable<TSecond> second)
...
+                {
+                    if (disposing)
+                    {
+                        Disposable.TryDispose(ref firstDisposable);
+                        Disposable.TryDispose(ref secondDisposable);
 
-                    SetUpstream(StableCompositeDisposable.Create(fstSubscription, sndSubscription, fstO, sndO));
+                        // clearing the queue should happen under the lock
+                        // as they are plain Queue<T>s, not concurrent queues.
+                        lock (_gate)
```
danielcweber: This has now become dangerous, _gate might not be assigned.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
```diff
@@ -556,22 +571,34 @@ public void Run()
...
-                SetUpstream(new CompositeDisposable(_subscriptions) { Disposable.Create(() => { foreach (var q in _queues) q.Clear(); }) });
+            protected override void Dispose(bool disposing)
+            {
+                if (disposing)
+                {
+                    for (int i = 0; i < _subscriptions.Length; i++)
+                    {
+                        Disposable.TryDispose(ref _subscriptions[i]);
+                    }
+                    lock (_gate)
```
danielcweber: This has now become dangerous, _gate might not be assigned.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
akarnokd: Can `Dispose` happen before `Run` executes?

danielcweber: Hopefully not in our code, but it's good to assume everything might be disposed right away. By the same reasoning, Run will definitely happen when the Sink was created, so the gate might be created while constructing.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
akarnokd: I'll move the creation into the constructor so it will be guaranteed to exist.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
```diff
@@ -571,30 +573,38 @@ public void Run()
...
-
-                _gate = new object();
+                var subscriptions = new IDisposable[N];
 
-                for (int i = 0; i < N; i++)
+                if (Interlocked.CompareExchange(ref _subscriptions, subscriptions, null) == null)
                 {
-                    var o = new SourceObserver(this, i);
-                    Disposable.SetSingle(ref _subscriptions[i], srcs[i].SubscribeSafe(o));
+                    _gate = new object();
```
danielcweber: I know it's safe now but please just make it a readonly field and assign it in the declaration.

PR #584, opened 2018-06-12
Copy .editorconfig from dotnet/corefx repo

danielcweber:
This repo should adhere to the coding guidlines defined in https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md. Assuming their .editorconfig exactly captures these guidelines, they have been copied here.


akarnokd:
Is there a way to have a list of nonconforming files and lines or it is required to open each .cs file?

danielcweber:
Investigating. So far I didn't find a way to scan the whole sln, however, ReSharper understands .editorconfig as well and is able to perform a full scan.

akarnokd:
I'm working on adding unit tests that verify the recent changes in components based on code coverage. I don't plan to change anything other than tests.

onovotny:
We do not need to follow their rules explicitly. Many of them sure, but for others like `var`, we don't need to be as strict as them. I'd go the other way and say we should always use var.

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
-dotnet_sort_system_directives_first = true
-# Avoid "this." and "Me." if not necessary
+# avoid this. unless absolutely necessary
 dotnet_style_qualification_for_field = false:suggestion
 dotnet_style_qualification_for_property = false:suggestion
 dotnet_style_qualification_for_method = false:suggestion
 dotnet_style_qualification_for_event = false:suggestion
 
-# Use language keywords instead of framework type names for type references
+# only use var when it's obvious what the variable type is
```
onovotny: I disagree with this this section

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+csharp_style_var_for_built_in_types = false:none
```
onovotny: should be `true:suggestion`

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+csharp_style_var_when_type_is_apparent = false:none
```
onovotny: should be `true:suggestion`

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+csharp_style_var_elsewhere = false:suggestion
```
onovotny: should be `true:suggestion`

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style    = pascal_case_style
+
+dotnet_naming_symbols.constant_fields.applicable_kinds   = field
+dotnet_naming_symbols.constant_fields.required_modifiers = const
+
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should have s_ prefix
```
onovotny: This rule is terrible...IMO.

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+
+# internal and private fields should be _camelCase
+dotnet_naming_rule.camel_case_for_private_internal_fields.severity = suggestion
+dotnet_naming_rule.camel_case_for_private_internal_fields.symbols  = private_internal_fields
+dotnet_naming_rule.camel_case_for_private_internal_fields.style    = camel_case_underscore_style
+
+dotnet_naming_symbols.private_internal_fields.applicable_kinds = field
+dotnet_naming_symbols.private_internal_fields.applicable_accessibilities = private, internal
+
+dotnet_naming_style.camel_case_underscore_style.required_prefix = _
```
onovotny: I also hate having an underscore prefix.

.editorconfig
```diff
@@ -1,79 +1,161 @@
...
+dotnet_naming_style.camel_case_underscore_style.capitalization = camel_case 
+
+# Code style defaults
+dotnet_sort_system_directives_first = true
+dotnet_style_require_accessibility_modifiers = always
```
onovotny: I don't like this but can go either way. I prefer `omit_if_default` so that accessibility modifiers are only present if they differ from what the language provides.

akarnokd:
It will probably create impossible cases, for example:

image

If you have `var` you have to default it to some type, which by definition will be ignored. Having `IDisposable[] previousDisposable;` will complain about not using `var`.

.editorconfig
danielcweber: ReSharper suggests upper camel case with no underscore, which I follow elswhere. Up to debate. This is what this thread is for.

.editorconfig
danielcweber: This, on the other hand, is preferred by me.

.editorconfig
danielcweber: I'd rather have it explicit to be clear for anybody who contributes.

.editorconfig
danielcweber: Also, this style is present throughout most of the repo, but I can change the config if everybody agrees.

danielcweber:
So, is there anything we definitely want to take from the corefx-config? We might as well settle with the original config or enhance that.

.editorconfig
danielcweber: Who prefers what for static fields?

.editorconfig
danielcweber: @akarnokd and me settled on keeping underscores as a compromise, for safety reasons, to make code more readably outside VS (github, git diff, etc) and since it's used throughout Rx sources today.

.editorconfig
danielcweber: Changed to `omit_if_default` to have less redundancies.

.editorconfig
danielcweber: The `s_`-prefix requirement has been dropped. Static fields should be pascal cased.

danielcweber:
@onovotny We have settled on some changed. Do you think we can go with the proposed .editorconfig?

onovotny:
Looks good, the only suggestion I would add is to specify yaml/yml as indent size of 2. Otherwise, LGTM.

danielcweber:
@onovotny Can you access the build-logs? I can't, that has been happening occasionally for a while now.

onovotny:
I don't see the logs either but am looking into it via other avenues.

PR #588, opened 2018-06-14
4.x: Check for license header and add them

akarnokd:
This PR adds an unit test that scans most `*.cs` files in the `Rx.NET` project for the existence of a license header:

```cs
// Licensed to the .NET Foundation under one or more agreements.

```

It creates a list of such files and by default, the header is added to those files. The logic excludes certain type of `cs` files such as generated ones and designer ones.

I'm not 100% certain for the file path detection part. Works for me locally, which assumes the work directory will be somewhere below `Rx.NET`.

### DIFF
--- a/Rx.NET/Source/facades/GlobalAssemblyVersion.cs
+++ b/Rx.NET/Source/facades/GlobalAssemblyVersion.cs
@@ -1,4 +1,8 @@
-﻿
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+
 using System.Reflection;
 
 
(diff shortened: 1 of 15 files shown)


danielcweber:
I have seen this included into CI in other projects at a really early stage in the pipelline . Do you think that would be an option?

akarnokd:
I don't know how to do that. In RxJava, we have a plugin that checks this but it is treated as warnings only. I haven't found one that can actually add the missing header so RxJava has this test as well which when run locally will add the headers too. There is no reason to do it on VSTS because the sources are not synced back to the repo anyway.

danielcweber:
I'd really prefer a solution that works on the CI level. Maybe there's a task on the marketplace for that...

akarnokd:
A quick search only brings up a VS extension for adding licenses manually but it doesn't verify if there is one. I can't find anything among the VSTS tasks either. This PR can fail on CI if there are missing headers as demonstrated by the first commit.

PR #590, opened 2018-06-14
4.x: Improve blocking First & Last operators

akarnokd:
Optimize the blocking `First` and `Last` (== `Wait`) by inlining the various classes into one container and not waiting if the operation terminated by the time `Subscribe` returned.


Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reactive.Disposables;
+using System.Text;
+using System.Threading;
+
+namespace System.Reactive.Linq.ObservableImpl
+{
+
+    internal abstract class BaseBlocking<T> : CountdownEvent, IObserver<T>
```
danielcweber: CountdownEvent vs. ManualResetEvent ?

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        protected IDisposable _upstream;
+
+        internal T _value;
+        internal bool _hasValue;
+        internal Exception _error;
+
+        int once;
```
danielcweber: See above, ManualResetEvent would save this field.

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Blocking.cs
```diff
@@ -67,45 +67,25 @@ public virtual TSource FirstOrDefault<TSource>(IObservable<TSource> source, Func
...
-                    e =>
-                    {
-                        ex = e;
-                        evt.Set();
-                    },
-                    () =>
-                    {
-                        evt.Set();
-                    })))
+                consumer.SetUpstream(d);
```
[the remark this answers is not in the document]
akarnokd: The unblocking may happen any time later and the upstream would still push items to be dropped unnecessarily.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
akarnokd: `ManualResetEvent` cannot be extended, `CountdownEvent` can, saving on an allocation.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
akarnokd: Allocation vs an atomic operation tradeoff. If `Signal` wouldn't throw by default, we'd not need this.

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Blocking.cs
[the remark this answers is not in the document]
danielcweber: Any time later, how?

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+
+        public override void OnNext(T value)
+        {
+            if (!_hasValue)
+            {
+                this._value = value;
+                this._hasValue = true;
+                Disposable.TryDispose(ref _upstream);
```
danielcweber: Why no check here (or, why check above) ?

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Blocking.cs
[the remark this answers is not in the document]
akarnokd: Pausing and resuming a thread takes time, about 5 microseconds on Windows based on my experience. During this time, the source may generate anything between zero to 1000 items, all of them wasting resources.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
akarnokd: OnNext is called in a serialized fashion and a previous call already set `_hasValue` to true. Checking that upfront guarantees this executes at most once, but `OnCompleted` may still run regardless of `_upstream` disposed so that we want to avoid doing again there.

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Blocking.cs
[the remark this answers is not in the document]
danielcweber: Ok, you're taking that into account.

danielcweber:
I'm thinking about whether CountdownEvent.Dispose should be overridden to dispose the upstream, not that it would conceptually be required but simply because there is the disposable field and the instance is disposable.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
danielcweber: Why is the order (Unblock, Dispose) reversed?

akarnokd:
Calling CountdownEvent.Dispose breaks the CountdownEvent and it cannot be unblocked after that. Methods will start throwing ObjectDisposedException as well.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        internal LastBlocking() : base() { }
+
+        public override void OnCompleted()
+        {
+            Unblock();
+            Disposable.TryDispose(ref _upstream);
+        }
+
+        public override void OnError(Exception error)
```
danielcweber: Save for the IsDisposed-check, the overriding code is almost identical. Enough ground to push it to the base-class ?

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
akarnokd: No practical reason other than the reverse will race the upstream disposition with the `using` disposing.

Rx.NET/Source/src/System.Reactive/Linq/Observable/FirstLastBlocking.cs
akarnokd: Last doesn't need that dispose check so why have it there. First needs it so why penalize everybody else?

danielcweber: At least push some of the code into the base class.

danielcweber:
>Calling CountdownEvent.Dispose breaks the CountdownEvent

What I mean is, if CountdownEvent.Dispose is called, should the upstream be disposed?

akarnokd:
It can't be called here as the class is not exposed.

danielcweber:
Still, BaseBlocking implements IDisposable and holds a disposable, it should override and dispose of the upstream.

danielcweber:
Maybe it should be put in a using block then and honour that it implements IDisposable.

akarnokd:
No. It makes no sense in this context. You dispose the subscription returned by `Subscribe` in `using`.

danielcweber:
Thinking about that, it might definitely make sense since the CountdownEvent probably holds onto an unmanaged resource and should be disposed as soon you're done with it.

danielcweber:
WaitAndSetOnce wrapped a ManualResetEvent before and it's in a using-block.

PR #592, opened 2018-06-14
4.x: Inline AnonymousObserver use at places

akarnokd:
Inline the use of `AnonymousObserver` at various locations:

- `Subscribe` with `CancellationTokenSource`
- Remoting `Subscribe`.

Also make `Notification.OnCompletedNotification` static singleton; it is stateless so there is no reason to create an instance of it all the time.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/ObserverWithToken.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/ObserverWithToken.cs
@@ -1,0 +1,63 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information. 
+
+using System;
+using System.Collections.Generic;
+using System.Reactive.Disposables;
+using System.Text;
+
+namespace System.Reactive
+{
+    /// <summary>
+    /// Wraps another IObserver, relays signals to it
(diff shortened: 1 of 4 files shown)


Rx.NET/Source/src/System.Reactive/Internal/ObserverWithToken.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace System.Reactive
+{
+    /// <summary>
+    /// Wraps another IObserver, relays signals to it
+    /// and hosts an external IDisposable
+    /// to be disposed upon termination.
+    /// </summary>
+    /// <typeparam name="T">The element type of the sequence.</typeparam>
+    internal sealed class ObserverWithToken<T> : IObserver<T>
```
danielcweber: Can you relate that to AnonymousSafeObserver and SafeObserver and evaluate whether we can save some code/inherit/reuse/etc. here ?

akarnokd: Those have safeguards and need a disposable upfront. This is just relaying events and calling dispose on the resource.

akarnokd: You can try that after this PR.

danielcweber: Alright.

PR #593, opened 2018-06-14
4.x: Inline SerialDisposable usages

akarnokd:
This PR inlines the usage of the `SerialDisposable` class into a field and `Disposable.TrySetSerial` calls.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Catch.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Catch.cs
@@ -89,22 +89,27 @@ namespace System.Reactive.Linq.ObservableImpl
                 _handler = handler;
             }
 
-            private SerialDisposable _subscription;
+            bool _once;
+
+            private IDisposable _subscription;
 
             public override void Run(IObservable<TSource> source)
             {
-                _subscription = new SerialDisposable();
+                Disposable.TrySetSingle(ref _subscription, source.SubscribeSafe(this));
+            }
 
-                var d1 = new SingleAssignmentDisposable();
-                _subscription.Disposable = d1;
-                d1.Disposable = source.SubscribeSafe(this);
-
-                SetUpstream(_subscription);
+            protected override void Dispose(bool disposing)
+            {
+                if (disposing)
+                {
+                    Disposable.TryDispose(ref _subscription);
+                }
+                base.Dispose(disposing);
             }
 
             public override void OnError(Exception error)
             {
-                if (error is TException e)
+                if (!_once && error is TException e)
                 {
                     var result = default(IObservable<TSource>);
                     try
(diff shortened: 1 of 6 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Catch.cs
```diff
@@ -117,40 +122,14 @@ public override void OnError(Exception error)
                         return;
                     }
 
-                    var d = new SingleAssignmentDisposable();
-                    _subscription.Disposable = d;
-                    d.Disposable = result.SubscribeSafe(new HandlerObserver(this));
+                    _once = true;
+                    Disposable.TrySetSerial(ref _subscription, result.SubscribeSafe(this));
```
danielcweber: I might not ever get threading right and be stuck with using `lock` forever, but aren't we introducing a race here, since the subscription might invoke OnError on a different thread almost immediately, and _once is not protected by anything, it's not a volatile write and there's no locks.

akarnokd: Whenever `this` is handed to another thread, there is a barrier inside the scheduler or task start that establishes a happens-before, so setting `_once` to true should be always visible to the next call to  `OnError`.

danielcweber: Got reference?

akarnokd: http://afana.me/archive/2015/07/10/memory-barriers-in-dot-net.aspx/

danielcweber: Where is the specific part about `this` being passed to another thread?

akarnokd: It's implicit as it is just another value/variable: See the "Full memory barrier" section about what constitutes as barrier and thus anything set before that will be visible after (provided that no further changes happen on the original thread to those references).

danielcweber: But we don't know nothing about how the work will be unloaded to another thread, threre might not be a new thread or a Task.Start. It might just go onto an existing EventLoopScheduler and be picked up from there.

akarnokd: Such async boundaries will go through eventually a lock or an atomic CAS/Exchange which are full barriers. If they didn't everything else would break way before this field is ever read.

https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs#L154

danielcweber: Please, just make it a Volatile.Write for now and push that optimization to a separate PR. I'm really not convinced but my experience with lock-free programming is also limited. Thus I can't merge it as it is in good faith.

PR #595, opened 2018-06-15
4.x: Add reflection-based Argument-Null validation test

akarnokd:
This PR adds an automatic, reflection based validation that goes through an extension method host (such as `Observable` or `ObservableEx`) and tries to call the various methods with `null` arguments (depending on the argument type) to see if the implementation actually verifies it. In addition, if the return type is `IObservable`, the same null check is performed with `Subscribe(null)`.

This type of tests are scattered around in all operators already, so the only missing check found was in the new `ConcatMany` operator. Such automatic test helps in the future if new operators or overloads get added to the library and remind the developer for the need for such checks. The existing tests should remain as they are.

### DIFF
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information. 
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reactive;
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
@@ -1,0 +1,522 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using System.Reactive.Joins;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+using System.Reflection;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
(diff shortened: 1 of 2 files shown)


Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
```diff
@@ -0,0 +1,526 @@
...
+                        // watch out for type constrains
+                        // the default typeof(int) will not work when
+                        // exception or IDisposable is required at minimum
+                        var gac = ga[k].GetGenericParameterConstraints();
+                        // no type constraints
+                        if (gac.Length == 0)
+                        {
+                            targs[k] = typeof(int);
+                        }
+                        else
```
danielcweber: Brackets please

Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
```diff
@@ -0,0 +1,526 @@
...
+
+        #region + Verification method +
+
+        /// <summary>
+        /// Verify that public static members of the class
+        /// check for nulls in their arguments as
+        /// well as when invoking Subscribe with null.
+        /// </summary>
+        /// <param name="clazz">The class to verify.</param>
+        static void VerifyClass(Type clazz)
```
danielcweber: type instead of clazz, looks more like .net.

Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
```diff
@@ -0,0 +1,526 @@
...
+                        if (gac.Length == 0)
+                        {
+                            targs[k] = typeof(int);
+                        }
+                        else
+                        if (gac[0] == typeof(Exception))
+                        {
+                            targs[k] = typeof(Exception);
+                        }
+                        else
```
danielcweber: Brackets please

Rx.NET/Source/tests/Tests.System.Reactive/Tests/ArgumentValidationTest.cs
akarnokd: What do you mean? It's an else if construct.

danielcweber: See, that got me, I read that wrong. Please write else and if on the same line

PR #596, opened 2018-06-15
Use the Count-property on a list instead of enumerating it with Count().

danielcweber:


JohanLarsson:
`Enumerable.Count()` special cases on `ICollectoin`, just trivia. Using the property is of course nicer and faster.

PR #597, opened 2018-06-15
Capture the concept of a "safe observer"

danielcweber:
Capture the concept of a "safe observer" since it has been used and copied multiple times in the repo. ISafeObserver implements IDisposable and IObserver and can hold onto a resource. Implementors must at least dispose the resource on disposal of the SafeObserver, inheriting classes may choose to dispose the resource on other occasions as well. So far, AnonymousSafeObserver, SafeObserver and ObserverWithToken work like that are have been changed to implement ISafeObserver.


Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
```diff
@@ -46,7 +46,9 @@ public IDisposable SubscribeRaw(IObserver<TSource> observer, bool enableSafeguar
             //
             if (enableSafeguard)
             {
-                observer = SafeObserver<TSource>.Create(observer, subscription);
+                var safeObserver = SafeObserver<TSource>.Create(observer);
+                safeObserver.SetResource(subscription);
```
akarnokd: Could the initialization of this `subscription` also be eliminated just like in the other `Producer` variant (in another PR of course)?

danielcweber: Got it already covered in a separate PR!

PR #599, opened 2018-06-15
4.x: CurrentThreadScheduler fast path Schedule

akarnokd:
This PR modifies the schedule logic of the `CurrentThreadScheduler` to avoid allocating the queue for the first task on an idle scheduler. If the action schedules further tasks in a reentrant manner, those will get queued up just like now.

Since this scheduler is per thread, there is no need for locks/synchronization as each caller thread will only see its own thread-local structures and it can't cross-post to some other thread.


akarnokd:
The fast-path could be added to the timed case as well (i.e., sleep before executing the plain action, then trampoline if necessary).

Rx.NET/Source/src/System.Reactive/Concurrency/CurrentThreadScheduler.cs
```diff
@@ -61,7 +64,79 @@ private static TimeSpan Time
...
+
+        /// <summary>
+        /// Schedules an action to be executed immediately.
+        /// </summary>
+        /// <typeparam name="TState">The type of the state passed to the scheduled action.</typeparam>
+        /// <param name="state">State passed to the action to be executed.</param>
+        /// <param name="action">Action to be executed.</param>
+        /// <returns>The disposable object used to cancel the scheduled action (best effort).</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="action"/> is <c>null</c>.</exception>
+        public override IDisposable Schedule<TState>(TState state, Func<IScheduler, TState, IDisposable> action)
```
danielcweber: Could you pack it into the timed overload for the dueTime == 0 case?

Rx.NET/Source/src/System.Reactive/Concurrency/CurrentThreadScheduler.cs
```diff
@@ -86,22 +161,23 @@ public override IDisposable Schedule<TState>(TState state, TimeSpan dueTime, Fun
...
+            if (!running)
+            {
+                running = true;
                 try
                 {
                     Trampoline.Run(queue);
                 }
                 finally
                 {
                     SetQueue(null);
```
danielcweber: If the queue is sufficiently small (maybe < 16 or so), we could think about just clearing the queue instead of recreating one next time.

akarnokd: Nope, it doesn't work. There is no current capacity to `ScheduledQueue` and the `PriorityQueue` shrinks itself when the number of items is 25% of its internal capacity upon remove/dequeue.

Rx.NET/Source/src/System.Reactive/Concurrency/CurrentThreadScheduler.cs
```diff
@@ -77,32 +80,69 @@ public override IDisposable Schedule<TState>(TState state, TimeSpan dueTime, Fun
...
+            if (!running)
             {
-                queue = new SchedulerQueue<TimeSpan>(4);
-                queue.Enqueue(si);
+                running = true;
 
-                SetQueue(queue);
+                if (dueTime > TimeSpan.Zero)
+                {
+                    ConcurrencyAbstractionLayer.Current.Sleep(dueTime);
```
danielcweber: Why can we omit the normalization of dueTime here?

akarnokd: https://github.com/dotnet/reactive/blob/d6d954f08a0a38b95ece899834b8f03f0a352f55/Rx.NET/Source/src/System.Reactive/Concurrency/ConcurrencyAbstractionLayerImpl.Windows.cs#L67

https://github.com/dotnet/reactive/blob/11506de19ab875ba4f07e594ebd6f9da8ee69870/Rx.NET/Source/src/System.Reactive/Concurrency/ConcurrencyAbstractionLayerImpl.cs#L64

akarnokd: We don't mind if it is negative and Sleep will normalize it anyway.

PR #600, opened 2018-06-17
Add dedicated class based implementations for Append and Prepend.

quinmars:
Here is my attempt to optimize the `Append` and `Prepend` operators. Please review it carefully, I'm not sure if I got the handling of the `IDisposables` right.


Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
```diff
@@ -0,0 +1,66 @@
...
+            _source = source;
+            _value = value;
+            _scheduler = scheduler;
+        }
+
+        protected override _ CreateSink(IObserver<TSource> observer) => new _(this, observer);
+
+        protected override void Run(_ sink) => sink.Run(_source);
+
+        internal sealed class _ : IdentitySink<TSource>
```
akarnokd: It would be great if we abandoned these underscore class names.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
```diff
@@ -0,0 +1,66 @@
...
+            public override void OnCompleted()
+            {
+                var disposable = _scheduler.Schedule(this, ForwardValue);
+                Disposable.TrySetSingle(ref _schedulerDisposable, disposable);
+            }
+
+            private static IDisposable ForwardValue(IScheduler scheduler, _ sink)
+            {
+                sink.ForwardOnNext(sink._value);
+                return scheduler.Schedule(sink, ForwardOnCompleted);
```
akarnokd: Is it necessary to schedule the `OnCompleted` signal separately?

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            private static IDisposable ForwardOnCompleted(IScheduler _unused, _ sink)
+            {
+                sink.ForwardOnCompleted();
+                return BooleanDisposable.True;
```
akarnokd: This should not really be returned by operators and is supposed to be confined to `Disposable` methods. Return `Disposable.Empty` instead.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+
+            protected override void Dispose(bool val)
+            {
+                base.Dispose(val);
+                Disposable.TryDispose(ref _schedulerDisposable);
```
akarnokd: Usually we do this as `if (disposing) { dispose custom resources } base.Dispose(disposing)`.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
quinmars: I agree with you, but that name is used in most - if not all - operators I have seen.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
quinmars: Well, that was something I tried to ask in #567. The conclusion was to do it like `ToObservable`/`StartWith` do it. And `ToObservable` does schedule the `OnCompleted` separately.

akarnokd:
One thing to consider is that `Append` invites the following pattern:

```cs
IObservable<int> someSource = Observable.Return(0);
for (int i = 1; i < 1000; i++) {
   someSource = someSource.Append(i);
}

someSource.Subscribe(Console.WriteLine);
```

Does this still work with your modifications?

quinmars:
It is tested for three consequtive `Append`s in [`Append_Many_Scheduler`](https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/AppendTest.cs) , but I haven't added any optimizations as it is done in CoreFX for `Enumerable.Append`, i.e., merging consequtive operators to one, instead of using the naive `O(n²)` implementation. I can't even say if it becomes quadratic for observables or not.

akarnokd:
`TailRecursiveSink` in `Concat` flattens out the deep structure so it is not a problem there. When this PR is merged, I can add the changes so that a chain of append is coalesced into one `Concat(source, items)`.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
danielcweber: Definitely desirable, maybe in a later PR.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Append.cs
danielcweber: I don't think we need the separate schedule here.

quinmars:
We could go the same route, as the corefx team has done it for the enumerable counter part [1], to optimize consecutive `Append`s and `Prepend`s. That'd be:

 1. Collect all appends and prepends in an AppendPrependObservable class
 2. For the observing part: a sink that consist roughly of a `IdentitySink` that has a `ToObservable.Sink` starting part and a `ToObservable.Sink` ending part.

If we do that or anything else in that direction, we should propably skip the extra scheduling of the `OnCompleted`, at least between two `Append`s/`Prepend`s.

I can do that, but that'd take some time. 

[1] https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/AppendPrepend.cs

akarnokd:
It is certainly more interesting and challenging doing this type of "operator fusion". I leave it up to you if you want to tackle it.

quinmars:
I'd give it a try than, if that's ok for @danielcweber as well.

danielcweber:
Sure, go ahead.

danielcweber:
Looks good. Is there still something missing to make the WIP go away?

quinmars:
Not really. I first thought I put the optimizations, I'm working on, into this PR. But I think it is a good point to merge. I'll put the next commits in a new PR.

quinmars:
@akarnokd now the stacktrace should be much smaller, hopefully. 😄

quinmars:
@danielcweber, I changed my mind and put the changes into this PR. Let me know if I can help you during the review.

danielcweber:
Unfortunately, I get 4 failing tests with NullReferenceExceptions locally:

Append_Many, Append_Many_Scheduler, Append_Many_Take, Append_Many_Take_Scheduler.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
```diff
@@ -0,0 +1,357 @@
...
+                        // is used to have LoopRec bail out and perform proper clean-up of the
+                        // enumerator.
+                        //
+                        var flag = new BooleanDisposable();
+                        _scheduler.Schedule(new State(flag, this, array, continueWith), LoopRec);
+                        return flag;
+                    }
+                }
+
+                private sealed class State
```
danielcweber: Please replace this class by the use of [C#7 ValueTuples](https://blogs.msdn.microsoft.com/mazhou/2017/05/26/c-7-series-part-1-value-tuples/) to have less allocations.

quinmars:
Sorry, I was pretty sure that I did had run all tests the day before. I'll fix that.

quinmars:
Man, in that commit was nearly everything wrong that could go wrong, sorry for that. I hope it's now ok. At least all tests pass. I changed the class to be a struct like you did in #644.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
```diff
@@ -0,0 +1,355 @@
...
+                    if (parent._appends != null)
+                        _appends = parent._appends.ToReverseArray();
+                }
+
+                public void Run()
+                {
+                    if (_prepends != null)
+                    {
+                        var disposable = Schedule(_prepends, s => s.SetUpstream(s._source.SubscribeSafe(s)));
+                        Disposable.TrySetSerial(ref _schedulerDisposable, disposable);
```
danielcweber: That could indeed lead to trouble, in case the subscription (one line above) will complete synchronously, you will effetively be overwriting the scheduling from OnCompletd, thus cancelling that work. TrySetSingle should be fine.

quinmars: I changed it to `TrySetSingle`. Should I than use two different backing fields for the two schedulers? Currently, they share one field variable.

danielcweber: No, the one field should be fine, it's just the `TrySetSingle` should only be on the first scheduling. Now you're using it on both, which is not what I meant. In fact, if you change both to `SetSingle` instead of `TrySetSingle`, you will probably notice one of the tests throwing. The second `TrySetSingle` will fail in the case that we have prepends and appends. So we're currently leaking the second scheduling.

danielcweber: The second time you schedule something, you want it to overwrite the previous scheduling, which is why you want to use `TrySetSerial`. The first time you schedule, however, you want to either set the disposable for the very first time or not set it at all. Because, if the scheduled work executes synchronously, the second scheduling will take place (successfully through TrySetSerial), and by the time Schedule returns, we definitely don't want to overwrite that second scheduling, which is why we want to use TrySetSingle.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
```diff
@@ -0,0 +1,355 @@
...
+                private IDisposable _schedulerDisposable;
+
+                public _(AppendPrependMultiple<TSource> parent, IObserver<TSource> observer)
+                    : base(observer)
+                {
+                    _source = parent._source;
+                    _scheduler = parent.Scheduler;
+
+                    if (parent._prepends != null)
+                        _prepends = parent._prepends.ToArray();
```
akarnokd: Could we avoid the creation of this array by simply walking the node structure?

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
```diff
@@ -0,0 +1,355 @@
...
+
+                    if (parent._appends != null)
+                        _appends = parent._appends.ToReverseArray();
```
akarnokd: This array creation could be avoided as well if the appends where in a doubly-linked node chain. In that case, it would be enough to traverse the chain once to find the very first value and then walking backwards.

danielcweber: Beware, by doubly linking the nodes, we can't pass around the original observable once we appended/prepended to it, or even append/prepend something to the original observable. The instances we get from appying operators should be, in a way, immutable. We would get subtle side effects here, if I got your idea correctly.

danielcweber: I was thinking about getting the immutable collections back in (https://github.com/dotnet/reactive/pull/222), then it's just a stack and a queue.

akarnokd: > The instances we get from appying operators should be, in a way, immutable.

Indeed, the `_appends` should be an immutable list. I was somehow thinking it being a copy-on-write list (i.e., ImmutableList) which requires way more allocation and copying. My mistake.

danielcweber: If we manage to mitigate the performance degradation that has been reported in #222, immutable collections could be helpful in a lot of places. It's used internally in the Roslyn compiler since it's highly concurrent.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
quinmars: Yes, we could. I haven't done it  because we then have to duplicate the two loops. It's a tradeoff between code reuse and one allocation.

Rx.NET/Source/src/System.Reactive/Linq/Observable/AppendPrepend.cs
quinmars: The current implementation has a time complexity of O(n) and needs n + 1 allocations. I doubt the immutable collections offer a datatype that can beat that.

danielcweber: Adding a node to the lists as they are currently implemented are even O(1). Yes, even immutable collections won't beat that. Maybe they'll find their way into the project eventually, until that, this solution is perfectly fine.

PR #601, opened 2018-06-17
Use LastIndexOf instead of IndexOf in LicenseHeaderTest

quinmars:
My Rx git repository happens to be in the directory `Rx.NET`. So the resulting global source path is:

`C:\Users\peter\Documents\src\Rx.NET\Rx.NET\Source`.

Unfortunately, that makes the license header test fail, because in the test, the second occurance of `Rx.NET` is searched for but the first one is found.

The PR fix this by using `LastIndexOf`, addditionally the test will now fail, if there is no occurance at all instead of silently passing.


danielcweber:
Thx. I was running into the exact same problem. I was wondering whether we can avoid the IndexOf altogether and just ..\..\.. back our way to the right directory. It might not work on all environments though (dotnet/net/local/CI/etc). Would you like to explore this?

quinmars:
I think there cannot be a compile time constant for the source directory, at least it would not improve portability. The biggest problem I can imagine is that the tests are not excuted on the build computer but on a different device, e.g. a mobile device or different server. Then the test will fail. I haven't found a way to skip a test in xunit from code. That would probably be the best.

What I can offer is a very small improvement on the path creation.

PR #602, opened 2018-06-18
An allocation in BasicProducer.SubscribeRaw can be saved in most cases.

danielcweber:


Rx.NET/Source/src/System.Reactive/Internal/Producer.cs
```diff
@@ -38,31 +38,35 @@ public IDisposable Subscribe(IObserver<TSource> observer)
...
             if (enableSafeguard)
             {
-                var safeObserver = SafeObserver<TSource>.Create(observer);
-                safeObserver.SetResource(subscription);
-                observer = safeObserver;
+                observer = safeObserver = SafeObserver<TSource>.Create(observer);
             }
 
             if (CurrentThreadScheduler.IsScheduleRequired)
             {
```
danielcweber: @akarnokd Can you review again, it occured to me that on this path, the safeObserver is not assigned the resource. I guess I may have introduced a bug here.

danielcweber: Should be fixed by https://github.com/dotnet/reactive/pull/602/commits/a041e2cfc87a767ebc710ee2def4ddcbe2b6671d

PR #604, opened 2018-06-18
Rework SelectMany

danielcweber:
Save some allocations and inline disposables.


Rx.NET/Source/src/System.Reactive/Linq/Observable/SelectMany.cs
```diff
@@ -875,9 +872,7 @@ public override void Run(IObservable<TSource> source)
                 {
                     _isStopped = false;
```
akarnokd: I'm not sure why setting to default here is even necessary. The field is already initialized to false upon class instantiation. By removing this, you don't even need to override Run.

Rx.NET/Source/src/System.Reactive/Linq/Observable/SelectMany.cs
```diff
@@ -1110,9 +1109,7 @@ public override void Run(IObservable<TSource> source)
...
```
akarnokd: Same here, init to default can be omitted.

Rx.NET/Source/src/System.Reactive/Linq/Observable/SelectMany.cs
danielcweber: Right, I figured this might have been from a time where Sinks were supposed to run more than once.

akarnokd: Even then it makes no sense as sinks had per subscription state. Having the same sink instance service two indenpendent flows invites state clash and concurrency errors.

PR #610, opened 2018-06-19
4.x: Have Timeout(Absolute|Selector) use lock-free methods.

akarnokd:
Make the absolute and selector-based `Timeout` operators lock-free, similar to #546.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
@@ -346,12 +346,10 @@ namespace System.Reactive.Linq.ObservableImpl
                 }
             }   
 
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Timeout.cs
```diff
@@ -308,113 +273,121 @@ protected override void Dispose(bool disposing)
...
+            {
+                var timeoutObserver = new TimeoutObserver(this, idx);
+                if (Disposable.TrySetSerial(ref _timerDisposable, timeoutObserver))
+                {
+                    var d = timeout.Subscribe(timeoutObserver);
+                    timeoutObserver.SetResource(d);
+                }
+            }   
+
+            private sealed class TimeoutObserver : ISafeObserver<TTimeout>
```
danielcweber: Inherit from SafeObserver.

Issue #489, opened 2018-05-04
4.x: possible new operators based on RxJava's feature set


akarnokd:
I've collected the operators and overloads of theirs which are available in RxJava but (likely) not in Rx.NET yet to enable discussions about feature synchronization.

Operator in RxJava | Description
-----------|--------------
`combineLatestDelayError` | Keep combining sources as long as possible, then signal the aggregate exception at the end.
`concatDelayError` | Keep concatenating and signal the aggregate exception at the end.
`concatEager` | Run many or all source observables concurrently, but provide the items from the sources in their original order.
`concatEagerDelayError` | Run many or all source observables concurrently, provide items in order and signal the aggregate exception at the end.
`error(Callable)` | Signal an onError with the generated exception per observer.
`fromCallable` | Invoke a function to generate a single item & completion per observer.
`generate` | Generate events in a callback fashion with an observer-abstraction and per observer state.
`intervalRange` | Produce a range of numbers with an initial delay for the first followed by a periodic signal of the rest.
`mergeDelayError` | Merges some or all sources into one sequence and signals an aggregate exception when all of them terminated.
`switchOnNextDelayError` | Switch to newer sources, signal the aggregate exception at the end.
`zip(delayError)` | Zips sources into one sequence and signals the error after the last row could be combined.
`blockingIterable` | Turn a sequence into a blocking enumerable sequence.
`blockingSubscribe` | Observe on the caller thread in a blocking fashion.
`cache` | Consume the upstream once, cache and replay all items to observers.
`collect` | Aggregate items into a single, custom container per observer with the help of a collector action and signal the container as a single result.
`compose` | Applies a transformation function from observable to observable when the sequence is constructed.
`concatMap` | Maps the upstream items into observables and concatenates them in order; an in-sequence version of `concat`. Can be implemented more efficiently than `concat(map(source, f))`.
`concatMapDelayError` | Maps the upstream items into observables and concatenates them in order, signaling an aggregate exception at the end.
`concatMapEager` | Maps the upstream items into observables, runs some or all of them eagerly but emits their items in order; an in-sequence version of `concatEager`.
`concatMapEagerDelayError` | Maps the upstream items into observables, runs some or all of them eagerly but emits their items in order while signaling an aggregate exception at the end.
`concatMapIterable` | Maps the upstream items into enumerations and emits their items in order; more efficient than `concatMap`ping an `Iterable` turned into an observable.
[...]

PR #612, opened 2018-06-19
4.x: Add the TakeUntil(Func<T, bool>) operator

akarnokd:
Add the `TakeUntil` operator overload with a predicate that will complete the sequence after an item has been emitted and then the predicate returns true for it:

```cs
Observable.Range(1, 10)
   .TakeUntil(item => item == 5)
   .Subscribe(Console.WriteLine);
```

which prints 1 through 5 and completes.

There is a similar operator: `TakeWhile` which performs the check **before** the item is emitted.

This is one of the operators present in RxJava and mentioned in #489.


danielcweber:
The difference of `TakeUntil(item => item == 5)` to `TakeWhile(item => item != 5)` is that `5` will be included?

danielcweber:
This is highly useful, we should just make sure the difference from TakeUntil(p) to TakeWhile(!p) is very clear.

akarnokd:
Beyond the XML doc, I've tried to indicate with the `stopPredicate` naming that it is `true` that stops whereas in `TakeWhile`, `false` stops.

PR #615, opened 2018-06-19
Adding ns2.0 and netcoreapp2.0 targets to Ix.NET

jbogard:
Adding `netstandard2.0` and `netcoreapp2.0` targets to Ix.NET.

.NET Core 2.0 added [TakeLast](https://docs.microsoft.com/en-us/dotnet/api/system.linq.queryable.takelast?view=netcore-2.1) and [SkipLast](https://docs.microsoft.com/en-us/dotnet/api/system.linq.queryable.skiplast?view=netcore-2.1) methods so you need to exclude those methods against that target.

### DIFF
--- a/Ix.NET/Source/Directory.build.targets
+++ b/Ix.NET/Source/Directory.build.targets
@@ -11,6 +11,12 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.3'">
     <DefineConstants>$(DefineConstants);NO_CODE_COVERAGE_ATTRIBUTE;CRIPPLED_REFLECTION;PLIB;SIGNED</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <DefineConstants>$(DefineConstants);NO_CODE_COVERAGE_ATTRIBUTE;PLIB;SIGNED</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp2.0'">
+    <DefineConstants>$(DefineConstants);BUILT_IN_QUERYABLE_METHODS;NO_CODE_COVERAGE_ATTRIBUTE;PLIB;SIGNED</DefineConstants>
+  </PropertyGroup>
   <PropertyGroup Condition="'$(TargetFramework)' == 'net45'">
     <DefineConstants>$(DefineConstants);HAS_APTCA;NO_ARRAY_EMPTY;DESKTOPCLR;DESKTOPCLR45;SIGNED</DefineConstants>
   </PropertyGroup>
--- a/Ix.NET/Source/System.Interactive.Async.Providers/System.Interactive.Async.Providers.csproj
+++ b/Ix.NET/Source/System.Interactive.Async.Providers/System.Interactive.Async.Providers.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <Description>Interactive Extensions Async Providers Library used to build query providers and express queries over async enumerable sequences.</Description>
     <AssemblyTitle>Interactive Extensions - Async Providers Library</AssemblyTitle>
-    <TargetFrameworks>net45;netstandard1.0</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard1.0;netstandard2.0</TargetFrameworks>
     <PackageTags>Ix;Interactive;Extensions;Enumerable;Asynchronous</PackageTags>
   </PropertyGroup>
 
--- a/Ix.NET/Source/System.Interactive.Async.Tests/System.Interactive.Async.Tests.csproj
+++ b/Ix.NET/Source/System.Interactive.Async.Tests/System.Interactive.Async.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.1;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;netcoreapp2.0;net461</TargetFrameworks>
     <NoWarn>$(NoWarn);CS0618</NoWarn>
   </PropertyGroup>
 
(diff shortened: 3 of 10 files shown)


onovotny:
What did .NET Core 2 do? We can't reduce surface area in the platform since that'll cause missing method exceptions if you're built against the netstandard version.

jbogard:
See those two links - it added TakeLast and SkipLast IQueryable methods that now conflict with this library. So while they're not in `netstandard`, they *are* in `netcoreapp2.x` but *not* in any other `netstandard` targets.

If you leave the methods in, the `netcoreapp2.0` tests won't even compile, you get a `CS0121` error:

```
Tests.Qbservable.cs(135,61,135,69): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.QueryableEx.TakeLast<TSource>(System.Linq.IQueryable<TSource>, int)' and 'System.Linq.Queryable.TakeLast<TSource>(System.Linq.IQueryable<TSource>, int)'
```

onovotny:
This is a problem and there's no support for global aliases. @terrajobst @davidfowl what do we do? We cannot remove these methods from the netcoreapp surface area or it'll throw missing method's.

onovotny:
How is this any different than what Ix does today? If you reference Ix in a netcoreapp2.0/2.1 today, I think you'll have the same issue?

jbogard:
@onovotny yep, you do. If you just add a `netcoreapp2.0` target to the test projects for example, they won't compile:

```
"C:\dev\dotnet-reactive\Ix.NET\Source\Ix.NET.sln" (restore;build target) (1) ->
"C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj" (default target) (7:18) ->
"C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj" (Build target) (7:19) ->
(CoreCompile target) ->
  Tests.Qbservable.cs(135,61): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.QueryableEx.TakeLast<TSource>(System.Linq.IQueryable<TSource>, int)' and 'System.Linq.Queryable.TakeLast<TSource>(System.Linq.IQue ryable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(355,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.TakeLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.TakeLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(363,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.TakeLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.TakeLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(371,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.TakeLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.TakeLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(379,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.TakeLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.TakeLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(394,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.SkipLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.SkipLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(402,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.SkipLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.SkipLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]
  Tests.Single.cs(410,23): error CS0121: The call is ambiguous between the following methods or properties: 'System.Linq.EnumerableEx.SkipLast<TSource>(System.Collections.Generic.IEnumerable<TSource>, int)' and 'System.Linq.Enumerable.SkipLast<TSource>(Sy stem.Collections.Generic.IEnumerable<TSource>, int)' [C:\dev\dotnet-reactive\Ix.NET\Source\System.Interactive.Tests\System.Interactive.Tests.csproj]

    74 Warning(s)
    8 Error(s)
```

onovotny:
They won't redirect, they're tied to the implementation.

Suppose you have a library compiled against the netstandard2 or 1.x version that uses `QueryableEx.TakeLatest() from System.Interactive`. Now that library is used on netcoreapp2.x and gets the netcoreapp2.x version of `System.Interactive` that does not contain that method. It'll blow up.

You can use typeforwarders for entire types, but not methods on types.

onovotny:
It's ambiguous due to overload resolution of the extension methods, not because they're in the same type or assembly even. They are in the same namespace and have the same name and signature, thus ambiguous unless called with the type name. It'd be worse if the type names were the same since the CPS project system doesn't support assembly aliases today, but here the type names are still different. (Queryable vs QueryableEx)

jbogard:
ruh roh. Won't this come up soon with the C# 8 API proposal you mentioned?

jbogard:
And I'm a bit surprised this hasn't come up already, but it's only those two methods atm

onovotny:
I don't think there's been a conversation about how methods that were first defined in Ix move to the BCL. That should happen.

jbogard:
The other option here is to remove those two methods from `netstandard2.0` altogether, but that doesn't seem that great either.

onovotny:
Keeping this open to capture the discussion since we need to resolve it...

jbogard:
Nice.

PR #616, opened 2018-06-19
Build Ix with VSTS

onovotny:


akarnokd:
Will this build separately or will it always build whenever Rx.NET is changed/PRd?

onovotny:
Unfortunately there's no current way to specify PR validations per path, so it'll build. That said, we have five agents in the pool, so they'll run in parallel.

akarnokd:
I understand.

I may have already mentioned this, but sooner or later the Ix parts should be moved into their own repository, especially if the contributors are otherwise distinct from the Rx contributors. Otherwise, we'd step on each other's toes all the time. Failure in one part may hinder progress on the other, independent part. Issues and PRs for one or the other would get mixed all the time. etc.

onovotny:
Understood, though I don't expect much churn on Ix. We just need to update it to provide a netstandard2 build. It's still not getting a lot of work until C# 8, and we'll figure out where the code moves then.

PR #617, opened 2018-06-19
4.x: Make Scheduler.Recursive more uniform and allocate less

akarnokd:
This PR unifies the `Action`-based recursive scheduler helper by capturing all state into an instance and managing the recursive calls from there.

The original also allocated at least the `scheduler.Schedule(state2, (scheduler1, state3) =>` part due to the state capture of all those variables.


Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Recursive.cs
```diff
@@ -208,45 +137,175 @@ public static IDisposable Schedule<TState>(this IScheduler scheduler, TState sta
...
                     }
+                    var idx = ++index;
+                    g.Add(idx, Disposable.Empty);
+                    return idx;
                 }
-            });
+            }
 
-            recursiveAction(tuple.state);
+            protected void SetDisposable(long idx, IDisposable d)
```
danielcweber: That stuff with the long indexes is really ugly IMO. I don't particularly like these solutions that will eventually overflow, even if it's not in our lifetime. Can we keep the CompositeDisposable?

akarnokd: There is usually 1-2 entries in the map, the previous and the next schedule's disposable. These can't get further away than 1 so even if the counter wraps, that doesn't matter because they would still go into different slots.

danielcweber: To make it wrap the increment should be unchecked?

danielcweber: I don't really know. To keep using the CompositeDisposable it would take the allocation of a SingleAssignmentDisposable, but aren't we saving a ton here anyway?

akarnokd: I'm for the maximum saving.

danielcweber: No actually it doesn't take a SingleAssignmentDisposable, does it? Why is the lookup with a dictionary better?

danielcweber: I get this but I don't really like the approach with those weak identities. I am not convinced it's good code. I see it'll be working for a gazillion of years but still, I have to think about that. I'd rather have something that you can compare by reference.

danielcweber:
I'm really sorry if the introduction of SingleAssignmentDisposables is just another case of you 'giving up', and I would not want it to be that way. I have elaborated on my objectives on Slack. I hope you understand.

akarnokd:
The main problem is that the review creates ambiguities about how to proceed due to being formed as questions and eventually leading to an unconvincable state on the side of the reviewer.

danielcweber:
Point taken. That's just me not wanting to tell you what to do.

akarnokd:
It is a matter of decision and taking responsibility for those decisions and the subsequent changes requested by a reviewer. Being a collaborator with merge rights is an elevated position that requires clear view on what should and shouldn't be done and if necessary, explain why. Posing such responses though as more and more questions that create an illusion that the contributor can change the reviewer's mind will lead to confusion, and perhaps even worse, self-censorship.

danielcweber:
Well, as I said, point taken. So let me state my view: I think that the solution with ulongs is too brittle in theory (albeit, as we have already discussed, not in practice) to justify the saving of an allocation, therefore I would prefer the SingleAssignmentDisposable solution.

PR #622, opened 2018-06-19
Add coverage

onovotny:
Add code coverage to Ix.NET via VSTest


onovotny:
Yes, still waiting to configure the exclusions -- it's including the test assemblies and fluent assertions test helpers.

Issue #625, opened 2018-06-19
Investigate skipped tests in Ix


onovotny:
Share4, Finally6.

akarnokd:
I think `Share4` was disabled because `e1.Dispose` has no effect on the `IEnumerator` as it is implemented via a `yield return` state machine. I don't know if it is even possible or not.

This could be fixed via a manual state machine.

As for `Finally6`, the `Finally()` part does not run. I see 3 cases:

1) I don't know how long a cancellation token is supposed to be valid for a `MoveNext` so the test might be too fast completing the first task such that `Cancel` has no effect.
2) The `token` could be reacted to until the next `MoveNext` is invoked thus even if a value is ready, cancel still propagates.
3) `Finally` could be registering with the token which it doesn't do currently.

onovotny:
Thanks @akarnokd.

I agree that's the cause for `Share4`. In the interest of correctness, it's probably worth turning this into a state machine.

Not sure what the right thing is yet for `Finally6`

akarnokd:
I'll post a PR shortly for the two.

PR #627, opened 2018-06-19
Ix: Allow Share to be disposed, Make Finally react to token cancellation

akarnokd:
The `Share4` was disabled because disposing a yield return-based IEnumerator has no effect. The code has been turned into a manual implementation.

The `Finally6` was disabled likely because the cancellation token could not trigger the finally action in time or due to a missing token registration. The code has been extended to react to the token as well as making sure the operator remains safe with now a possible concurrent `MoveNextCore` and `Dispose`.

Fixes: #625

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async/Finally.cs
+++ b/Ix.NET/Source/System.Interactive.Async/Finally.cs
@@ -28,6 +28,10 @@ namespace System.Linq
 
             private IAsyncEnumerator<TSource> enumerator;
 
+            CancellationTokenRegistration _tokenRegistration;
+
+            int _once;
+
             public FinallyAsyncIterator(IAsyncEnumerable<TSource> source, Action finallyAction)
             {
                 Debug.Assert(source != null);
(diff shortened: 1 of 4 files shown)


Ix.NET/Source/System.Interactive.Async/Finally.cs
```diff
@@ -65,19 +78,36 @@ public override void Dispose()
                         goto case AsyncIteratorState.Iterating;
 
                     case AsyncIteratorState.Iterating:
-                        if (await enumerator.MoveNext(cancellationToken)
-                                            .ConfigureAwait(false))
+                        // clear any previous registration
+                        _tokenRegistration.Dispose();
```
onovotny: Is this pattern something we need in the base Iterator?

akarnokd: The base iterator does something similar but that registration ends when the `MoveNext` method returns. I think it is the expected base behavior for all operators, so extending the cancellability window this way is likely unique to this operator. Alternatively, the test could changed so that the first item arrives much later so cancelling the token has a chance.

akarnokd: https://github.com/dotnet/reactive/blob/master/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs#L87

onovotny: Makes sense

onovotny:
VSTS is having an issue...will merge once they resolve it. Thanks! 🎉🎉🎉

PR #628, opened 2018-06-19
Add .NET Standard 2 to Ix

onovotny:
Pending discussion on #615


jbogard:
So what happens for folks using these methods on `netcoreapp2.1`? Will they still get compile errors, and need to explicitly call one or the other w/o an extension method?

onovotny:
Nope, through the magic of reference assemblies, `netcoreapp2.0`, `netcoreapp2.1` and `netstandard vnext` all get a reference assembly that excludes those methods.

However, the lib, the runtime, still has all the methods and that's what gets loaded. What this means is that at a recompile, it'll bind to the BCL version, but code that already has a binary reference to Ix, will continue to work because the method is still there in the runtime library.

jbogard:
Ah, just saw your blog, couldn’t tell from your PR.

PR #633, opened 2018-06-20
Simplify thread pool scheduler timed schedule

danielcweber:


Rx.NET/Source/src/System.Reactive/Concurrency/ThreadPoolScheduler.cs
```diff
@@ -190,76 +169,6 @@ public void Dispose()
...
-
-                try
-                {
-                    timer._disposable.Disposable = timer._action(timer._parent, timer._state);
-                }
-                finally
-                {
-                    SpinWait.SpinUntil(timer.IsTimerAssigned);
-                    timer.Stop();
-                }
```
quinmars: As far as I understand it, the original code uses here try-catch-guard to  guarantee that the assignment of `_timer` happens and it waits in the `Tick()` method, until the _timer is  really set. Is this not needed in the new version?

danielcweber: It's IMO a poor way of achieving synchronization. We can do better in UserWorkItem. If you have a look at UserWorkItem, the backing field of CancelQueueDisposable is only touched by the `Disposable`-helper methods. These ensure that, no matter what order the calls occur (assignment/disposal or disposal/assignment), we get the desired result. `Disposable` does it with `Interlocked` methods.

quinmars: Ah, now I understand. so

```csharp
Disposable.TryDispose(ref _field);
Disposable.TrySetSingle(ref _field, disposable);
```

Will dispose `disposable`, that's very clever.

PR #639, opened 2018-06-21
Review Merge, save some allocations.

danielcweber:


Rx.NET/Source/src/System.Reactive/Linq/Observable/Merge.cs
```diff
@@ -224,39 +200,45 @@ public override void OnCompleted()
...
                         }
                     }
 
-                    public void OnCompleted()
+                    public override void OnCompleted()
                     {
-                        _parent._group.Remove(_self);
-                        if (_parent._isStopped && _parent._group.Count == 1)
+                        _parent._group.Remove(this);
+                        if (_parent._isStopped && _parent._group.Count == 0)
```
akarnokd: Same possible problem as with `SelectMany`: `_isStopped` is plain and can be reordered after `_group.Count`.

danielcweber: Are there any current open issues that could be explained by this going wrong sometimes?

akarnokd: Do you want me to post an issue which demonstrates this race condition?

danielcweber: Now that we're fixing it, no I wouldn't suggest that. We could ask @onovotny about whether we should put out another preview to have something to point people to, should someone encounter the issue.

akarnokd: I could not trigger this specific reordering on my machine but I found another issue via #660.

danielcweber: That's already on the fix branch ?

akarnokd: The main branch. This PR did not change the algorithm so the test should fail as well. You could try it in this PR as the test is standalone.

danielcweber: I guess I'll merge this, then make the failing test (and subsequent fix) a (failing) new WIP-PR.

akarnokd: Tried with your branch. #660 still fails.

PR #640, opened 2018-06-21
Save some more allocations in SelectMany.

danielcweber:


Rx.NET/Source/src/System.Reactive/Linq/Observable/SelectMany.cs
```diff
@@ -144,18 +138,18 @@ public void OnNext(TCollection value)
...
                         }
                     }
 
-                    public void OnCompleted()
+                    public override void OnCompleted()
                     {
-                        _parent._group.Remove(_self);
-                        if (_parent._isStopped && _parent._group.Count == 1)
+                        _parent._group.Remove(this);
+                        if (_parent._isStopped && _parent._group.Count == 0)
```
akarnokd: I'd be more comfortable if `_isStopped` was `volatile` here. `_group.Count` is volatile but the memory model allows moving the plain read of `_isStopped` after it, which creates a window where `Count` is zero at the moment, the main source creates one more group and sets `_isStopped`, then this thread reads `_isStopped` and thinks it can terminate. 

Reference: http://afana.me/archive/2015/07/10/memory-barriers-in-dot-net.aspx/ "Implicit barriers: acquire".

danielcweber: That makes sense.

danielcweber: There is even more allocations to save. I will revise the PR in a moment.

PR #642, opened 2018-06-22
Less allocations in SubscribeOnCtxObservable

danielcweber:


Rx.NET/Source/src/System.Reactive/Concurrency/Synchronization.cs
```diff
@@ -102,27 +102,49 @@ public static IObservable<TSource> SubscribeOn<TSource>(IObservable<TSource> sou
...
+                    _observer = observer;
+
+                    context.PostWithStartComplete(
+                        @this =>
+                        {
+                            if (!Disposable.GetIsDisposed(ref @this._cancel))
+                            {
+                                Disposable.SetSingle(ref @this._cancel, new ContextDisposable(@this._context, @this._source.SubscribeSafe(@this._observer)));
+                            }
+                        },
```
quinmars: Maybe you can convert the lambda into method. I find the code with all the `@this` hard to read.

danielcweber: I couldn't come up with another name that would capture that we're really dealling with `this`. I could change it. I would, however, not like to remove the lambda because apart from that, it makes for good readability IMO. We could pass a method group instead but due to [this](https://github.com/dotnet/roslyn/issues/5835), that would cost an allocation.

PR #644, opened 2018-06-22
Save some allocations in ToObservable

danielcweber:
Done by using a ValueTuple instead of a dedicated state class and passing state to IScheduler.Schedule.


danielcweber:
Revised it.

PR #647, opened 2018-06-25
Review join observer

danielcweber:


Rx.NET/Source/src/System.Reactive/Joins/JoinObserver.cs
```diff
@@ -51,18 +50,18 @@ public void Dequeue()
...
                     if (notification.Kind == NotificationKind.OnError)
                     {
-                        onError(notification.Exception);
+                        _onError(notification.Exception);
                         return;
                     }
 
                     Queue.Enqueue(notification);
-                    foreach (var activePlan in activePlans.ToArray())
+                    foreach (var activePlan in _activePlans.ToArray())
```
akarnokd: Could you add small a comment here stating that `_activePlan` can change during the iteration and thus it is necessary to work on a snapshot of the plans?

PR #649, opened 2018-06-25
Use var where appropriate.

danielcweber:


mpostol:
Let me know why using `var` instead of `int` adds any value. From the very beginning type in-front of a variable was used to check the type compatibility. Using `var` doesn't change the variable type, but do not allow using the compiler to test direct programmer expectation against indirect value type retrieved from the expression. Type compatibility is a very easy way to find mistakes and stupid issues. 

I know this problem is not related to the project, but changes like this don't really fix the code. This way we can change the code over and over making hundred unimportant changes.

danielcweber:
The value added here is reducing redundancy. I agree the change is rather small, but code rots if not touched, so as C# progresses, the codebase should too. We could argue about implicit vs. explicit typing, both is ok, but having a mix is not. And the codebase was 99% var, the .editorconfig suggests using var and that, in turn has been taken from the corefx project.

What do you mean about the compiler not being able to test direct programmer expectation? I think it's the opposite. By using var, I have one possibility less to contradict myself, and that's a good thing.

mpostol:
To explain my students `why about the compiler not being able to test direct programmer expectation` I use a LINQ example. Suppose the programmer expects `IQerable`, but the source variable only implements `IEnumerable`. The explicit decoration of the variable as `IQerable` causes the compiler error - it is a rad lamp that something is wrong. Using `var` to decorate the variable silently mitigate the problem for a while, because as a consequence the programmer expects that the expression will be translated to SQL and sent to DBMS for execution. Unfortunately, instead, the expression is executed locally on the selected object. Using `var` we have hidden fundamental difference of the software behavior ?!

The whole story is illustrated by the code in an example I am using to prepare a videobook:

[Linq](https://github.com/mpostol/TP/tree/master/Linq)

> Note: My students are allowed to use `var` only for anonymous types, i.e. only where it is really necessary.  

I know it is not a popular statement and MSDN is flooded by examples in contradiction with it.

mpostol:
Probably you have forgotten about the master rule: using testing you can only prove that there are errors. Testing is not enough to prove that the software is correct. There is always one more error :).

danielcweber:
@mpostol I can't directly see why this is a problem with `var`. If I can `var`-assign something from a source variable and I go on assuming it's an `IQueryable`, but it's not, isn't there something wrong with the source variable, or how it is declared?

mpostol:
At last, you know that there is a problem that must be addressed. It seems that the misunderstanding is that what you are calling `to contradict myself` I am coiling `double checking`. 
Sorry for commencing this academic discussion related to best practice, but I appreciate what experts say. For someone with a lot of experience assessing the expression type is usually a simple task. For the novice, it isn't and explicit typing could help.
Thanks for the conversation.

onovotny:
@mpostol appreciate the discussion, but to be clear, `var` is always statically typed. Most IDE's will provide a hint about the resolved concrete type, though I'll concede that it can be harder for offline/printed code. I'll also say that I believe the benefits of clarity outweigh uses of people printing code.

Edrigo:
In my opinion VAR shouldn't be used. Let the referees work at the World Cup...

onovotny:
I'm going to lock this thread as we don't need to continue this here.

PR #655, opened 2018-06-25
Review Delay-operator

danielcweber:


Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
```diff
@@ -74,18 +72,17 @@ public override void Run(TParent parent)
 
                     RunCore(parent);
 
-                    Disposable.SetSingle(ref _sourceSubscription, parent._source.SubscribeSafe(this));
+                    base.Run(parent._source);
```
akarnokd: I suggest removing those init-to-default lines above this and move `_queue = ...` into the constructor/field initializer.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Delay.cs
```diff
@@ -300,17 +295,17 @@ public override void Run(TParent parent)
...
```
akarnokd: Same here, no need to initialize fields to their default value.

danielcweber: I removed the default value initializations.

PR #661, opened 2018-06-26
4.x: Add ConcatEager & ConcatManyEager operators

akarnokd:
This PR adds the `ConcatEager` and `ConcatManyEager` operators that allow running multiple sources concurrently but still emit items in an ordered fashion like `Concat`. It means that items from the non-active sources are buffered until their turn.

`ConcatEager` is implemented as `ConcatManyEager(sources, v => v)`.

The operator supports delaying all errors until all sources terminate and allows specifying the maximum number of inner sources to run at once. When one terminates and also has been fully consumed, the a new source will be started.

Proposed in #489.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
@@ -1115,7 +1115,7 @@ namespace System.Reactive.Linq
         /// Concatenates a sequence of observables eagerly by running some
         /// or all of them at once and emitting their items in order.
         /// </summary>
-        /// <typeparam name="T">The value type of the inner observables.</typeparam>
+        /// <typeparam name="TSource">The value type of the inner observables.</typeparam>
         /// <param name="sources">The sequence of observables to concatenate eagerly.</param>
         /// <param name="maxConcurrency">The maximum number of observables to run at a time.</param>
         /// <exception cref="ArgumentNullException"><paramref name="source"/> is null.</exception>
@@ -1141,8 +1141,8 @@ namespace System.Reactive.Linq
         /// Maps the upstream items into observables, runs some or all of them at once, emits items from one
         /// of the observables until it completes, then switches to the next observable.
         /// </summary>
-        /// <typeparam name="T">The value type of the upstream.</typeparam>
-        /// <typeparam name="R">The output value type.</typeparam>
+        /// <typeparam name="TSource">The value type of the upstream.</typeparam>
+        /// <typeparam name="TResult">The output value type.</typeparam>
         /// <param name="source">The source observable to be mapper and concatenated eagerly.</param>
         /// <param name="mapper">The function that returns an observable for an upstream item.</param>
         /// <param name="maxConcurrency">The maximum number of observables to run at a time.</param>
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
@@ -1118,7 +1118,8 @@ namespace System.Reactive.Linq
         /// <typeparam name="T">The value type of the inner observables.</typeparam>
         /// <param name="sources">The sequence of observables to concatenate eagerly.</param>
         /// <param name="maxConcurrency">The maximum number of observables to run at a time.</param>
-        /// <param name="capacityHint">The number of items expected from each observable.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="maxConcurrency"/> is non-positive.</exception>
         /// <returns>An observable sequence that eagerly runs some or all inner sources but relays
         /// their elements still in order.</returns>
         public static IObservable<TSource> ConcatEager<TSource>(this IObservable<IObservable<TSource>> sources, bool delayErrors = false, int maxConcurrency = int.MaxValue)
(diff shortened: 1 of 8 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
```diff
@@ -1109,6 +1109,63 @@ public static IObservable<TSource> Concat<TSource>(this IObservable<Task<TSource
...
+        /// or all of them at once and emitting their items in order.
+        /// </summary>
+        /// <typeparam name="TSource">The value type of the inner observables.</typeparam>
+        /// <param name="sources">The sequence of observables to concatenate eagerly.</param>
+        /// <param name="maxConcurrency">The maximum number of observables to run at a time.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="maxConcurrency"/> is non-positive.</exception>
+        /// <returns>An observable sequence that eagerly runs some or all inner sources but relays
+        /// their elements still in order.</returns>
+        public static IObservable<TSource> ConcatEager<TSource>(this IObservable<IObservable<TSource>> sources, bool delayErrors = false, int maxConcurrency = int.MaxValue)
```
danielcweber: 'magic' values like `int.MaxValue` should be discouraged in times of nullable value types, please use 'int? maxConcurrency = null).

Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
```diff
@@ -1109,6 +1109,63 @@ public static IObservable<TSource> Concat<TSource>(this IObservable<Task<TSource
...
+        /// <typeparam name="TSource">The value type of the upstream.</typeparam>
+        /// <typeparam name="TResult">The output value type.</typeparam>
+        /// <param name="source">The source observable to be mapper and concatenated eagerly.</param>
+        /// <param name="mapper">The function that returns an observable for an upstream item.</param>
+        /// <param name="maxConcurrency">The maximum number of observables to run at a time.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="source"/> or <paramref name="mapper"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="maxConcurrency"/> is non-positive.</exception>
+        /// <returns>An observable sequence that eagerly runs some or all mapped-in sources but relays
+        /// their elements still in order.</returns>
+        public static IObservable<TResult> ConcatManyEager<TSource, TResult>(this IObservable<TSource> source, Func<TSource, IObservable<TResult>> mapper, bool delayErrors = false, int maxConcurrency = int.MaxValue)
```
danielcweber: Same as above, assuming that `int.MaxValue` means 'no maximum concurrency'.

Rx.NET/Source/src/System.Reactive/Linq/Observable.Multiple.cs
akarnokd: Is there a precedence for such design in Rx.NET?

/cc @onovotny

danielcweber: I guess so, after all, what this really wants to express is 'I don't want a maximum concurrency'. Nullable types allow just that, not specifying anything. It would also make the whole range of integers valid values, meaning more robust code.

quinmars: The precedence would be `Merge` where it is solved with separate methods. I undestand, however, that you do not want to add 3 additional, very similar methods.

quinmars:
As I've already written in #489, I don't like the `Many` part of the name very much.

danielcweber:
If the difference to Concat is only in the eagerness, then Many doesn't make too much sense, I agree.

quinmars:
The `Many` means here: with selector.

glopesdev:
Agree with adding overloads. There is already enough confusion with the sheer number of operators and their complicated naming choices, refraining from adding anymore seems like in general a very good choice.

danielcweber:
I also think that the functionality presented here is very special but that's just my two cents, I have never come across a situation where I would have needed eager evaluation. 

It's a philosophical question whether this lib should just contain basic operators. However, the effort should not be wasted. If this wasn't called 'reactive extensions', I would propose for an extra lib with that name. Maybe we should have 'reactive extensions extensions' ;-). Something 'official' with much lower barriers for new operators.

danielcweber:
I know. We also have our [own extensions lib](https://github.com/ExRam/ExRam.Extensions/blob/master/ExRam.Extensions/System/Reactive/Linq/ObservableExtensions.cs) (no nuget but open sourced). But only the lazy RefCount was broad enough to have a PR here, I figured. The rest is pretty special to our own needs.

akarnokd:
The drawback of having an external library is that `Producer`/`Sink` are inaccessible, thus Rx.NET considers those operators as unsafe/untrusted, applying safety wrappers around it.

glopesdev:
> If this wasn't called 'reactive extensions', I would propose for an extra lib with that name. Maybe we should have 'reactive extensions extensions' ;-)

@RxDave had the [Rxx](https://github.com/RxDave/Rxx) project going for a while, which was exactly what you described.

quinmars:
>> If this wasn't called 'reactive extensions', I would propose for an extra lib with that name. Maybe we should have 'reactive extensions extensions' ;-)

> @RxDave had the Rxx project going for a while, which was exactly what you described.

MoreRx like [MoreLINQ](https://github.com/morelinq/MoreLINQ) could be an option.

PR #664, opened 2018-06-27
Save allocations of closures and allow delegate caching in Windows code that schedules async work.

danielcweber:


Rx.NET/Source/src/System.Reactive/Platforms/Windows/ThreadPoolTimerExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the LICENSE file in the project root for more information. 
+
+#if WINDOWS
+using System.Reactive.Disposables;
+
+namespace System.Reactive.Concurrency
+{
+    internal static class ThreadPoolTimerExtensions
+    {
+        public static IDisposable AsDisposable(this global::Windows.System.Threading.ThreadPoolTimer threadPoolTimer)
```
akarnokd: Is there an ambiguity without `global::` here?

danielcweber: I guess the ambiguity is with other classes, which is why it's all put in the global namespace. I can do some cleanup.

PR #665, opened 2018-06-27
4.x: Improve Buffer via dedicated exact/skip implementations

akarnokd:
This PR improves the counted `Buffer` operator by having dedicated exact (count == skip) and gapped (count < skip) implementations because these cases require only one buffer to be handled instead of the overhead of a queue holding at most one buffer.

The PR also adds a benchmark to verify the gains:
```
 // * Summary *

BenchmarkDotNet=v0.10.14, OS=Windows 7 SP1 (6.1.7601.0)
Intel Core i7-4770K CPU 3.50GHz (Haswell), 1 CPU, 8 logical and 4 physical cores

Frequency=3418007 Hz, Resolution=292.5682 ns, Timer=TSC
  [Host]     : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.255 8.0
  DefaultJob : .NET Framework 4.7.1 (CLR 4.0.30319.42000), 64bit RyuJIT-v4.7.255 8.0

// Before
  Method |     Mean |    Error |   StdDev |   Gen 0 | Allocated |
-------- |---------:|---------:|---------:|--------:|----------:|
   Exact | 451.4 us | 2.592 us | 2.425 us | 49.8047 | 205.57 KB |
    Skip | 426.2 us | 4.431 us | 4.145 us | 40.5273 | 166.51 KB |
 Overlap | 459.6 us | 3.067 us | 2.869 us | 49.8047 | 205.57 KB |


// After
  Method |     Mean |    Error |   StdDev |   Gen 0 | Allocated |
-------- |---------:|---------:|---------:|--------:|----------:|
   Exact | 410.2 us | 1.822 us | 1.615 us | 49.8047 | 205.43 KB |
    Skip | 400.0 us | 2.446 us | 2.168 us | 40.5273 | 166.37 KB |
 Overlap | 458.8 us | 2.248 us | 2.102 us | 49.8047 | 205.57 KB |
```

Improvement of **Exact**: 9.1%
Improvement of **Skip**: 6.1%

The **Overlap** (aka original algorithm) is practically the same within the noise range.

I've tried with a non-modulo algorithm for Overlap but it consistently produced 10 us slower results. I'm not sure why; I assume the special small numbers in the benchmark (1, 2) were JIT'd into binary-and and thus was much faster than an increment-test-store approach like the other two.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
@@ -10,45 +10,196 @@ namespace System.Reactive.Linq.ObservableImpl
 {
     internal static class Buffer<TSource>
     {
-        internal sealed class Count : Producer<IList<TSource>, Count._>
+        internal sealed class CountExact : Producer<IList<TSource>, CountExact.ExactSink>
         {
-            private readonly IObservable<TSource> _source;
-            private readonly int _count;
-            private readonly int _skip;
+            readonly IObservable<TSource> _source;
 
-            public Count(IObservable<TSource> source, int count, int skip)
+            readonly int _count;
+
+            public CountExact(IObservable<TSource> source, int count)
+            {
+                _source = source;
+                _count = count;
+            }
+
+            protected override ExactSink CreateSink(IObserver<IList<TSource>> observer) => new ExactSink(observer, _count);
+
+            protected override void Run(ExactSink sink) => sink.Run(_source);
+
+            internal sealed class ExactSink : Sink<TSource, IList<TSource>>
+            {
+                readonly int _count;
+
+                int _index;
+
+                IList<TSource> _buffer;
+
+                internal ExactSink(IObserver<IList<TSource>> observer, int count) : base(observer)
+                {
+                    _count = count;
+                }
+
+                public override void OnNext(TSource value)
+                {
+                    var idx = _index;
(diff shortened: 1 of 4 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
```diff
@@ -10,45 +10,196 @@ namespace System.Reactive.Linq.ObservableImpl
...
+                internal ExactSink(IObserver<IList<TSource>> observer, int count) : base(observer)
+                {
+                    _count = count;
+                }
+
+                public override void OnNext(TSource value)
+                {
+                    var idx = _index;
+                    var buffer = _buffer;
+                    if (idx == 0)
```
quinmars: I'd prefer if you test here for buffer ==  null. I know it's the same, but when I read it the first time without knowing the end I was a little bit alarmed.

akarnokd: That would be logically incorrect because once the short buffer ends after count, there is a window between count + 1 .. skip where buffer is null and no new buffer should be created.

akarnokd: Wait, the conversation page shows your comment for line 46 but the changes page shows this for line 125:

image

PR #666, opened 2018-06-27
4.x: Remove an indirection observer in Case()

akarnokd:
This PR changes so that it directly uses the downstream's `IObserver` without wrapping it into a layer of `IdentitySink`. I think this avoids an allocation if I understood the recent internal changes to sinks and producers.

(Remark: it extends the `BasicProducer` because the full `Producer` requires an `IDisposable` sink type, which this PR tries to avoid.)


danielcweber:
See https://github.com/dotnet/reactive/pull/675

PR #667, opened 2018-06-27
4.x: Improve Collect() internals

akarnokd:
This PR improves the internals of the `Collect()` operator (also suggested by a comment in it). See the various justifications inline.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
@@ -18,37 +18,27 @@ namespace System.Reactive.Linq.ObservableImpl
             _getNewCollector = getNewCollector;
         }
 
-        protected override PushToPullSink<TSource, TResult> Run(IDisposable subscription)
-        {
-            var sink = new _(this, subscription);
-            sink.Initialize();
-            return sink;
-        }
+        protected override PushToPullSink<TSource, TResult> Run(IDisposable subscription) => new _(_merge, _getNewCollector, _getInitialCollector(), subscription);
 
         private sealed class _ : PushToPullSink<TSource, TResult>
         {
-            // CONSIDER: This sink has a parent reference that can be considered for removal.
+            readonly object _gate;
+            readonly Func<TResult, TSource, TResult> _merge;
+            readonly Func<TResult, TResult> _getNewCollector;
 
-            private readonly Collect<TSource, TResult> _parent;
-
-            public _(Collect<TSource, TResult> parent, IDisposable subscription)
+            public _(Func<TResult, TSource, TResult> merge, Func<TResult, TResult> getNewCollector, TResult collector, IDisposable subscription)
                 : base(subscription)
             {
-                _parent = parent;
+                _gate = new object();
+                _merge = merge;
+                _getNewCollector = getNewCollector;
+                _collector = collector;
             }
 
-            private object _gate;
             private TResult _collector;
-            private bool _hasFailed;
             private Exception _error;
             private bool _hasCompleted;
             private bool _done;
-
-            public void Initialize()
-            {
-                _gate = new object();
-                _collector = _parent._getInitialCollector();
-            }
 
             public override void OnNext(TSource value)
             {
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -18,50 +18,39 @@ public Collect(IObservable<TSource> source, Func<TResult> getInitialCollector, F
             _getNewCollector = getNewCollector;
         }
 
-        protected override PushToPullSink<TSource, TResult> Run(IDisposable subscription)
-        {
-            var sink = new _(this, subscription);
-            sink.Initialize();
-            return sink;
-        }
+        protected override PushToPullSink<TSource, TResult> Run(IDisposable subscription) => new _(_merge, _getNewCollector, _getInitialCollector(), subscription);
```
akarnokd: As the comment suggested, disconnect the parent from the sink and simply pass on the parameter functions as well as generate the initial value via direct call.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -18,50 +18,39 @@ public Collect(IObservable<TSource> source, Func<TResult> getInitialCollector, F
...
                 : base(subscription)
             {
-                _parent = parent;
+                _gate = new object();
+                _merge = merge;
+                _getNewCollector = getNewCollector;
+                _collector = collector;
             }
 
-            private object _gate;
```
akarnokd: `_gate` should have been readonly, moved to top.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -18,50 +18,39 @@ public Collect(IObservable<TSource> source, Func<TResult> getInitialCollector, F
...
             private TResult _collector;
-            private bool _hasFailed;
             private Exception _error;
             private bool _hasCompleted;
             private bool _done;
 
-            public void Initialize()
```
akarnokd: No need to have an initialization method when only the result of the `_getInitialCollector()` is actually needed by the sink.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -18,50 +18,39 @@ public Collect(IObservable<TSource> source, Func<TResult> getInitialCollector, F
...
-                _parent = parent;
+                _gate = new object();
+                _merge = merge;
+                _getNewCollector = getNewCollector;
+                _collector = collector;
             }
 
-            private object _gate;
             private TResult _collector;
-            private bool _hasFailed;
```
akarnokd: Failure is when `_error` is not null, no need to have a separate bool field for it as well.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -93,18 +81,19 @@ public override bool TryMoveNext(out TResult current)
             {
                 lock (_gate)
                 {
-                    if (_hasFailed)
+                    var error = _error;
+                    if (error != null)
                     {
-                        current = default(TResult);
-                        _error.Throw();
+                        current = default;
```
akarnokd: The `_collector` should probably be cleared here too, but the original code didn't do it for some reason. Should I add it?

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
```diff
@@ -93,18 +81,19 @@ public override bool TryMoveNext(out TResult current)
...
+                        error.Throw();
                     }
                     else
                     {
                         if (_hasCompleted)
                         {
                             if (_done)
                             {
-                                current = default(TResult);
+                                current = default;
```
akarnokd: The `_collector` should probably be cleared here too, but the original code didn't do it for some reason. Should I add it?

akarnokd:
Note that a newer PR, #676, also changes this file so that PR should take priority if the changes are accepted.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
danielcweber: Yes

Rx.NET/Source/src/System.Reactive/Linq/Observable/Collect.cs
danielcweber: Yes

akarnokd: I'll close this and change #676.

PR #669, opened 2018-06-27
4.x: Remove an indirection observer from Defer()

akarnokd:
This PR changes so that it directly uses the downstream's `IObserver` without wrapping it into a layer of `IdentitySink`. I think this avoids an allocation if I understood the recent internal changes to sinks and producers.

(Remark: it extends the `BasicProducer` because the full Producer requires an `IDisposable` sink type, which this PR tries to avoid.)

Similar improvement: #666

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Defer.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Defer.cs
@@ -6,7 +6,7 @@ using System.Reactive.Disposables;
 
 namespace System.Reactive.Linq.ObservableImpl
 {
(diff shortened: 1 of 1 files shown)


akarnokd:
I don't know who could decide upon this, maybe @bartdesmet ? There are still a lot of implications in the Rx.NET code which I can't determine them being intentional or just remnants of an old style. Sometimes there is a comment for that (such as the `enableSafeguard` in `BasicProducer` L45).

As I mentioned, there is still the `SubscribeSafe` wrapping and the actual `IObservable` is also likely using a `Sink` wrapper to talk to the downstream.

PR #670, opened 2018-06-27
4.x: DistinctUntilChanged() don't init to default values

akarnokd:
There is no need to initialize fields to their default values.


akarnokd:
Ready for merge.

PR #671, opened 2018-06-27
4.x: ElementAt don't create exception after finding the item

akarnokd:
When `ElementAt` finds an item, it disposes the upstream, however, the upstream may still emit an `OnCompleted` if it is unable to detect this dispose call (or ignores it). In this case, the `ElementAt`'s `OnCompleted` gets invoked which creates an exception that gets thrown away as the base observer is already the no-op observer, wasting resources creating the stacktrace in the process.

I don't know how to verify this via unit test, but this extra call can be seen when placing a breakpoint in `ElementAt`'s `OnCompleted` method and running this flow:

```cs
Observable.Return(1).ElementAt(0).Subscribe();
```


danielcweber:
Just to clarify, a stacktrace isn't created until the exception is actually thrown.

PR #675, opened 2018-06-27
4.x: Remove an indirection observer from If()

akarnokd:
This PR changes `If` so that it directly uses the downstream's `IObserver` without wrapping it into a layer of `IdentitySink`. I think this avoids an allocation if I understood the recent internal changes to sinks and producers.


danielcweber:
Don't know. You lose all the auto-muting features that Sinks provide. I guess the BasicProducers are just for the cases where you don't subscribe to an upstream observable.

danielcweber:
`SafeObserver` (or `WrappingSafeObserver` for that matter) doesn't do the muting, it's the Sinks that do.

PR #676, opened 2018-06-27
4.x: Improve PushPullAdapter+Sink dispose management

akarnokd:
Inline the `SingleAssignmentDisposable` into the `PushPullSink` (similar to how `Sink`s were inlined).

This PR also includes the logic changes from #667 as the `Collect()` operator also uses the push-pull adapter.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/PushToPullAdapter.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/PushToPullAdapter.cs
@@ -21,23 +21,17 @@ namespace System.Reactive.Linq.ObservableImpl
 
         public IEnumerator<TResult> GetEnumerator()
         {
-            var d = new SingleAssignmentDisposable();
-            var res = Run(d);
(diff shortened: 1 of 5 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/PushToPullAdapter.cs
```diff
@@ -21,23 +21,17 @@ public PushToPullAdapter(IObservable<TSource> source)
 
         public IEnumerator<TResult> GetEnumerator()
         {
-            var d = new SingleAssignmentDisposable();
-            var res = Run(d);
-            d.Disposable = _source.SubscribeSafe(res);
+            var res = Run();
+            res.SetUpstream(_source.SubscribeSafe(res));
```
danielcweber: Please evaluate inheriting PushToPullSink from Sink and using base.Run here.

akarnokd: `Sink` is not good here as there is no downstream `IObserver` to talk to.

PR #685, opened 2018-06-28
Make non-recursive schedulings explicit.

danielcweber:
In preparation for https://github.com/dotnet/reactive/pull/598, we evaluate where to replace IScheduler.Schedule by the ScheduleAction extension which does not allow recursive scheduling.


quinmars:
AppendPrependSingle.AppendValue would also be a candidate.

danielcweber:
I omitted that because PrependValue is not (yet) a candidate. I'll fix that later.

danielcweber:
Added more explicit `ScheduleAction`s.

danielcweber:
Turns out there are quite a few places that don't need to schedule recursively and would benefit from #598 if the optimization would live in the `ScheduleAction` extensions.

PR #697, opened 2018-06-28
4.x: TakeLast() improve structure and recursive scheduling

akarnokd:
- Inline the relevant parent fields into the sinks
- Use a more efficient recursive scheduling disposable-tracking scheme
- Delegate to base.Run where applicable.


Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeLast.cs
```diff
@@ -123,37 +131,36 @@ public Time(IObservable<TSource> source, TimeSpan duration, IScheduler scheduler
...
+                    _loopScheduler = parent._loopScheduler;
                     _queue = new Queue<System.Reactive.TimeInterval<TSource>>();
                 }
 
-                private IDisposable _sourceDisposable;
                 private IDisposable _loopDisposable;
                 private IStopwatch _watch;
 
-                public void Run()
+                public void Run(IObservable<TSource> source, IScheduler scheduler)
```
akarnokd: `scheduler` is only needed to get the `StartStopwatch` before the operator executes.

Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeLast.cs
```diff
@@ -56,32 +56,40 @@ protected override void Dispose(bool disposing)
...
+                    }
                 }
 
-                private void LoopRec(Action<_> recurse)
+                private IDisposable LoopRec(IScheduler scheduler)
                 {
                     if (_queue.Count > 0)
                     {
                         ForwardOnNext(_queue.Dequeue());
-                        recurse(this);
```
danielcweber: Can you elaborate more on the issue with this?

akarnokd: It is the very same situation with `Range` and `ToObservable`: the Action-based recursive scheduling is more expensive.

https://github.com/dotnet/reactive/pull/684/files#diff-76edecc51f6af371eae34be29849253aR62

danielcweber: Any possibility to optimize the action based recursive scheduling? The syntax is much less noisy.

akarnokd: I don't see how. It has to maintain some state across schedules. There was an optimization for it in #617, particularly the `InvokeRec1State` which holds a `CompositeDisposable` and creates a `SingleAssignmentDisposable` for each schedule invocation.

akarnokd:
Is there something else needed for this PR?

Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeLast.cs
onovotny: Would local functions help? You can capture the state without needing a delegate.

akarnokd: I think that still makes an allocation just like a capturing delegate would upon each invocation.

onovotny: Yes, it does still have a state variable for the capture

danielcweber: I think the issue here is, why does the action based scheduling have to keep track of all the schedulings in a `CompositeDisposable`. Seems odd, at least for tail recursion.

PR #700, opened 2018-06-28
4.x: Improve Timer() scheduler handling

akarnokd:
- Use a field `_index` to store the value to be emitted between schedule calls.
- Use stateful `Schedule` calls where appropriate.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Timer.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Timer.cs
@@ -120,6 +120,8 @@ namespace System.Reactive.Linq.ObservableImpl
             {
                 private readonly TimeSpan _period;
 
+                long _index;
+
                 public _(TimeSpan period, IObserver<long> observer)
                     : base(observer)
                 {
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Timer.cs
```diff
@@ -120,6 +120,8 @@ internal sealed class _ : IdentitySink<long>
             {
                 private readonly TimeSpan _period;
 
+                long _index;
```
danielcweber: Good without volatile ?

akarnokd: Yes, the schedulers won't execute `Tick` and `Tock` concurrently.

akarnokd: There is a `ForwardOnNext` in both `Tick` and `Tock`, which has always required those methods to run non-concurrently with themselves and the schedulers ensure that by default. The fact that they now take the value of a field is irrelevant because reading/writing that field followed by the `ForwardOnNext` call still has to be non-concurrent.

PR #710, opened 2018-06-28
Add non-recursive extensions for IScheduler allowing to return an IDisposable

danielcweber:
This will enable more usages of non-recursive scheduling. Before C# 7.3, the compiler will complain about some ambiguities, so we increase the language version.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Simple.cs
@@ -68,6 +68,27 @@ namespace System.Reactive.Concurrency
         }
 
         /// <summary>
+        /// Schedules an action to be executed.
+        /// </summary>
+        /// <param name="scheduler">Scheduler to execute the action on.</param>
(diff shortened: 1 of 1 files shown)


akarnokd:
"System.Security.VerificationException : Operation could destabilize the runtime."?

danielcweber:
Maybe https://github.com/dotnet/roslyn/issues/27382. Investigating.

danielcweber:
Removing `[assembly: AllowPartiallyTrustedCallers]` makes it disappear. I have to investigate which direction to go here, the language version has to be able to progress.

danielcweber:
More reference: https://twitter.com/jonskeet/status/1002876525119754240

danielcweber:
Removing `readonly` from [this](https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Linq/Observable/GroupBy.cs#L16) also does the trick.

onovotny:
Thread says to remove APTCA:
https://twitter.com/onovotny/status/1012392223755980800

onovotny:
And the docs are changing to say this directly too: https://github.com/dotnet/docs/pull/6199

PR #711, opened 2018-06-28
[WIP] Increment the used C# language version to 7.3.

danielcweber:
Probably due to the new features (readonly-structs etc.) some tests related to GroupBy will now fail. The error has been described [here](https://github.com/dotnet/roslyn/issues/27382). It's directly related to [this readonly struct field](https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Linq/Observable/GroupBy.cs#L16), removing the readonl modifier solves the issue. What exact message the error is trying to get across is still unclear, the workaround presented in this commit however fixes it.


onovotny:
We can, and should, still get rid of the AllowPartiallyTrustedCallers attribute as well.

PR #712, opened 2018-06-28
4.x: simplify default(X) usages

akarnokd:
- Simplify `default(X)` where `X` is inferable from the usage.
- `Aggregate`: 
  - don't initialize to default values
  - clear accumulation upon failure or completion
- `LastAsync`: 
  - don't initialize to default values
  - clear accumulation upon failure or completion
- `LastOrDefaultAsync`: 
  - don't initialize to default values
  - clear accumulation upon failure or completion
- Upgrade Test.System.Reactive project to C# 7.1

### DIFF
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests.System.Reactive.csproj
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests.System.Reactive.csproj
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>net46;netcoreapp2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);CS0618</NoWarn>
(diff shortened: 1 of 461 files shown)


Rx.NET/Source/tests/Tests.System.Reactive/Tests.System.Reactive.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFrameworks>net46;netcoreapp2.0</TargetFrameworks>
     <NoWarn>$(NoWarn);CS0618</NoWarn>
+    <LangVersion>7.1</LangVersion>
```
onovotny: This should be removed. The LangVersion is set to `latest` in the Directory.Build.props. This lowers it.

PR #713, opened 2018-06-28
Update config

onovotny:
This .editorconfig is mostly the same as the existing one but rearranged.

PR also contains code fixes using VS 2017 15.8 p3's format document that applies the styles.


akarnokd:
I won't be able to check this until next morning.

onovotny:
That's fine, I'm still fixing styles

akarnokd:
It would be great if this got merged as it prevents further developments due to likely conflicts.

PR #714, opened 2018-06-28
4.x: Cleanup unnecessary type arguments/usings

akarnokd:
- Remove unnecessary `using` statements
- Remove type arguments that can be inferred from the context
- Remove unnecessary casts
- Change some `x as Y` into patterns of `x is Y z`


akarnokd:
You mean they show up as warnings in the error list. One of the difficulties was to have a list of them so it can be walked. Had to enable full solution analysis and then actually open 100 .cs files to get them show up.

akarnokd:
Closing in favor of #713

onovotny:
VS 2017 15.8 preview 3 added support for applying code/style fixes based on editorconfig to the FormatDocument command:

image

There's no batch support yet, but I used this extension to format all `.cs` files:
https://marketplace.visualstudio.com/items?itemName=munyabe.FormatAllFiles

It was slow to apply to the solution since it effectively opens each file, applies the command, and saves, but it was something you could walk away from.

PR #717, opened 2018-06-29
4.x: Cleanup unnecessary type arguments/usings

akarnokd:
- Remove unnecessary using statements
- Remove type arguments that can be inferred from the context
- Remove unnecessary casts
- Change some x as Y into patterns of x is Y z


Rx.NET/Source/src/Microsoft.Reactive.Testing/Microsoft.Reactive.Testing.csproj
```diff
@@ -7,6 +7,7 @@
     <DefineConstants>$(DefineConstants);PLATFORM_DOTNET;XUNIT_VISIBILITY_INTERNAL</DefineConstants>        
     <PackageTags>Rx;Reactive;Extensions;Observable;LINQ;Events</PackageTags>    
     <Description>Reactive Extensions (Rx) for .NET - Testing Library</Description>
+    <LangVersion>7.1</LangVersion>
```
onovotny: Please remove this from here. It's set in the top-level Directory.Build.props

PR #722, opened 2018-06-29
Save some allocations and allow delegate caching in ScheduleAsync.

danielcweber:


Rx.NET/Source/src/System.Reactive/Concurrency/Scheduler.Async.cs
```diff
@@ -10,6 +10,36 @@ namespace System.Reactive.Concurrency
...
     public static partial class Scheduler
     {
+        private sealed class AsyncInvocation<TState> : IDisposable
+        {
+            private readonly CancellationTokenSource _cts = new CancellationTokenSource();
+            private IDisposable _run;
+
+            public AsyncInvocation(IScheduler self, TState s, Func<IScheduler, TState, CancellationToken, Task<IDisposable>> action)
+            {
+                action(new CancelableScheduler(self, _cts.Token), s, _cts.Token).ContinueWith(
```
akarnokd: I wouldn't execute this in the constructor in case the class initialization hasn't become properly visible at this point. This could be a plain method returning `this` and simply use `return new AsyncInvocation().Run(self, s, action);`

danielcweber: Good point. Fixed in a minute.

PR #726, opened 2018-06-29
4.x: Clean unnecessary arguments, fix field namings, use patterns, inline out vars

akarnokd:
- Remove type arguments inferable from the context
- Remove unnecessary method/type qualifiers
- Add missing visibility modifiers
- Remove unused imports
- Correct naming of private and public fields
- Use patterns instead of `as`
- Inline out vars


Rx.NET/Source/tests/Tests.System.Reactive/Tests/ImmutableListTest.cs
```diff
@@ -1,10 +1,12 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+#if SIGNED
```
onovotny: Hard to see this in context, but was there a reason SIGNED was added?

akarnokd: The test class is conditioned on `SIGNED`. See #542.

PR #727, opened 2018-06-29
4.x: System.Reactive fix field names, type args, usings, operator infrastructure

akarnokd:
- Add missing `_` to private fields
- Rename internal fields to start with a capital letter
- Remove type arguments inferable from context
- Remove unnecessary `this.` and `base.` prefixes
- Remove unnecessary namespace prefixes
- Remove unused imports
- Confine certain imports to a narrower `#if <label>` cases
- Use `IdentitySink` with `Amb` and `AmbMany`
- Use `object` instead of `Object`
- Use `string` instead of `String`
- Use `decimal` instead of `Decimal`
- Use `int` instead of `Int32`
- Use `long` instead of `Int64`
- Move the common `Action _onCompleted` field of `ActionPlan<...>` into the abstract base class `ActionPlan`


Rx.NET/Source/src/System.Reactive/Concurrency/CurrentThreadScheduler.cs
```diff
@@ -12,7 +12,7 @@ namespace System.Reactive.Concurrency
     /// <seealso cref="Scheduler.CurrentThread">Singleton instance of this type exposed through this static property.</seealso>
     public sealed class CurrentThreadScheduler : LocalScheduler
     {
-        private static readonly Lazy<CurrentThreadScheduler> s_instance = new Lazy<CurrentThreadScheduler>(() => new CurrentThreadScheduler());
+        private static readonly Lazy<CurrentThreadScheduler> _staticInstance = new Lazy<CurrentThreadScheduler>(() => new CurrentThreadScheduler());
```
quinmars: When I understand the .editorconfig right, than `static readonly` fields should be in PascalCase. Although it does not state if this is the case independently of the accessibility modifiers. @onovotny used PascalCase  in #720.

akarnokd: There is a property with name `Instance` already.

quinmars: Personally I also prefer to start private fields with an underscore else you have to become creative for backing fields. Like `LazyInstance` for example.

onovotny: We can always bring `s_` back for static instance fields, but I thought that was nasty.

danielcweber:
@akarnokd Sorry, I attempted to fix the conflicts and missed out on `scheduler` vs. `Scheduler`.

PR #728, opened 2018-07-02
Mark fields as readonly as suggested by ReSharper.

danielcweber:


Rx.NET/Source/src/System.Reactive/Linq/ObservableEx.cs
```diff
@@ -13,7 +13,7 @@ namespace System.Reactive.Linq
     public static class ObservableEx
     {
 #pragma warning disable IDE1006 // Naming Styles: 3rd party code is known to reflect for this specific field name
-        private static IQueryLanguageEx s_impl = QueryServices.GetQueryImpl<IQueryLanguageEx>(new QueryLanguageEx());
+        private static readonly IQueryLanguageEx s_impl = QueryServices.GetQueryImpl<IQueryLanguageEx>(new QueryLanguageEx());
```
akarnokd: This is a "hack" entry point. Remember why we couldn't inline the IQueryLanguage? Because RxSpy and others use this type of fields to inject their instrumentation.

Rx.NET/Source/src/System.Reactive/Linq/Observable_.cs
```diff
@@ -10,7 +10,7 @@ namespace System.Reactive.Linq
     public static partial class Observable
     {
 #pragma warning disable IDE1006 // Naming Styles: 3rd party code is known to reflect for this specific field name
-        private static IQueryLanguage s_impl = QueryServices.GetQueryImpl<IQueryLanguage>(new QueryLanguage());
+        private static readonly IQueryLanguage s_impl = QueryServices.GetQueryImpl<IQueryLanguage>(new QueryLanguage());
```
akarnokd: This is a "hack" entry point. Remember why we couldn't inline the IQueryLanguage? Because RxSpy and others use this type of fields to inject their instrumentation.

Rx.NET/Source/src/System.Reactive/Linq/ObservableEx.cs
danielcweber: Alright, I though they did it through the `PlatformEnlightenmentProvider`. That should still be possible though.

We should try to get in touch with the RxSpy folks though soon.

Rx.NET/Source/src/System.Reactive/Linq/Observable_.cs
danielcweber: I'll revise that in a minute.

Rx.NET/Source/src/System.Reactive/Linq/ObservableEx.cs
akarnokd: https://github.com/niik/RxSpy/blob/ab0dad44f999a9434afdef581fb55085207642bb/RxSpy/RxSpySession.cs#L136

danielcweber:
This has been revised.

PR #730, opened 2018-07-02
4.x: ComparisonBenchmarks solution

akarnokd:
This PR adds a standalone netcoreapp2.1 solution with BenchmarkDotnet and Rx.NET as a NuGet dependency and should enable comparative benchmark runs via changing the dependency on the actual Rx.NET version.

Example comparative benchmark for `Range`:

```
BenchmarkDotNet=v0.10.14, OS=Windows 10.0.17134
Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
Frequency=3513588 Hz, Resolution=284.6094 ns, Timer=TSC
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
```

#### Rx.NET 3.1.1

 Method |       N |           Mean |       Error |      StdDev |      Gen 0 |    Allocated |
------- |--------:|---------------:|------------:|------------:|-----------:|-------------:|
  Range |       1 |       1.391 us |   0.0059 us |   0.0052 us |     0.4406 |      1.81 KB |
  Range |      10 |       4.460 us |   0.0136 us |   0.0127 us |     1.1368 |      4.67 KB |
  Range |     100 |      34.751 us |   0.1546 us |   0.1446 us |     7.7515 |     31.86 KB |
  Range |    1000 |     336.790 us |   1.0798 us |   1.0100 us |    73.2422 |    300.08 KB |
  Range |   10000 |   3,470.563 us |   8.8414 us |   8.2702 us |   726.5625 |   2980.11 KB |
  Range |  100000 |  34,452.518 us | 169.7244 us | 150.4562 us |  7250.0000 |  29785.81 KB |
  Range | 1000000 | 341,178.302 us | 563.2819 us | 439.7734 us | 72687.5000 | 297838.53 KB |

#### Rx.NET 4.0.0

 Method |       N |           Mean |       Error |      StdDev |      Gen 0 |   Allocated |
------- |--------:|---------------:|------------:|------------:|-----------:|------------:|
  Range |       1 |       1.115 us |   0.0073 us |   0.0068 us |     0.4082 |     1.68 KB |
[...]

PR #731, opened 2018-07-02
Enable switching Rx versions for benchmarks.

danielcweber:
This PR makes the referenced Rx-version switchable. It works by force-overriding the used Rx-version (if taken from nuget). It can be combined with the standalone solution from https://github.com/dotnet/reactive/pull/730, the solution should then have configurations that match the project configurations (current, Rx3.1.1, etc), so the config can easily be switched in the VS dropdown.


akarnokd:
I'm not sure how this would work with #730 as it is a separate solution and has no access to CURRENT.

Besides, I found it much easier to keep the NuGet manager open and switch the Rx library with one dropdown and one update click. This requires clicking on the releases dropdown, clicking on the configuration manager, clicking on the configuration dropdown, clicking on it again because the first click just selects the cell, click on the setting, close the window and finally clean the solution.

Regardless, I don't see a reason to keep both of them around and copy benchmarks between the two.

danielcweber:
>I'm not sure how this would work with #730 as it is a separate solution and has no access to CURRENT.

Unless you reference the project in the separate solution.

>Besides, I found it much easier to keep the NuGet manager open and switch the Rx library with one dropdown and one update click.

That's what I'm saying, the hybrid of #730 and 731 can have both, access to the current sources and easier access to the configuration.

I'm good with either.

PR #737, opened 2018-07-03
Hybrid of #730 and #731

danielcweber:
Hybrid between #730 and #731: The projects are referenced in their own solution, the configuration is easily accessible by dropdown (one click!, no cleanup), the current sources are available. File structure cleanup will be needed.


akarnokd:
It still makes no sense to me to have both.

danielcweber:
No need to have both.

PR #738, opened 2018-07-03
4.x: Fix long sequence ObserveOn StackOverflowException upon Dispose

akarnokd:
This PR fixes the `StackOverflowException` when a long running `ObserveOn` (> 100.000 items) gets disposed due to the `ScheduledItem`s linked together when using the recursive scheduler.

The fix is to track the next recursive task's disposable directly and not allow the chaining to happen. For reference, this is the same effect that had to be considered in the improved `Range` and `ToObservable` implementations.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs
@@ -536,7 +536,10 @@ namespace System.Reactive
 
             if (Interlocked.Decrement(ref _wip) != 0)
             {
-                return recursiveScheduler.Schedule(this, DRAIN_SHORT_RUNNING);
+                // Don't return the disposable of Schedule() because that may chain together
+                // a long string of ScheduledItems causing StackOverflowException upon Dispose()
+                var d = recursiveScheduler.Schedule(this, DRAIN_SHORT_RUNNING);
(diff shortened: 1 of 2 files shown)


Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs
```diff
@@ -536,7 +536,10 @@ private IDisposable DrainShortRunning(IScheduler recursiveScheduler)
 
             if (Interlocked.Decrement(ref _wip) != 0)
             {
-                return recursiveScheduler.Schedule(this, DRAIN_SHORT_RUNNING);
+                // Don't return the disposable of Schedule() because that may chain together
+                // a long string of ScheduledItems causing StackOverflowException upon Dispose()
+                var d = recursiveScheduler.Schedule(this, DRAIN_SHORT_RUNNING);
+                Disposable.TrySetMultiple(ref _task, d);
             }
             return Disposable.Empty;
```
danielcweber: Shouldn't it at least return `this` here?

akarnokd: That would recurse upon calling `Dispose()` for no practical purpose. The `public void Dispose()` takes care of the cleanup when the downstream disposes, including the outstanding tasks.

PR #740, opened 2018-07-03
Remove redundancies in symbol declarations

danielcweber:


Rx.NET/Source/src/System.Reactive/Internal/ScheduledObserver.cs
```diff
@@ -637,7 +632,7 @@ private void DrainLongRunning()
                     // delayError: true - because of 
                     //      ObserveOn_LongRunning_HoldUpDuringDispatchAndFail
                     // expects something that almost looks like full delayError
-                    if (DrainStep(q, downstream, true))
+                    if (DrainStep(downstream, true))
```
akarnokd: Please don't remove this argument. As the comment states above the atomics inside DrainStep forces the re-read of this queue field instead of keeping it in a register/private stack.

akarnokd: http://igoro.com/archive/volatile-keyword-in-c-memory-model-explained/

"You may find it surprising that a volatile read refreshes the entire cache, not just the  read value. Similarly, a volatile write (i.e., every C# write) flushes the entire cache, not just the written value. These semantics are sometimes referred to as “strong volatile semantics”."

danielcweber: It's not obvious. `_queue` is readonly, why does it have to be passed into `DrainStep`?

akarnokd: http://afana.me/archive/2015/07/10/memory-barriers-in-dot-net.aspx/

"Load memory barrier (or read memory barrier)

A load memory barrier ensures that no LOAD operation can move across the barrier. All the LOAD operations that appear before the barrier will appear to happen before all the LOAD operations that appear after the memory barrier. "

So "LOAD _queue" has to happen after any load or full barriers, and `_queue.TryDequeue` has these types of barriers. So practically dequeueing from it "invalidates" the reference to it and has to be re-read upon next access from memory, even though the reference will be the same.

akarnokd: > unused argument being passed

It is supposed to be used, but apparently there is a mistake in the `DrainStep` body:

```cs
var empty = !_queue.TryDequeue(out var v);
```

should be

```cs
var empty = !q.TryDequeue(out var v);
```

danielcweber: I'll remove this, but it should be commented really well. Somebody will remove it eventually, because this is far from obvious. It's also not good code because it's not intention revealing at all, it misguides sophisticated tools as ReSharper and it's easily broken (by afforementioned reviews).

Please try to work around this or comment it really well.

akarnokd: If you correct `DrainStep` as I suggested, the argument won't show up as unused.

danielcweber: Ok, that does make sense.

danielcweber: It's restored for now, fix will go into a separate PR.

PR #759, opened 2018-07-04
Fix a few (possibly) missed out synchronizations.

danielcweber:
ReSharper suggested there are inconsistencies with some fields being locked sometimes and not other times.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
@@ -307,7 +307,11 @@ namespace System.Reactive.Linq.ObservableImpl
                 private void CreateWindow()
                 {
                     var s = new List<TSource>();
-                    _q.Enqueue(s);
+
+                    lock (_gate)
+                    {
+                        _q.Enqueue(s);
+                    }
                 }
 
                 private void CreateTimer()
@@ -444,7 +448,10 @@ namespace System.Reactive.Linq.ObservableImpl
 
                 public void Run(TimeHopping parent)
                 {
-                    _list = new List<TSource>();
+                    lock (_gate)
+                    {
+                        _list = new List<TSource>();
+                    }
 
                     Disposable.SetSingle(ref _periodicDisposable, parent._scheduler.SchedulePeriodic(this, parent._timeSpan, @this => @this.Tick()));
                     Run(parent._source);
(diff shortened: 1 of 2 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
```diff
@@ -307,7 +307,11 @@ protected override void Dispose(bool disposing)
                 private void CreateWindow()
                 {
                     var s = new List<TSource>();
-                    _q.Enqueue(s);
+
+                    lock (_gate)
```
akarnokd: No need for this. The `CreateWindow` is called from `Run` at which point there is no chance for concurrent access. Later on, the `CreateWindow` is called from inside a lock so no need to add the lock again.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Buffer.cs
```diff
@@ -444,7 +448,10 @@ public _(IObserver<IList<TSource>> observer)
 
                 public void Run(TimeHopping parent)
                 {
-                    _list = new List<TSource>();
+                    lock (_gate)
```
akarnokd: No need for this either. This part happens before `SchedulePeriodic` and `Run` and thus it will be correctly visible to those. `SetSingle` and `SchedulePeriodic` are guaranteed to be full memory barriers.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Window.cs
```diff
@@ -169,7 +169,12 @@ public void Run(TimeSliding parent)
                 private void CreateWindow()
                 {
                     var s = new Subject<TSource>();
-                    _q.Enqueue(s);
+
+                    lock (_gate)
```
akarnokd: Not necessary, same as with `Buffer`.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Window.cs
```diff
@@ -328,8 +333,14 @@ private void Tick()
 
                 private void CreateWindow()
                 {
-                    _subject = new Subject<TSource>();
-                    ForwardOnNext(new WindowObservable<TSource>(_subject, _refCountDisposable));
+                    var newSubject = new Subject<TSource>();
+
+                    lock (_gate)
```
akarnokd: Not needed. Same as with the equivalent `Buffer` operator.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Window.cs
```diff
@@ -328,8 +333,14 @@ private void Tick()
...
+                    {
+                        _subject = newSubject;
+                    }
+
+                    ForwardOnNext(new WindowObservable<TSource>(newSubject, _refCountDisposable));
```
akarnokd: If you see `ForwardOnNext` it means the enclosing block has to be already guaranteed to not run concurrently with each other, thus the lock above is unnecessary.

PR #767, opened 2018-07-05
Use StableCompositeDisposable.Create for two disposables.

danielcweber:


Rx.NET/Source/src/System.Reactive/Platforms/UWP/Concurrency/CoreDispatcherScheduler.cs
```diff
@@ -123,7 +123,7 @@ public override IDisposable Schedule<TState>(TState state, Func<IScheduler, TSta
                 }
             });
 
-            return new CompositeDisposable(
+            return StableCompositeDisposable.Create(
```
akarnokd: `CreateTrusted`?

akarnokd: Right, two arguments only, my mistake.

PR #771, opened 2018-07-06
Sink.Dispose should only call Dispose(true) once.

### LINKED ISSUE #768: Multiple subscriptions if using .Publish().RefCount() in 4.0.0
Hello guys I have a very strange behavior with using `.Publish().RefCount()` from `System.Reactive v4.0.0`. Please have a look at the following example program:
```
static class Program
{
    static void Main(string[] args)
    {
        var o1 = Observable.Create<string>(observer =>
        {
            Console.WriteLine("Observable.Create<string>()");

            observer.OnNext("A+");
            observer.OnNext("C+");
            observer.OnNext("B+");
            observer.OnNext("B-");
            observer.OnNext("A-");
            observer.OnCompleted();

            return Disposable.Empty;
        });

        var o2 = o1.Publish().RefCount(); // Variant A): Bad
        //var o2 = o1.Publish(); // Variant B): OK

        var s1 = o2.Subscribe(text => Console.WriteLine($"First:  {text}"));
        //var c1 = o2.Connect(); // Variant B): OK
        var s2 = o2.Subscribe(text => Console.WriteLine($"Second: {text}"));

        

        Console.WriteLine("Press any key to exit.");
        Console.ReadKey();

        s1.Dispose();
        s2.Dispose();
        //c1.Dispose(); // Variant B): OK
    }
}
```

Output Variant A):
```
Observable.Create<string>()
First:  A+
First:  C+
First:  B+
First:  B-
First:  A-
Observable.Create<string>()
Press any key to exit.
```

Output Variant B):
```
Observable.Create<string>()
First:  A+
First:  C+
First:  B+
First:  B-
First:  A-
Press any key to exit.
```
Based on the documentation I would expect that `Observable.Create<string>()` is only printed once on the console. Because `s1` subscribed to `o2`; so `s2` should additionally not also subscribe to the underlying observable.
Do I miss something?

However if I change the usage of `.Publish().RefCount()` into `.Publish()` and then `.Connect()`, `Observable.Create<string>()` will be only printed once like expected. Is this a bug in `.RefCount()`?

danielcweber:
This is part of a a two part fix for #768. Dispose.Create has been inlined in 6ecc899a4ba35db5834f55dcfde019e19b431940, which, for RefCount, will lead to garbage if Sink.Dispose is called multiple times. This PR makes Sink.Dispose a nop from the second call on.

Note: This is not enough to fix #768, although the expectation of the OP about the behaviour is incorrect, there is another bug that will be fixed in a following PR.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information. 
 
 using System.Reactive.Disposables;
+using System.Threading;
 
 namespace System.Reactive
 {
@@ -25,12 +26,16 @@ namespace System.Reactive
 
         public void Dispose()
         {
-            Dispose(true);
+            if (_observer != NopObserver<TTarget>.Instance && Interlocked.Exchange(ref _observer, NopObserver<TTarget>.Instance) != NopObserver<TTarget>.Instance)
+                Dispose(true);
         }
 
         protected virtual void Dispose(bool disposing)
         {
-            _observer = NopObserver<TTarget>.Instance;
+            //Calling base.Dispose(true) is not a proper disposal, so we can omit the assignment here.
+            //Sink is internal so this can pretty much be enforced.
+            //_observer = NopObserver<TTarget>.Instance;
+
             Disposable.TryDispose(ref _upstream);
         }
 
--- a/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
@@ -26,7 +26,7 @@ namespace System.Reactive
 
         public void Dispose()
         {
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Internal/Sink.cs
```diff
@@ -25,12 +26,16 @@ protected Sink(IObserver<TTarget> observer)
 
         public void Dispose()
         {
-            Dispose(true);
+            if (_observer != NopObserver<TTarget>.Instance && Interlocked.Exchange(ref _observer, NopObserver<TTarget>.Instance) != NopObserver<TTarget>.Instance)
```
danielcweber: @akarnokd Check first and then exchange is the more efficient way to go I guess?

akarnokd: It has slightly more overhead when the method is called only once (mainly due to targeting the same address and causing a data dependency). Its benefits show when 3+ calls are typical. In our case, we'd expect typically 1 call, sometimes two so the `Interlocked.Exchange` should suffice and should be practically equivalent to the original volatile field store (they both are usually implemented as `LOCK XCHG` or `MOV + LOCK XADD stack[0], 0`).

PR #772, opened 2018-07-06
Revert changes to RefCount.

### LINKED ISSUE #768: Multiple subscriptions if using .Publish().RefCount() in 4.0.0

danielcweber:
Revert changes to RefCount from 6ecc899a4ba35db5834f55dcfde019e19b431940.
We might take another attempt to optimize it.

Fixes #768.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
@@ -14,6 +14,7 @@ namespace System.Reactive.Linq.ObservableImpl
         internal sealed class Eager : Producer<TSource, Eager._>
         {
             private readonly IConnectableObservable<TSource> _source;
+
             private readonly object _gate;
             private int _count;
             private IDisposable _connectableSubscription;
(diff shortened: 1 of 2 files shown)


akarnokd:
Maybe it should include an unit test based on #768 so future optimizations don't run into the same problem.

PR #773, opened 2018-07-06
Attempt to restore the original (buggy) RefCount optimization with additional logic.

danielcweber:


Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
```diff
@@ -23,45 +22,54 @@ public Eager(IConnectableObservable<TSource> source)
...
                     : base(observer)
                 {
+                    _parent = parent;
                 }
 
-                public void Run(Eager parent)
+                public void Run()
                 {
-                    var subscription = parent._source.SubscribeSafe(this);
+                    Run(_parent._source);
```
danielcweber: We would have very obvious correctness if Run... would go into the lock.

In this solution, _count might become negative again but I think it's still correct, if 2 subscribers would interleave, one of them would still succeed in connecting.

akarnokd:
I'm currently running a multi-hour benchmark but after that, I can have a look at the problem.

PR #774, opened 2018-07-06
4.x: Improve RefCount() internals

akarnokd:
This PR adds a dedicated connection state object to avoid reconnection and disconnection interference as well as some comments about what each part of the algorithm does.

(Note: it does not add the unit test from #773 as it is unclear to me that unit test is supposed to ensure, an old or new behavior?).

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
@@ -89,25 +89,29 @@ namespace System.Reactive.Linq.ObservableImpl
                 protected override void Dispose(bool disposing)
                 {
                     base.Dispose(disposing);
-                    // get and forget the saved connection
-                    var targetConnection = _targetConnection;
-                    _targetConnection = null;
+                    if (disposing)
+                    {
+                        // get and forget the saved connection
+                        var targetConnection = _targetConnection;
+                        _targetConnection = null;
 
-                    lock (_parent._gate) {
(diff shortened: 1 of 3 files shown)


danielcweber:
The purpose of that unit test was to ensure that, should Run(...) dispose the Sink synchronously, it will be discovered and the call to Connect will be avoided.

akarnokd:
Only in the synchronous case? What about an async completing source? Should we detect any termination and prevent further connections?

danielcweber:
I guess 99% of the time `RefCount` will deal with some `Subject`, and once that is completed, the notification will come in synchronously. I don't think there is enough reason to cover more complex scenarios.

Rx.NET/Source/src/System.Reactive/Linq/Observable/RefCount.cs
```diff
@@ -16,54 +16,114 @@ internal sealed class Eager : Producer<TSource, Eager._>
...
+                    Run(_parent._source);
+                    // then connect the source if necessary
+                    if (doConnect)
                     {
-                        subscription.Dispose();
+                        // this makes sure if the connection ends synchronously
+                        // only the currently known connection is affected
+                        // and a connection from a concurrent reconnection won't
+                        // interfere
+                        Disposable.SetSingle(ref conn._disposable, _parent._source.Connect());
```
danielcweber: Checked it with the unit test from #773 and it's red. It should be enough to guard this by `if (!Disposable.GetIsDisposed(ref conn._disposable))` to avoid connecting and immediately unconnecting.

PR #779, opened 2018-07-08
Experimental: Subjects with ImmutableList

danielcweber:


akarnokd:
#### BehaviorSubject

Mostly within noise range, almost no memory footprint improvement over 3.1.1

image

#### Subject

Massive performance drop.

image

#### ReplaySubject

Within noise range.

image

PR #780, opened 2018-07-08
Experimental: Subjects with ImmutableDictionary

danielcweber:


akarnokd:
#### BehaviorSubject

Slightly slower than 3.1.1

image

#### ReplaySubject

Within noise range (about +/- 3%)

image

#### Subject

Massive performance drop and significantly increased memory usage:

image

PR #781, opened 2018-07-08
Experimental: Subjects with ImmutableSortedDictionary

danielcweber:


akarnokd:
#### BehaviorSubject

image

#### ReplaySubject

image

#### Subject

Massive performance drop and significantly more memory usage

image

danielcweber:
My interpretation of these results:

First, only `Subject` was affected by the changes, any results on `AsyncSubject` or `BehaviorSubject` are noise.

If there ever was some performance gain expected from Immutable Collections, it would be during subscription, when repeated re-allocation of an array/copying (and thus, garbage) may be avoided. The benefits might actually show with a lot of upfront subscriptions and/or randomized subscription/unsubscription.

The performance penalty we see is mostly due to enumeration, which can't beat the array since the immutable collection traverses a data structure while the array is, well, aligned in memory and thus very cache friendly.

The benefits might show eventually for massive amounts of live subscriptions, however, it's probably not the case for 99% of `Subject` consumers. The penalty is way to high to go in that direction I guess.

PR #793, opened 2018-07-21
Remove tail recursion from the PriorityQueue

quinmars:
The JIT does sometimes optimize tail recursion under some circumstances, but as far as I know, not always. Hence I would not rely on it and do the conversion by hand.


danielcweber:
Good idea. I'll try to fix the build failure first.

PR #795, opened 2018-07-21
Hide List<E> with .Skip(0)

quinmars:
Instead of yielding the elements in a foreach loop, zero elements are skipped. Technically the result is the same with the benefit that the LINQ implementation can internally generate a type that keeps track of count of the enumerable. Consecutive operators can benefit from that knowledge.


danielcweber:
Please add a unit test asserting that `Skip(0)` won't be optimized into a nop in the future. I like how the name `Hide` is intention revealing, what is it's disadvantage?

quinmars:
I've added a unit test to check that the internal list gets really hidden.

Issue #794, opened 2018-07-21
CI fails to fetch .NET core sdk 2.1.400-preview-009063


quinmars:
See https://dotnetfoundation.visualstudio.com/Rx.NET/_build/results?buildId=3261&view=logs from the PR #793.

quinmars:
Same with #795.

danielcweber:
This seems to be related to https://github.com/Microsoft/vsts-tasks/issues/7754.

quinmars:
Seems to be fixed by  7d5254e.

PR #796, opened 2018-07-24
Use a stable .net core sdk instead of the preview.

danielcweber:
See https://github.com/dotnet/reactive/issues/794.

### DIFF
--- a/.vsts.ix-shared.yml
+++ b/.vsts.ix-shared.yml
@@ -4,7 +4,7 @@ parameters:
 steps:
 - task: DotNetCoreInstaller@0
   inputs:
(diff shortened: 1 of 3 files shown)


danielcweber:
@onovotny Is there a specific reason for using the 2.1.400-preview?

danielcweber:
Pretty bizarre. The latest stable would lead to the approval tests failing although they are run on the full framework and there is no change in the API. Moreover, the Reporter set on the approvals seems to be ignored.

Apparently 2.1.400-preview-009180 is the latest preview, however, CI will fail to download this as well.

danielcweber:
Same for 2.2.100-preview1-009232.

PR #806, opened 2018-08-22
Fix Zip(IEnumerable) NullReferenceException if a source completes immediately

akarnokd:
If a source to the `Zip` operator completes immediately before the `SubscribeSafe` returned, the associated entry in `_subscriptions` would be still null, thus the `Dispose()` call fails with a `NullReferenceException`.

The fix checks if the `_subscriptions` is still valid and uses the deferred dispose approach to clean up the indexed entry.


danielcweber:
Early termination aside, the fact that we wouldn't handle the disposal with TryDispose at that specific location was a bug in its own right, wasn't it?

akarnokd:
Yes, a concurrent completion would have potentially experience that null slot too.

PR #807, opened 2018-08-28
Emit a warning if the half-serializer ignores a value in ForwardOnNext.

danielcweber:
In case an upstream observable doesn't behave according to the protocol, the half-serializer will swallow values in ForwardOnNext. While the guidelines state that it is safe to assume that upstreams behave properly, this behaviour is different to <4.1.0 and it is fair to produce a warning.


danielcweber:
@onovotny I fear we got some hanging check here...

onovotny:
I think there's an issue with VSTS triggers...looking into it.

Issue #815, opened 2018-09-28
Behaviour change in Task<IDisposable> overload of Observable.Create in version 4.1?


daniel-smith:
When updating from version 4.0 to 4.1, I observed some odd behaviour, which I boiled down to the following:

```
var obs = Observable.Create<int>(async o =>
{
    var inner = Observable.Range(1, 3);

    return inner.Subscribe(x => o.OnNext(x));
});

var result = obs.Take(1).Wait();
```

In our actual code we are calling an async method prior to subscribing to the inner observable, but there's no need to do that in order to reproduce the issue.

In version 4.0, this works perfectly fine, and the value of result is 1.

In 4.1, an InvalidOperationException "Sequence contains no elements" is thrown. This only seems to occur if the inner stream does not complete, i.e. if I replace the `Observable.Range(1, 3)` to `Observable.Return(1)`, it works fine.

We fixed our code by converting the actual async call we were doing to an observable by using `Observable.FromAsync`, then doing a `SelectMany` over that (which is probably a better solution anyway). Surprised to see this behaviour change though. Could this be a bug?

dunkymole:
Just to add to this. The OP didnt include an await in the async lambda. I have observed that one not only needs an await, the await needs to complete asynchronously.

This wont work:

```
    var obs = Observable.Create<int>(async o =>
    {
        await Task.CompletedTask;
        
        var inner = Observable.Range(1, 3);
        
        return inner.Subscribe(x => o.OnNext(x));
    });
    
    var result = obs.Take(1).Wait();
```
This is fine:


``` var obs = Observable.Create<int>(async o =>
    {
        await Task.Factory.StartNew(() => {});
        
        var inner = Observable.Range(1, 3);
        
        return inner.Subscribe(x => o.OnNext(x));
    });
    
    var result = obs.Take(1).Wait();```

akarnokd:
The `Task` returned by the function is subscribed (!) to by `TaskDisposeCompletionObserver` which then completes the `Observer` of `Take` and thus no items are likely received from range.

Looks like there was a mistake in https://github.com/dotnet/reactive/pull/549/files#diff-88bb42c08f322e657242aaba642022feL66 where the rewrite added that `OnComplete` call. I'll post a PR shortly to fix that.

akarnokd:
Interestingly, different overloads behaved differently:

[...]

PR #816, opened 2018-09-28
4.x: Fix accidental behavior change with Task-based Create methods completing when the body ends

akarnokd:
Before the accidental change, `Observable.Create(Func<IObserver<T>,[...], Task<IDisposable>)` and `Observable.Create(Func<IObserver<T>,[...], Task<Action>)` did not complete when the task body ended, unlike `Observable.Create(Func<IObserver<T>,[...],Task)`. 

There was an oversight when these operators have been upgraded and the plain `Task` version's logic was copy-pasted to them. Note though that there were no unit tests verifying the expected behavior at the time.

Fixes: #815

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Creation.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.Creation.cs
@@ -155,7 +155,6 @@ namespace System.Reactive.Linq
 
                     public void OnCompleted()
                     {
-                        _observer.OnCompleted();
                     }
 
                     public void OnError(Exception error)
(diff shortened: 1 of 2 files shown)


akarnokd:
I'd say **4.1.1** as this is a patch-level change.

#806, #807 and #809 perhaps?

onovotny:
Okay -- my suggestion would be to create a PR to the rel/v4.1 branch that cherry-picks the changes you want in it. We can then release that. In the PR, it can bump the version.json to 4.1.1.

danielcweber:
#809 adds a attributes to publicly visible classes and should bump to 4.2.0 then AFAIK. If we definitely want that one in too, we could keep it simple and just release 4.2.0 from master.

onovotny:
I don't think adding attributes is considered a breaking change. Removing, perhaps.

danielcweber:
Ok. I can prepare the PR later.

PR #821, opened 2018-10-02
netstandard 2.0/UWP 10-support for System.Reactive.Experimental-facade

### LINKED ISSUE #812: Update System.Reactive.Experimental to target .NET Standard
[System.Reactive.Experimental (4.1.0 at time of reporting)](https://www.nuget.org/packages/System.Reactive.Experimental/4.1.0) targets .NET Framework 4.6:

https://github.com/dotnet/reactive/blob/1888ca8e4aced0f7168b5e3e23b40b92fe0a5ec2/Rx.NET/Source/facades/System.Reactive.Experimental/System.Reactive.Experimental.csproj#L3-L5

Any reason it cannot be updated to target .NET Standard 2.0 like System.Reactive?

danielcweber:
Fixes #812.


onovotny:
I'm not sure we need this since the old facades only had net46. It's not clear how any other tfm could need it?

danielcweber:
I don't really have an opinion on that. Someone raised an issue, I figured that there was indeed an asymmetry, it was easily done and it compiled ;-)

onovotny:
I'm going to close this for now until there's a reason for it.

atifaziz:
> until there's a reason for it.

@onovotny What kind of reasoning are you looking for? It's annoying to see the following warning during compilation:

```
warning NU1701: Package 'System.Reactive.Experimental 4.1.2' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project.
```

onovotny:
Why are you adding a reference to System.Reactive.Experimental to your .NET Core project? That library  only ever existed for .NET Framework, so there shouldn't be anything trying to use that façade on a different platform.

PR #828, opened 2018-10-08
4.x: Fix 3+ arg Zip not working with immediate sources

akarnokd:
This PR fixes the issue with the 3+ argument `Zip` operator not producing anything with an immediate `Observable.Return` source(s).

The problem was that the per-source `OnCompleted` called the `Dispose` method which then called the overridden `Dispose(bool)` method inside `ZipObservable` that cleared the queue, thus the operator could never make a full row. The fix is to call the original `Dispose(bool)` which only disposes the upstream but leaves the queue alone.

Fixes: #824

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
@@ -524,7 +524,7 @@ namespace System.Reactive.Linq.ObservableImpl
 
         public override void OnCompleted()
         {
-            Dispose();
+            base.Dispose(true);
 
             lock (_gate)
             {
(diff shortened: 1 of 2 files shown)


Rx.NET/Source/src/System.Reactive/Linq/Observable/Zip.cs
```diff
@@ -524,7 +524,7 @@ public override void OnError(Exception error)
 
         public override void OnCompleted()
         {
-            Dispose();
+            base.Dispose(true);
```
quinmars: A small comment why the base (and not the own) method is called could be helpful for future readers.

danielcweber:
@onovotny I cherry picked this bugfix for a v4.1.2 release, bumped the version, pushed it to rel/v4.1.2.

onovotny:
Okay, it's rolling out to NuGet now. BTW, for bugfix/patch releases, I think we should make those to the rel/v4.1 branch and keep branches for major/minor updates. I've pushed the changes into the rel/v4.1 branch.

danielcweber:
Alright!

PR #830, opened 2018-10-16
light up the NuGet packages on GitHub

shiftkey:
You might have noticed [this announcement](https://blog.github.com/2018-10-16-future-of-software/#java-and-net-support-for-security-vulnerability-alerts) about adding support for NuGet packages to GitHub's security vulnerability alerts.

I tested out Octokit.net today and noticed this, and got a bit sad:

Nevermind, it's easy to fix - if the `PackageProjectUrl` or `RepositoryUrl` attributes point to a GitHub repository URL (we're not following redirects, ain't nobody got time for that), these will light up differently:

This sadly won't be backported to the old package formats, but anyone using `System.Reactive` will get the benefit after a package update.

I'm not sure how attached to `fwlink`s you are these days, so `RepositoryUrl` might be the necessary workaroudn here if you want to keep them Just In Case™.


shiftkey:
@onovotny I'll check in with the team after Universe is done and see if they can help

PR #831, opened 2018-10-17
fix packageUrl property

olevett:
was stopping the build - https://dev.azure.com/dotnet/Rx.NET/_build/results?buildId=5813&view=logs


shiftkey:
Apologies all!

PR #832, opened 2018-10-17
Fix order of observer and resource disposal of the Using and Finally operator

### LINKED ISSUE #829: changed the order of the call Observable.Using in 4.1
`var d1 =
Observable.Using(() => System.Reactive.Disposables.Disposable.Create(() => Console.WriteLine("3")), _ =>
    Observable.Using(() => System.Reactive.Disposables.Disposable.Create(() => Console.WriteLine("2")), __ =>
        Observable.Using(() => System.Reactive.Disposables.Disposable.Create(() => Console.WriteLine("1")), ___ =>
            //Observable.Never<System.Reactive.Unit>()// 4.0 (1234) -> 4.1 (3214)
            Observable.Return(System.Reactive.Unit.Default)// 4.0 (1234) -> 4.1 (4321)
        )
    )
)
.Finally(() => Console.WriteLine("4"))
.Subscribe();

d1.Dispose();`

This code show "1234" in console  with **Observable.Never** and **Observable.Return** in **Using** in version 4.0 or earlier.
But in 4.1 or high it show "3214" with **Observable.Never** and "4321" with **Observable.Return** in **Using**.

quinmars:
This should fix #829. For the `Using` operator, I found a solution without adding extra allocations. I have no idea how that could be done for the `Finally` operator, hence one extra allocation.

### DIFF
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/FinallyTest.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/FinallyTest.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information. 
 
 using System;
+using System.Reactive;
 using System.Reactive.Linq;
 using Microsoft.Reactive.Testing;
 using Xunit;
@@ -142,5 +143,48 @@ namespace ReactiveTests.Tests
             );
         }
 
+        [Fact]
+        public void Finally_DisposeOrder_Empty()
+        {
+            var order = "";
+            Observable
+                .Empty<Unit>()
+                .Finally(() => order += "1")
+                .Finally(() => order += "2")
+                .Finally(() => order += "3")
+                .Subscribe();
+
+            Assert.Equal("123", order);
+        }
+
+        [Fact]
+        public void Finally_DisposeOrder_Return()
+        {
+            var order = "";
+            Observable
+                .Return(Unit.Default)
+                .Finally(() => order += "1")
+                .Finally(() => order += "2")
+                .Finally(() => order += "3")
+                .Subscribe();
+
+            Assert.Equal("123", order);
+        }
+
+        [Fact]
+        public void Finally_DisposeOrder_Never()
+        {
+            var order = "";
+            var d = Observable
+                .Never<Unit>()
+                .Finally(() => order += "1")
+                .Finally(() => order += "2")
+                .Finally(() => order += "3")
+                .Subscribe();
+
+            d.Dispose();
+
+            Assert.Equal("123", order);
+        }
     }
 }


Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
```diff
@@ -34,33 +34,37 @@ public _(IObserver<TSource> observer)
...
                 {
-                    SetUpstream(Observable.Throw<TSource>(exception).SubscribeSafe(this));
-
-                    return;
+                    source = Observable.Throw<TSource>(exception);
                 }
 
+                // It is important to set the disposable resource after
+                // Run(). In the synchronous case this would else dispose
+                // the the resource before the source subscription.
```
danielcweber: What do you mean by "synchronous" case? If `Run` will observe termination of the source, why is it important to wait for `Run` to return?

(I realize I haven't dealt with this for quite a while ;-))

Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
```diff
@@ -34,33 +34,37 @@ public _(IObserver<TSource> observer)
...
                 Run(source);
+                Disposable.SetSingle(ref _disposable, disposable);
             }
 
             protected override void Dispose(bool disposing)
             {
+                base.Dispose(disposing);
```
danielcweber: I might be missing it, but I can't see why the switch in `Dispose` alone won't already do.

Besides: Do we really guarantee a specific order of disposal? I don't think this is a breaking change.  You should not rely on the order though, I guess...

Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
quinmars: If the source is immediate terminated, `_.Dispose()` will be called before the upstream disposable is set. So the resource would (and currently will) be disposed in the sink's dispose method. Later when `SetUpstream` is called the source subscription will be disposed, due to the magic of "TryDispose/SetSingle". With this PR the `TryDispose/SetSingle` magic is used for both the source subscription and the resource -- in the correct order.

Rx.NET/Source/src/System.Reactive/Linq/Observable/Using.cs
quinmars: This was my first idea as well, but it only solved the second unit (with never terminating sequence). The switch does not help if the source sequence is terminated before the subscription is passed to `SetUpstream`.


If you take

```csharp
using (disposable1)
{
    using (disposable2)
    {
        ...
    }
}
```

`disposable2` is guaranteed to be disposed before `disposable1` under any circumstances. I would excpect that the same for

```csharp
Observable.Using(() => disposable1,
    _ => Observable.Using(() => disposable2,
        ...
));
```

and indeed that was the case in v4.0.

quinmars:
Ok, I see. I will add that this evening, thanks. I will do the same for the `Using` operator, because I think in the current PR version it could happen that the resource disposal is started while the subscription disposal is still running.

quinmars:
I changed the `Finally` implementation to not use an anonymous as @akarnokd suggested. So the current PR does not introduce any additional allocations.

Regarding my concerns with the `Using` operator, I play out some possible scenarious and I'm now very confident that the execution of the two `Dispose` methods will not overlap, hence I haven't changed that implementation.

danielcweber:
Could you please provide a test for Finally as well?

I think the changes are fine, I just don't know whether we want to give any guarantees wrt. to the order of disposals. The recent performance enhancements have obviously changed some of the orders, there might be more occasions throughout the codebase. This PR's changes might definitely be desirable, but they also add some extra complexity. If we decide to guarantee a specific order from now on, we might run into even more issues like this down the road, and we might block ourselves from evolving the codebase.

I could use some opinions here.

quinmars:
From a user perspective constant behavior is desirable, but I do understand your concerns about the lost flexibility and one can see with this PR that those guarantees lead indeed to extra complexity. I do not advocate to nail down the order of all disposals, with the `Using` and `Finally` operator, however, it is appropriate in my opinion.

The `Using` operator would be pretty useless if the dispoable resource could be disposed while it is still in used. Similar, a `Finally` callback should be invoked at the very end and not somewhere along the line.

PR #833, opened 2018-10-20
Ix 4.0: Add shared state/concurrency tests

quinmars:
The PR adds concurrency tests for System.Linq.Async (see https://github.com/dotnet/corefx/issues/32640). Currently there are only tests for `SequenceEqual` and `Zip`; more will follow.

### DIFF
--- a/Ix.NET/Source/System.Linq.Async.Tests/SharedState.cs
+++ b/Ix.NET/Source/System.Linq.Async.Tests/SharedState.cs
@@ -26,7 +26,7 @@ namespace Tests
         {
             using (Use())
             {
-                await Task.Yield();
+                await Task.Delay(TimeSpan.FromMilliseconds(10));
             }
         }
 
--- a/Ix.NET/Source/System.Linq.Async.Tests/SharedState.cs
+++ b/Ix.NET/Source/System.Linq.Async.Tests/SharedState.cs
@@ -10,12 +10,16 @@ namespace Tests
         private readonly object _locker = new object();
         private bool _inUse = false;
 
+        public int ConcurrentAccessCount { get; private set; }
+
         public IDisposable Use()
         {
             lock (_locker)
             {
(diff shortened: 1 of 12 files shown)


onovotny:
I think right now there's a back-and-fourth about what the default behavior could/should be. There may need to be an additional overload that allows specifying concurrency. 

In the end, I think it needs a discussion.

quinmars:
The two test cases test if there is any overlapping/non-sequential access to the shared state. This happens for example with the use of `Task.WhenAny`/`Task.WhenAll`. The `Zip` implemenation uses `Task.WhenAll` so it will await both tasks, but in parallel.

As I understand the discussion, the problem comes with shared state manipulating enumerators. For example on every MoveNextAsync() an element is popped from a (shared) stack. If you now zip such a sequence with itself the `Pop` method needs to be thread-safe or the calls to it needs to be serialized. With a `Zip` operator that does await the Tasks sequentially, the calls to `Pop` are already in sequence. I'm not sure, though, if I hit the nail with the example.

I'm also not sure if I really like shared state manipulating enumerators, as I learned that a query should be sideeffect free, but I can imagine they can come in handy.

akarnokd:
It should be similar to how Rx expects things: don't share state and use external synchronization if needed. Otherwise, you'd enforce single source interaction in operators, which at best adds more latency to operators like `Zip` (limiting utilization) or makes them practically impossible like `Timeout` with another async sequence.

quinmars:
I added some more concurrency tests, currently `Zip`, `Except` and `Intersect` fail. Basically all operators that use `Task.WhenAll`. So not really a suprise.

onovotny:
@bartdesmet thoughts?

quinmars:
@bartdesmet, good point. In my initial tries the shared state was private, so there was no other opportunity than to throw. But now it can also be done with a testable property.

PR #842, opened 2018-11-08
Ignore JetBrains Rider-created .idea directories

akarnokd:
Rider loves to add these directories and show up as changed whenever one evaluates/changes code with Rider. 

*(I personally use Rider for code coverage as it is much faster than VS+dotCover even though Rider uses the dotCover too).*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,3 +185,7 @@ nuget.exe
 
 # dotnet local cache
 .dotnet
+
+# JetBrains Rider adds these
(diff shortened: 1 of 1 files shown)


akarnokd:
I don't know, maybe.

danielcweber:
We could as well just use https://github.com/github/gitignore/blob/master/VisualStudio.gitignore, it's way bigger than our current file and includes the Rider stuff.

PR #843, opened 2018-11-08
Async.Ix: Some fixes and proposed changes to operators

akarnokd:
I did a more detailed run-through of the code and these are my proposed changes/fixes. See additional comments among the code changes below.

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Amb.cs
+++ b/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Amb.cs
@@ -1,0 +1,228 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information. 
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Tests
+{
+    public class Amb : AsyncEnumerableExTests
+    {
+        [Fact]
+        public void TwoArg_First_Wins_NonEmpty()
+        {
+            var en = AsyncEnumerableEx.Amb(
+                    AsyncEnumerable.Range(1, 5),
+                    AsyncEnumerable.Range(6, 5).Select(async v =>
+                    {
+                        if (v == 6)
+                        {
+                            await Task.Delay(200);
+                        }
+                        return v;
+                    })
+                ).GetAsyncEnumerator();
+
+            try
+            {
+                HasNext(en, 1);
+                HasNext(en, 2);
+                HasNext(en, 3);
+                HasNext(en, 4);
+                HasNext(en, 5);
+                NoNext(en);
+
+            }
+            finally
+            {
+                en.DisposeAsync().AsTask().Wait();
+            }
+        }
+
+        [Fact]
+        public void TwoArg_Second_Wins_NonEmpty()
+        {
+            var en = AsyncEnumerableEx.Amb(
+                    AsyncEnumerable.Range(1, 5).Select(async v =>
+                    {
+                        if (v == 1)
+                        {
+                            await Task.Delay(200);
+                        }
+                        return v;
+                    }),
+                    AsyncEnumerable.Range(6, 5)
+                ).GetAsyncEnumerator();
+
+            try
+            {
+                HasNext(en, 6);
+                HasNext(en, 7);
+                HasNext(en, 8);
+                HasNext(en, 9);
+                HasNext(en, 10);
+                NoNext(en);
+            }
+            finally
+            {
+                en.DisposeAsync().AsTask().Wait();
+            }
+        }
+
+        [Fact]
+        public void TwoArg_First_Wins_Empty()
+        {
+            var en = AsyncEnumerableEx.Amb(
+                    AsyncEnumerable.Empty<int>(),
+                    AsyncEnumerable.Range(6, 5).Select(async v =>
+                    {
+                        if (v == 6)
+                        {
+                            await Task.Delay(200);
+                        }
+                        return v;
+                    })
+                ).GetAsyncEnumerator();
+
+            try
+            {
+                NoNext(en);
+            }
+            finally
+            {
+                en.DisposeAsync().AsTask().Wait();
+            }
+        }
+
+        [Fact]
+        public void TwoArg_Second_Wins_Empty()
+        {
+            var en = AsyncEnumerableEx.Amb(
+                    AsyncEnumerable.Range(1, 5).Select(async v =>
+                    {
+                        if (v == 1)
+                        {
+                            await Task.Delay(200);
+                        }
+                        return v;
+                    }),
+                    AsyncEnumerable.Empty<int>()
+                ).GetAsyncEnumerator();
+
+            try
+            {
+                NoNext(en);
+            }
+            finally
+            {
+                en.DisposeAsync().AsTask().Wait();
+            }
+        }
+
+        [Fact]
+        public void TwoArg_Cancel()
+        {
+            var cts = new CancellationTokenSource();
+
+            var en = AsyncEnumerableEx.Amb(
+                    AsyncEnumerable.Range(1, 5).Select(async (v, t) =>
+                    {
+                        if (v == 1)
+                        {
+                            await Task.Delay(2000, t);
+                        }
+                        return v;
+                    }),
+                    AsyncEnumerable.Range(6, 5).Select(async (v, t) =>
+                    {
+                        if (v == 6)
+                        {
+                            await Task.Delay(2000, t);
+                        }
+                        return v;
+                    })
+                ).GetAsyncEnumerator(cts.Token);
+
+            try
+            {
+                en.MoveNextAsync();
+
+                Task.Delay(200).Wait();
+
+                cts.Cancel();
+
+                Task.Delay(200).Wait();
+            }
+            finally
+            {
+                Assert.True(en.DisposeAsync().AsTask().Wait(1000));
+            }
+        }
+    }
+
+    internal static class AsyncTestEx
+    {
+        internal static IAsyncEnumerable<TResult> Select<TSource, TResult>(this IAsyncEnumerable<TSource> source, Func<TSource, CancellationToken, Task<TResult>> mapper)
+        {
+            return new SelectTokenTask<TSource, TResult>(source, mapper);
+        }
+
+        private sealed class SelectTokenTask<TSource, TResult> : IAsyncEnumerable<TResult>
+        {
+            private readonly IAsyncEnumerable<TSource> _source;
+
+            private readonly Func<TSource, CancellationToken, Task<TResult>> _mapper;
+
+            public SelectTokenTask(IAsyncEnumerable<TSource> source, Func<TSource, CancellationToken, Task<TResult>> mapper)
+            {
+                _source = source;
+                _mapper = mapper;
+            }
+
+            public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+            {
+                return new SelectTokenTaskEnumerator(_source.GetAsyncEnumerator(cancellationToken), _mapper, cancellationToken);
+            }
+
+            private sealed class SelectTokenTaskEnumerator : IAsyncEnumerator<TResult>
+            {
+                private readonly IAsyncEnumerator<TSource> _source;
+
+                private readonly Func<TSource, CancellationToken, Task<TResult>> _mapper;
+
+                private readonly CancellationToken _token;
+
+                public SelectTokenTaskEnumerator(IAsyncEnumerator<TSource> source, Func<TSource, CancellationToken, Task<TResult>> mapper, CancellationToken token)
+                {
+                    _source = source;
+                    _mapper = mapper;
+                    _token = token;
+                }
+
+                public TResult Current { get; private set; }
+
+                public ValueTask DisposeAsync()
+                {
+                    return _source.DisposeAsync();
+                }
+
+                public async ValueTask<bool> MoveNextAsync()
+                {
+                    if (await _source.MoveNextAsync())
+                    {
+                        Current = await _mapper(_source.Current, _token);
+                        return true;
+                    }
+                    return false;
+                }
+            }
+        }
+    }
+
+}
--- a/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Using.cs
+++ b/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Using.cs
@@ -65,7 +65,7 @@ namespace Tests
         }
 
         [Fact]
-        public void Using3()
+        public async void Using3()
         {
             var ex = new Exception("Bang!");
             var i = 0;
@@ -82,7 +82,22 @@ namespace Tests
 
             Assert.Equal(0, i);
 
-            AssertThrows<Exception>(() => xs.GetAsyncEnumerator(), ex_ => ex_ == ex);
+            var enumerator = xs.GetAsyncEnumerator();
+
+            while (await enumerator.MoveNextAsync())
+            {
+                // ignore any items
+            }
+
+            try
+            {
+                await enumerator.DisposeAsync();
+                Assert.False(true, "Should have thrown!");
+            }
+            catch (Exception exc)
+            {
+                Assert.Same(exc, ex);
+            }
 
             Assert.Equal(1, d);
         }
--- a/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
@@ -36,10 +36,10 @@ namespace System.Linq
             {
                 enumerator.OnGetEnumerator(token);
             }
-            catch
+            catch (Exception ex)
             {
-                enumerator.DisposeAsync(); // REVIEW: fire-and-forget?
-                throw;
+                // GetAsyncEnumerator should not throw but rather return a failing enumerator
+                return new FailingEnumerator(enumerator, ex);
             }
 
             return enumerator;
@@ -98,6 +98,35 @@ namespace System.Linq
         protected virtual void OnGetEnumerator(CancellationToken cancellationToken)
         {
         }
+
+        /// <summary>
+        /// Disposes a source enumerator then fails with an exception in DisposeAsync.
+        /// </summary>
+        private sealed class FailingEnumerator : IAsyncEnumerator<TSource>
+        {
+            private readonly IAsyncEnumerator<TSource> _source;
+
+            private readonly Exception _error;
+
+            public FailingEnumerator(IAsyncEnumerator<TSource> source, Exception error)
+            {
+                _source = source;
+                _error = error;
+            }
+
+            public TSource Current => throw new InvalidOperationException("Enumerator is in an invalid state");
+
+            public async ValueTask DisposeAsync()
+            {
+                await _source.DisposeAsync();
+                throw _error;
+            }
+
+            public ValueTask<bool> MoveNextAsync()
+            {
+                return TaskExt.False;
+            }
+        }
     }
 
     internal enum AsyncIteratorState
@@ -107,4 +136,5 @@ namespace System.Linq
         Iterating = 2,
         Disposed = -1,
     }
+
 }
--- a/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
+++ b/Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information. 
 
 using System.Collections.Generic;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -118,8 +119,8 @@ namespace System.Linq
 
             public async ValueTask DisposeAsync()
             {
-                await _source.DisposeAsync();
-                throw _error;
+                await _source.DisposeAsync().ConfigureAwait(false);
+                ExceptionDispatchInfo.Capture(_error).Throw();
             }
 
             public ValueTask<bool> MoveNextAsync()
--- a/Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
+++ b/Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
@@ -18,7 +18,7 @@ namespace System.Linq
             if (second == null)
                 throw new ArgumentNullException(nameof(second));
 
-            return new AmbAsyncIterator<TSource>(first, second);
+            return new AmbEnumerable<TSource>(first, second);
         }
 
         public static IAsyncEnumerable<TSource> Amb<TSource>(params IAsyncEnumerable<TSource>[] sources)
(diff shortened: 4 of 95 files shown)


Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Amb.cs
```diff
@@ -0,0 +1,228 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Tests
+{
+    public class Amb : AsyncEnumerableExTests
```
akarnokd: There was no unit test for `Amb` at all.

Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Amb.cs
```diff
@@ -0,0 +1,228 @@
...
+            finally
+            {
+                Assert.True(en.DisposeAsync().AsTask().Wait(1000));
+            }
+        }
+    }
+
+    internal static class AsyncTestEx
+    {
+        internal static IAsyncEnumerable<TResult> Select<TSource, TResult>(this IAsyncEnumerable<TSource> source, Func<TSource, CancellationToken, Task<TResult>> mapper)
```
akarnokd: A `Select` overload exposing a cancellation token is missing from the standard API. I've added one here so I can test `Amb` propagating cancellation properly.

Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Using.cs
```diff
@@ -82,7 +82,22 @@ public void Using3()
 
             Assert.Equal(0, i);
 
-            AssertThrows<Exception>(() => xs.GetAsyncEnumerator(), ex_ => ex_ == ex);
+            var enumerator = xs.GetAsyncEnumerator();
```
akarnokd: DisposeAsync failures are somewhat peculiar as the chain may not follow through upon an intermediate crash.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -36,10 +36,10 @@ public IAsyncEnumerator<TSource> GetAsyncEnumerator(CancellationToken token)
             {
                 enumerator.OnGetEnumerator(token);
             }
-            catch
+            catch (Exception ex)
             {
-                enumerator.DisposeAsync(); // REVIEW: fire-and-forget?
-                throw;
+                // GetAsyncEnumerator should not throw but rather return a failing enumerator
```
akarnokd: Simply return a special IAsyncEnumerator that fails after the source `DisposeAsync` is called with the original exception.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
```diff
@@ -98,6 +98,35 @@ public async ValueTask<bool> MoveNextAsync()
...
+                _source = source;
+                _error = error;
+            }
+
+            public TSource Current => throw new InvalidOperationException("Enumerator is in an invalid state");
+
+            public async ValueTask DisposeAsync()
+            {
+                await _source.DisposeAsync();
+                throw _error;
```
akarnokd: C# `throw` is awkward as it rewrites the stacktrace upon throwing an existing instance. Probably that's why there is so many `AggregateException` wrapping is going around. Could be necessary here too.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
```diff
@@ -37,101 +37,176 @@ public static IAsyncEnumerable<TSource> Amb<TSource>(this IEnumerable<IAsyncEnum
             return new AmbAsyncIteratorN<TSource>(sources.ToArray());
         }
 
-        private sealed class AmbAsyncIterator<TSource> : AsyncIterator<TSource>
+        private sealed class AmbEnumerable<TSource> : IAsyncEnumerable<TSource>
```
akarnokd: This is a full rewrite because the cancellation has to be properly linked and they need to cancel each other.

Ix.NET/Source/System.Linq.Async/System/Linq/AsyncEnumerablePartition.cs
```diff
@@ -89,7 +89,7 @@ public async Task<int> GetCountAsync(bool onlyIfCheap, CancellationToken cancell
                 // At the same time, however, we are guaranteed that our max count can fit
                 // in an int because if that is true, then _minIndexInclusive must > 0.
 
-                var count = await SkipAndCountAsync((uint)_maxIndexInclusive + 1, en, cancellationToken).ConfigureAwait(false);
+                var count = await SkipAndCountAsync((uint)_maxIndexInclusive + 1, en).ConfigureAwait(false);
```
akarnokd: Cancellation would have no effect on this code path anyway.

Ix.NET/Source/System.Linq.Async/System/Linq/AsyncEnumerablePartition.cs
```diff
@@ -351,21 +351,21 @@ public async Task<List<TSource>> ToListAsync(CancellationToken cancellationToken
...
-        private static async Task<int> SkipAndCountAsync(int index, IAsyncEnumerator<TSource> en, CancellationToken cancellationToken)
+        private static async Task<int> SkipAndCountAsync(int index, IAsyncEnumerator<TSource> en)
         {
             Debug.Assert(index >= 0);
-            return (int)await SkipAndCountAsync((uint)index, en, cancellationToken).ConfigureAwait(false);
+            return (int)await SkipAndCountAsync((uint)index, en).ConfigureAwait(false);
         }
 
-        private static async Task<uint> SkipAndCountAsync(uint index, IAsyncEnumerator<TSource> en, CancellationToken cancellationToken)
+        private static async Task<uint> SkipAndCountAsync(uint index, IAsyncEnumerator<TSource> en)
```
akarnokd: Not sure if cancellation would have had effect here as `en` is already retrieved with a token thus the `MoveNextAsync` should unblock anyway.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Except.cs
```diff
@@ -124,14 +124,9 @@ protected override async ValueTask<bool> MoveNextCore(CancellationToken cancella
...
-            private async Task FillSetAsync()
+            private async Task FillSetAsync(CancellationToken cancellationToken)
             {
-                var array = await _second.ToArray().ConfigureAwait(false);
-
-                foreach (var t in array)
-                {
-                    _set.Add(t);
-                }
+                await _second.ForEachAsync(t => _set.Add(t), cancellationToken).ConfigureAwait(false);
```
akarnokd: Why create an array, then copy it into a set? Also the token here has purpose to stop the for-eaching.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Intersect.cs
```diff
@@ -124,13 +124,9 @@ protected override async ValueTask<bool> MoveNextCore(CancellationToken cancella
...
 
-            private async Task FillSet()
+            private async Task FillSet(CancellationToken cancellationToken)
             {
-                var array = await _second.ToArray().ConfigureAwait(false);
-                for (var i = 0; i < array.Length; i++)
-                {
-                    _set.Add(array[i]);
-                }
+                await _second.ForEachAsync(t => _set.Add(t), cancellationToken).ConfigureAwait(false);
```
akarnokd: Why create an array, then copy it into a set? Also the token here has purpose to stop the for-eaching.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
```diff
@@ -104,13 +104,28 @@ public override AsyncIterator<TResult> Clone()
...
             public override async ValueTask DisposeAsync()
             {
-                if (_enumerator != null)
+                var enumerator = _enumerator;
+                var error = default(Exception);
+                if (enumerator != null)
                 {
-                    await _enumerator.DisposeAsync().ConfigureAwait(false);
                     _enumerator = null;
+                    try
```
akarnokd: An overall problem I see with all such DisposeAsyncs, namely if the upstream's DisposeAsync crashes, the `base.DisposeAsync` doesn't run. Also depending on what `base` is, it may also crash we could lose the first exception.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
```diff
@@ -104,13 +104,28 @@ public override AsyncIterator<TResult> Clone()
...
+                        error = ex;
+                    }
                 }
 
                 await base.DisposeAsync().ConfigureAwait(false);
+                if (error != null)
+                {
+                    // 'throw error' doesn't preserve the stack trace
+                    // this, however, doesn't preserve the flat error structure.
+                    throw new AggregateException("The enumerator's DisposeAsync failed", error);
```
akarnokd: Again, wrapping so the stacktrace remains intact. Makes testing difficult as exceptions may end up being nested many levels...

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SelectMany.cs
```diff
@@ -156,12 +156,16 @@ protected override async ValueTask<bool> MoveNextCore(CancellationToken cancella
                         switch (_mode)
                         {
                             case State_Source:
+                                // Make sure we dispose the previous inner source
```
akarnokd: Not sure why the original only disposed the previous _resultEnumerator when the new one responded.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Skip.cs
```diff
@@ -15,14 +15,9 @@ public static IAsyncEnumerable<TSource> Skip<TSource>(this IAsyncEnumerable<TSou
...
-                // issues if collections are used as keys or otherwise must not be aliased.
-                if (source is AsyncIterator<TSource> || source is IAsyncPartition<TSource>)
-                {
-                    return source;
-                }
-
-                count = 0;
+                // If there is  nothing to skip, why would we ever create a partition.
+                // Let the next, non-trivial operator create the partition if needed.
+                return source;
```
akarnokd: If there is nothing to skip, why was it different from not having `Skip` in the chain to begin with?

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SkipLast.cs
```diff
@@ -54,15 +54,20 @@ public override AsyncIterator<TSource> Clone()
...
             public override async ValueTask DisposeAsync()
             {
-                if (_enumerator != null)
+                try
                 {
-                    await _enumerator.DisposeAsync().ConfigureAwait(false);
-                    _enumerator = null;
+                    if (_enumerator != null)
+                    {
+                        await _enumerator.DisposeAsync().ConfigureAwait(false);
```
akarnokd: Again, if DisposeAsync crashed, the `_enumerator` and `_queue` were not cleared.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Take.cs
```diff
@@ -15,7 +17,7 @@ public static IAsyncEnumerable<TSource> Take<TSource>(this IAsyncEnumerable<TSou
 
             if (count <= 0)
             {
-                return Empty<TSource>();
+                return new EmptyTake<TSource>(source);
```
akarnokd: This needs always some consideration. Do we want to have the "subscription side-effects" take place even if we are not really interested in any items?

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.cs
```diff
@@ -58,64 +59,187 @@ public static IAsyncEnumerable<TSource> ToAsyncEnumerable<TSource>(this IObserva
...
-                () =>
+            return new ObservableToAsyncEnumerable<TSource>(source);
+        }
+
+        /// <summary>
+        /// Wraps an IObservable and exposes it as an IAsyncEnumerable, buffering
+        /// all source items until they are requested via MoveNextAsync.
+        /// </summary>
+        /// <typeparam name="TSource">The element type of the source and result.</typeparam>
+        private sealed class ObservableToAsyncEnumerable<TSource> : IAsyncEnumerable<TSource>
```
akarnokd: Complete rewrite. Uses atomics instead of locks.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.cs
```diff
@@ -58,64 +59,187 @@ public static IAsyncEnumerable<TSource> ToAsyncEnumerable<TSource>(this IObserva
...
+                public void OnNext(TSource value)
+                {
+                    _queue?.Enqueue(value);
+
+                    Signal();
+                }
+
+                private void DisposeSource()
+                {
+                    var old = Interlocked.Exchange(ref _disposable, this);
```
akarnokd: Deferred disposing, the usual Rx stuff.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
```diff
@@ -27,47 +29,158 @@ public ToObservableObservable(IAsyncEnumerable<T> source)
...
             {
-                var ctd = new CancellationTokenDisposable();
-                var e = _source.GetAsyncEnumerator(ctd.Token);
+                var enumerator = new ToObservableEnumerator(observer);
+                enumerator.Source = _source.GetAsyncEnumerator(enumerator.Token);
+                enumerator.MoveNext();
+                return enumerator;
+            }
+
+            private sealed class ToObservableEnumerator : IDisposable
```
akarnokd: Complete rewrite. Now features all the dance around possible reentrant `MoveNextAsync`s and the need for an anytime dispose to wait out a `MoveNextAsync`.

Ix.NET/Source/System.Linq.Async/System/Threading/Tasks/TaskExt.cs
```diff
@@ -9,5 +9,16 @@ internal static class TaskExt
         public static readonly ValueTask<bool> True = new ValueTask<bool>(true);
         public static readonly ValueTask<bool> False = new ValueTask<bool>(false);
         public static readonly ValueTask CompletedTask = new ValueTask(Task.FromResult(true));
+
+        /// <summary>
+        ///  A constant TaskCompletionSource already completed with true.
+        /// </summary>
+        public static readonly TaskCompletionSource<bool> ResumeTrue;
```
akarnokd: This is part of the atomic resumption mechanism used in `ToAsyncEnumerable` and in many more coordinating operators for the future.

onovotny:
@bartdesmet any thoughts?

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
onovotny: What is the rationale here? Why can't a call to GetAsyncEnumerator throw?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
```diff
@@ -37,101 +37,176 @@ public static IAsyncEnumerable<TSource> Amb<TSource>(this IEnumerable<IAsyncEnum
...
         {
-            private readonly IAsyncEnumerable<TSource> _first;
-            private readonly IAsyncEnumerable<TSource> _second;
+            private readonly IAsyncEnumerable<TSource> _source1;
 
-            private IAsyncEnumerator<TSource> _enumerator;
+            private readonly IAsyncEnumerable<TSource> _source2;
 
-            public AmbAsyncIterator(IAsyncEnumerable<TSource> first, IAsyncEnumerable<TSource> second)
+            public AmbEnumerable(IAsyncEnumerable<TSource> source1, IAsyncEnumerable<TSource> source2)
```
onovotny: Nit, but should the parameter names still be `first`, and `second`?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
```diff
@@ -37,101 +37,176 @@ public static IAsyncEnumerable<TSource> Amb<TSource>(this IEnumerable<IAsyncEnum
...
-                        // 1. Concurrent DisposeAsync while a MoveNextAsync is in flight.
-                        // 2. The winner elected by Amb is blocked to yield results until the loser unblocks.
-                        //
-                        // The approach below has one drawback, namely that exceptions raised by loser are dropped on the floor.
-                        //
+                        _enumerator1 = _source1.GetAsyncEnumerator(_token1.Token);
+                        _enumerator2 = _source2.GetAsyncEnumerator(_token2.Token);
+
+#pragma warning disable CS4014 // ContinueWith used
+                        _enumerator1.MoveNextAsync()
```
onovotny: Can an await with a local function be used instead of a ContinueWith?

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
akarnokd: My undestanding is that `GetAsyncEnumerator` on an arbitrary `IAsyncEnumerable` should be trusted, otherwise every such call should be defended with try-catches and also make sure the resulting cleanup does everything correctly and at the right time. Besides, the main point here is to avoid "fire-and-forget".

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
akarnokd: Can be renamed, but this is internal stuff so doesn't matter.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
onovotny: I'm not sure if that's a fair assumption, but it's a good question @stephentoub, any thoughts on what the API contract is for `GetAsyncEnumerator`?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
akarnokd: The problem with `await` is that only one thing can be awaited in a logical flow, so which one to await? The first or the second source? By awaiting one before the other, we give a guaranteed pereference to one of them thus makes `Amb` obsolete. Also `await Task.WhenAny` is no good either because if any of the source fails, await crashes and without an explicit winner, the operator can break.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Take.cs
quinmars: I think so, else `.Finally(...).Take(0)` will not invoke the finally action.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Skip.cs
quinmars: `Skip(0)` hides the underlying implementation. So if you return `myAsyncList.Skip(calculatedValue)`, you would leak the original collection, if the `calculatedValue` happens to be zero. That is also how `Enumerable.Skip()` works.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
onovotny: @davidfowl what would you suggest here?

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Skip.cs
onovotny: Seems like there should be a unit test to cover this behavior (leakage of the original collection)?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
stephentoub: > Also await Task.WhenAny is no good either because if any of the source fails, await crashes and without an explicit winner, the operator can break.

What do you mean by `await crashes? `Task.WhenAny` will return a `Task<Task<bool>>` here, so awaiting it will give you back the task that completed, at which point you can choose to do with it whatever you want.

akarnokd: Right, I overlooked the definition of `Task.WhenAny`.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Skip.cs
akarnokd: I kept running the unit tests locally and nothing failed, otherwise I wouldn't have done this change.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
akarnokd: Updated to use `Task.WhenAny`, but I can't see it scale to the N-ary variant as the `DisposeAsync()` has to wait for all non-winners.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
bartdesmet: Could `ExceptionDispatchInfo` be of use to retain the stack trace in this case?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
```diff
@@ -37,101 +37,107 @@ public static IAsyncEnumerable<TSource> Amb<TSource>(this IEnumerable<IAsyncEnum
...
                 }
 
-                await base.DisposeAsync().ConfigureAwait(false);
-            }
+                public async ValueTask DisposeAsync()
+                {
+                    _cancelBoth.Dispose();
+                    if (_winner != null)
+                    {
+                        await _winner.DisposeAsync();
```
bartdesmet: We should be using `ConfigureAwait(false)` consistently in this library.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Except.cs
bartdesmet: Not sure about this history of this part of the code (@onovotny may be able to chime in). One thing I could think of is that an exception during iteration won't keep the set half-filled. This said, `_setFilled` should guard against it getting exposed.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
akarnokd: I didn't know about that, but [looks like](http://derpturkey.com/preserving-a-stacktrace-when-rethrowing-exception-with-exceptiondispatchinfo/) it is there for that purpose.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
bartdesmet: Yeah, this needs addressing, so it becomes equivalent to what the compiler will generate for:

```csharp
async IAsyncEnumerable<R> Select<T, R>(this IAsyncEnumerable<T> source, Func<T, Task<R>> selector)
{
    await foreach (var x in source)
      yield return await selector(x);
}
```

In this case, `base.DisposeAsync()` deals with the state machine the compiler would otherwise emit. All paths would properly transition to the state machine to the appropriate state.

What we're dealing with here is really nested `await using` statements in disguise. Behavior should guarantee inner-to-outer cleanup, where it's totally possible for the outer cleanup to supersede the exception of an inner cleanup operation.

```csharp
IEnumerable<int> X()
{
  using (new D(() => { Console.WriteLine("1"); throw new Exception("Oops 1"); }))
  {
    using (new D(() => { Console.WriteLine("2"); throw new Exception("Oops 2"); }))
    {
      yield return 42;
      yield return 43;
    }
  }
}
```

where `D` is an anonymous disposable. Behavior should be:

```csharp
var x = X();
var e = x.GetEnumerator();
e.Dispose();
// no cleanup; resources never acquired
```

and

```csharp
var x = X();
var e = x.GetEnumerator();
e.MoveNext();
e.Dispose();
// prints 2, 1, and throws outer "Oops 1"
```

and

```csharp
var x = X();
var e = x.GetEnumerator();
e.MoveNext();
e.MoveNext();
e.MoveNext();
// prints 2, 1, and throws outer "Oops 1"
```

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
bartdesmet: `ExceptionDispatchInfo.Capture(error).Throw()` will be the equivalent of a "rethrow" and is what async methods also use to emit code that retains the stack trace. See https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Lowering/AsyncRewriter/AsyncExceptionHandlerRewriter.cs#L426

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Except.cs
akarnokd: Wouldn't it fail the Task.WaitAll before that and `DisposeAsync` clear the set anyway?

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SelectMany.cs
bartdesmet: Seems we fell through to calling `DisposeAsync`, which would take care of that. The change seems fine too, and setting the field to `null` is defensive.

One thing I remarked is that the dispose order in `DisposeAsync` doesn't seem right. We have two nested `await foreach` statements, so inner should always dispose prior to outer. It should really be the equivalent of:

```csharp
await using (base)
  await using (_sourceEnumerator)
    await using (_resultEnumerator)
      ;
```

So all disposes are guaranteed and run in inner-to-outer order.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Take.cs
bartdesmet: Let's be consistent with the sync version of these operators. See https://github.com/dotnet/corefx/blob/143df51926f2ad397fef9c9ca7ede88e2721e801/src/System.Linq/src/System/Linq/Take.cs#L20

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
akarnokd: Got it, thanks.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Select.cs
akarnokd: I would have aggregated the failures, but we can probably do some (nested) try-finallys:

```cs
try {
    if (_enumerator != null) {
        await _enumerator.DisposeAsync().ConfigureAwait(false);
    }
} finally {
    _enumerator = null;
    await base.DisposeAsync().ConfigureAwait(false);
}
```

There are a lot of places so I'd rather do them in a separate PR.

Ix.NET/Source/System.Interactive.Async/AsyncIterator.cs
akarnokd: Done.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Amb.cs
akarnokd: Done.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/Take.cs
akarnokd: Done.

akarnokd:
There were a lot of recent commits conflicting with this one. I'll have to do a re-review anyway so I'm closing this.

PR #845, opened 2018-11-09
AsyncIx: More reviews/fixes/proposed changes

akarnokd:
Additional reviews, fixes and proposals remaining from #843. See comments below.


Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Finally.cs
```diff
@@ -114,7 +114,46 @@ public async Task Finally7()
             var xs = new[] { 1, 2 }.ToAsyncEnumerable().Finally(() => { i++; });
 
             await SequenceIdentity(xs);
-            Assert.Equal(2, i);
+            Assert.Equal(4, i);
```
akarnokd: Proposal: `Finally` to run when the `MoveNextAsync` is not called at all but the async sequence gets disposed.

SequenceIdentity first disposes an enumerator, then runs the comparison again, consuming the same asnync enumerable 4 times total.

Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Finally.cs
```diff
@@ -114,7 +114,46 @@ public async Task Finally7()
...
+        }
+
+        [Fact]
+        public async Task Finally8()
+        {
+            var i = 0;
+            var en = AsyncEnumerable.Range(1, 5).Finally(() => i++).GetAsyncEnumerator();
+
+            await en.DisposeAsync();
```
akarnokd: This tests if disposing without moving still executes the finally action.

Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Never.cs
```diff
@@ -25,5 +26,38 @@ public async Task Never1()
         private void Nop(object o)
         {
         }
+
+        [Fact]
+        public void CancelToken_Unblocks()
```
akarnokd: Now with `CancellationToken` support, we can unblock a never if it waits on an infinite delay.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Never.cs
```diff
@@ -11,14 +13,43 @@ public static partial class AsyncEnumerableEx
...
         {
-            //
-            // REVIEW: The C# 8.0 proposed interfaces don't allow for cancellation, so this "Never" is
-            //         as never as never can be; it can't be interrupted *at all*, similar to the sync
-            //         variant in Ix. Passing a *hot* CancellationToken to the Never operator doesn't
-            //         seem correct either, given that we return a *cold* sequence.
-            //
-
-            return AsyncEnumerable.CreateEnumerable(() => AsyncEnumerable.CreateEnumerator<TValue>(() => TaskExt.Never, current: null, dispose: null));
+            return NeverAsyncEnumerable<TValue>.Instance;
```
akarnokd: Never could wait on an infinite Task.Delay that can be cancelled.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Return.cs
```diff
@@ -3,14 +3,61 @@
...
     public static partial class AsyncEnumerableEx
     {
         public static IAsyncEnumerable<TValue> Return<TValue>(TValue value)
         {
-            return new[] { value }.ToAsyncEnumerable();
+            return new ReturnEnumerable<TValue>(value);
+        }
+
+        // FIXME: AsyncListPartition is internal to the project System.Linq.Async
+        // project, not sure how to expose it here
```
akarnokd: What if the projects could be merged, or the optimization interfaces and types exposed as public? If `System.Linq.Async` becomes a core library feature, how would it expose such optimization interfaces?

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Return.cs
```diff
@@ -3,14 +3,61 @@
...
+
+                public ValueTask DisposeAsync()
+                {
+                    Current = default;
+                    return TaskExt.CompletedTask;
+                }
+
+                public ValueTask<bool> MoveNextAsync()
+                {
+                    if (_once)
```
akarnokd: Return should benefit it being synchronously available.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Timeout.cs
```diff
@@ -73,10 +81,15 @@ protected override async ValueTask<bool> MoveNextCore(CancellationToken cancella
...
                                 var delay = Task.Delay(_timeout, delayCts.Token);
 
-                                var winner = await Task.WhenAny(moveNext.AsTask(), delay).ConfigureAwait(false);
+                                var next = moveNext.AsTask();
+
+                                var winner = await Task.WhenAny(next, delay).ConfigureAwait(false);
 
                                 if (winner == delay)
                                 {
+                                    // we still have to wait for the "next" to complete
```
akarnokd: Like with `Amb`.

akarnokd:
Benchmark results for the `Return` implementation:

```
BenchmarkDotNet=v0.11.2, OS=Windows 10.0.17134.345 (1803/April2018Update/Redstone4)
Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
Frequency=3513594 Hz, Resolution=284.6089 ns, Timer=TSC
.NET Core SDK=2.1.403
  [Host]     : .NET Core 2.1.5 (CoreCLR 4.6.26919.02, CoreFX 4.6.26919.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.5 (CoreCLR 4.6.26919.02, CoreFX 4.6.26919.02), 64bit RyuJIT


            Method |     Mean |     Error |    StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
------------------ |---------:|----------:|----------:|------------:|------------:|------------:|--------------------:|
 ToAsyncEnumerable | 488.0 ns | 3.9536 ns | 3.0867 ns |      0.0486 |           - |           - |               208 B |
            Direct | 154.5 ns | 0.9334 ns | 0.8731 ns |      0.0322 |           - |           - |               136 B |
          Iterator | 274.1 ns | 1.0725 ns | 0.9507 ns |      0.0319 |           - |           - |               136 B |
```

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Never.cs
bartdesmet: How about using a `TaskCompletionSource<bool>` to return a `Task<bool>` (and wrap it in a `ValueTask<bool>`) and using `Register` on the `CancellationToken` to call `SetCanceled` on it? That way, we won't need to use `Task.Run` and `Task.Delay`.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Return.cs
onovotny: The core interfaces will be in CoreFx fwiw

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Never.cs
akarnokd: Yes, that could work too.

Ix.NET/Source/Benchmarks.System.Interactive/AsyncReturnBenchmark.cs
```diff
@@ -0,0 +1,164 @@
...
+        internal CancellationToken token;
+
+        protected AsyncIterator()
+        {
+            _threadId = Environment.CurrentManagedThreadId;
+        }
+
+        public IAsyncEnumerator<TSource> GetAsyncEnumerator(CancellationToken token)
+        {
+            var enumerator = state == AsyncIteratorState.New && _threadId == Environment.CurrentManagedThreadId
```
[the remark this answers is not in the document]
onovotny: What is the penalty for doing this?

There have been a lot of work done across the board to minimize any extra allocations.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Return.cs
bartdesmet: Good question; we'll need to give that some thought. Same for the use of an iterator base class versus leveraging C# 8.0 iterator support when it lands and/or relying on types used by the compiler itself to build the state machine for async iterators (likely to end up in `System.Runtime.CompilerServices`).

For now, I'd prefer we stick with `System.Linq.Async` to only contain "standard operators" (i.e. an async mirror image of `System.Linq`) and leave extensions to the `Interactive` world. The most valuable extensions could be brought up as proposals for proper `System.Linq.*`.

Any operators that involve time should be considered separately whether they make sense in the `System.Linq.*` world. Any variants of existing operators that involve many sources and exploit concurrency (rather than pure sequential execution) should be reviewed at some point as well. A good example is the default behavior for `Zip`: concurrent or not?

Once interfaces are stable, we could check in with the .NET and C# folks for their opinions on the design stance for LINQ operators. There are many aspects to it:

* Aggregates have `Async` suffix or not, e.g. `SumAsync`. Return `Task<T>` or `ValueTask<T>`?
* Concurrent versus sequential nature of operators? `ZipConcurrent` and `Zip`? Etc.
* Use of `ValueTask<T>` or `Task<T>` in functions passed to operators?
* Overload hell for operators with many functions (e.g. `GroupJoin`). Which combinations to provide?
* Ensure overloads don't cause ambiguity when used with query comprehension syntax.

We can start this laundry list of things to check off elsewhere outside this PR.

akarnokd: https://github.com/dotnet/reactive/blob/IxAsyncCSharp8/Ix.NET/Source/System.Linq.Async/System/Linq/IAsyncPartition.cs#L13

Exposing the assembly is generally done via `[assembly: InternalsVisibleTo("...")]` but I got a warning about using proper assembly key or something.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Throw.cs
```diff
@@ -29,5 +30,36 @@ public static IAsyncEnumerable<TValue> Throw<TValue>(Exception exception)
...
+            private readonly ValueTask<bool> _moveNextThrows;
+
+            public TValue Current => default;
+
+            public ThrowEnumerable(ValueTask<bool> moveNextThrows)
+            {
+                _moveNextThrows = moveNextThrows;
+            }
+
+            public IAsyncEnumerator<TValue> GetAsyncEnumerator(CancellationToken cancellationToken = default)
```
bartdesmet: Nit: For implementations, we can omit the `= default` for the parameter. That way, we can make sure we don't end up calling this directly somewhere else, forgetting to pass down a cancellation token we have floating around. In fact, I took away the default value on the interface initially to spot many places where we accidentially omitted passing a token we had.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Throw.cs
```diff
@@ -29,5 +30,36 @@ public static IAsyncEnumerable<TValue> Throw<TValue>(Exception exception)
...
+                return TaskExt.CompletedTask;
+            }
+
+            public ValueTask<bool> MoveNextAsync()
+            {
+                // May we let this fail over and over?
+                // If so, the class doesn't need extra state
+                // and thus can be reused without creating an
+                // async enumerator
+                return _moveNextThrows;
```
bartdesmet: We should try to be consistent with the behavior of an iterator. We don't have async iterators yet:

```csharp
async IAsyncEnumerable<T> Throw<T>(Exception ex)
{
    throw ex;
    yield break; // unreachable
}
```

but have sync ones as a data point:

```csharp
IEnumerable<T> Throw<T>(Exception ex)
{
    throw ex;
    yield break; // unreachable
}
```

and the behavior is stateful:

```csharp
var e = Throw<int>(new Exception()).GetEnumerator();
e.MoveNext(); // throws
e.MoveNext(); // returns false
```

Ix.NET/Source/Benchmarks.System.Interactive/AsyncReturnBenchmark.cs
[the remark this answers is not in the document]
akarnokd: I guess this approach is biased towards `IAsyncEnumerable`s being typically used at most once and thus saving an allocation of a `IAsyncEnumerator`. On one hand, there is the cost of a new instance. On the other hand, there is the cost of retrieving the current managed thread id and doing comparison/memory activities.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Throw.cs
akarnokd: Sure. I'm not deliberately writing `= default` everywhere btw. I'm using VS' Alt+Enter > implement methods.

Ix.NET/Source/System.Interactive.Async/System/Linq/Operators/Throw.cs
akarnokd: Okay.

Ix.NET/Source/Benchmarks.System.Interactive/Benchmarks.System.Interactive.csproj
```diff
@@ -32,6 +32,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\System.Interactive.Async\System.Interactive.Async.csproj" />
     <ProjectReference Include="..\System.Interactive.Tests\System.Interactive.Tests.csproj" />
+    <ProjectReference Include="..\System.Linq.Async\System.Linq.Async.csproj" />
```
bartdesmet: Could you create a new `Benchmark.System.Linq.Async` project and move this work into there? That way, we can keep the project dependencies minimal and can easily slice-and-dice the repo to tear off the `System.Linq.Async` work for a possible move closer to the heart of .NET.

bartdesmet: Maybe it'd be best to do that work in a separate PR, so we keep this PR to a minimum.

Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Finally.cs
bartdesmet: Goal of `Finally` should be to be behaviorally equivalent to:

```csharp
static IEnumerable<T> Finally<T>(this IEnumerable<T>source, Action dispose)
{
    try
    {
        foreach (var x in source)
            yield return x;
    }
    finally
    {
        dispose();
    }
}
```

Output is:

```csharp
var xs = Finally(new[] { 1, 2, 3 }, () => { Console.WriteLine("Disposed!"); });

// nothing
xs.GetEnumerator().Dispose(); 

// prints "Disposed!"
var e = xs.GetEnumerator();
e.MoveNext();
e.Dispose();
```

PR #867, opened 2019-02-06
Discussion: IObservable/IObserver hybrid class similar to AsyncIterator.

quinmars:
While looking through the _AsyncIx_ code for C# 8, I stumbled across the [`AsyncIterator`](https://github.com/dotnet/reactive/blob/IxAsyncCSharp8/Ix.NET/Source/System.Linq.Async/System/Linq/AsyncIterator.cs) class. It is a hybrid of `IAsyncEnumerable` and `IAsyncEnumerator` and reduce the number of allocations by reusing the `IAsyncEnumerable` instance for the first enumerator. In this PR I do the same for `IObservable` and `IObserver`.

There are some downsides in doing that. It glues two things together that should be (at least mentally) kept separated. While the operators still only expose the `IObservable` interface, it'd be possible to access implementation details with a dynamic cast:

```csharp
    var query = obs.Where(x => x < 10);

    if (query is IDisposable disposable)
        disposable.Dispose();        // Oops
```

On the other hand it brings some reduction in allocations. What I like most is that it removes several lines of boiler plate code.

I used the name `Pipe` because it connects sources/producers and sinks. For the implementation I used the `Producer` class as base class and literally copy and pasted the code of the `Sink`, plus some minor additions, like the implemented `CreateSink` method. There are also some example usage for some random operators. If you like the idea I can transform more of them.

I marked the PR as discussion, because I'm not sure myself if it is a good idea. I definitely  don't like that operators would leak the `IDisposable` interface. That could lead to strange problems if one is using, for example, an auto disposing object store. I don't know, however, if such a thing exists and is in use. On the other hand the same would apply for the `AsyncIterator`. What I really like is the reduction of boiler plate code. The `Select` class only contains: the constructor, the clone method, and `OnNext` method.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Internal/Pipe.cs
+++ b/Rx.NET/Source/src/System.Reactive/Internal/Pipe.cs
@@ -1,0 +1,145 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information. 
+
+using System.Reactive.Disposables;
+using System.Threading;
+
+namespace System.Reactive
+{
+    /// <summary>
+    /// Base class for implementation of query operators.
+    /// </summary>
+    /// <typeparam name="TSource">The type of the elements in the source sequence.</typeparam>
+    /// <typeparam name="TResult">The type of the elements in the result sequence.</typeparam>
+    internal abstract class Pipe<TSource, TResult> : Producer<TResult, Pipe<TSource, TResult>>, IObserver<TSource>, ISink<TResult>, IDisposable
+    {
+        protected readonly IObservable<TSource> _source;
+
+        private IDisposable _upstream;
+        private volatile IObserver<TResult> _downstream;
+
+        public Pipe(IObservable<TSource> source)
+        {
+            _source = source;
+        }
+
+        protected override Pipe<TSource, TResult> CreateSink(IObserver<TResult> observer)
+        {
+            if (Interlocked.CompareExchange(ref _downstream, observer, null) == null)
+                return this;
+            else
+            {
+                var pipe = Clone();
+                pipe._downstream = observer;
+                return pipe;
+            }
+        }
+
+        public void Dispose()
+        {
+            if (Interlocked.Exchange(ref _downstream, NopObserver<TResult>.Instance) != NopObserver<TResult>.Instance)
+                Dispose(true);
+        }
+
+        protected abstract Pipe<TSource, TResult> Clone();
+
+        protected override void Run(Pipe<TSource, TResult> sink)
+        {
+            sink.SetUpstream(_source.SubscribeSafe(sink));
+        }
+
+        /// <summary>
+        /// Override this method to dispose additional resources.
+        /// The method is guaranteed to be called at most once.
+        /// </summary>
+        /// <param name="disposing">If true, the method was called from <see cref="Dispose()"/>.</param>
+        protected virtual void Dispose(bool disposing)
+        {
+            //Calling base.Dispose(true) is not a proper disposal, so we can omit the assignment here.
+            //Sink is internal so this can pretty much be enforced.
+            //_observer = NopObserver<TTarget>.Instance;
+
+            Disposable.TryDispose(ref _upstream);
+        }
+
+        public void ForwardOnNext(TResult value)
+        {
+            _downstream.OnNext(value);
+        }
+
+        public void ForwardOnCompleted()
+        {
+            _downstream.OnCompleted();
+            Dispose();
+        }
+
+        public void ForwardOnError(Exception error)
+        {
+            _downstream.OnError(error);
+            Dispose();
+        }
+
+        protected void SetUpstream(IDisposable upstream)
+        {
+            Disposable.SetSingle(ref _upstream, upstream);
+        }
+
+        protected void DisposeUpstream()
+        {
+            Disposable.TryDispose(ref _upstream);
+        }
+
+
+        public abstract void OnNext(TSource value);
+
+        public virtual void OnError(Exception error)
+        {
+            ForwardOnError(error);
+        }
+
+        public virtual void OnCompleted()
+        {
+            ForwardOnCompleted();
+        }
+
+        public IObserver<TResult> GetForwarder() => new _(this);
+
+        private sealed class _ : IObserver<TResult>
+        {
+            private readonly ISink<TResult> _forward;
+
+            public _(ISink<TResult> forward)
+            {
+                _forward = forward;
+            }
+
+            public void OnNext(TResult value)
+            {
+                _forward.ForwardOnNext(value);
+            }
+
+            public void OnError(Exception error)
+            {
+                _forward.ForwardOnError(error);
+            }
+
+            public void OnCompleted()
+            {
+                _forward.ForwardOnCompleted();
+            }
+        }
+    }
+
+    internal abstract class Pipe<TSource> : Pipe<TSource, TSource>
+    {
+        internal Pipe(IObservable<TSource> source) : base(source)
+        {
+        }
+
+        public override void OnNext(TSource value)
+        {
+            ForwardOnNext(value);
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


quinmars:
For some reasons the unit test `ReactiveTests.Tests.SubscribeOnTest.SubscribeOn_ControlScheduler' hung. Looking at the code, I cannot see that the PR changed any code related to it.

akarnokd:
Some comparative benchmarks would be great. In general, I don't know how much this architecture benefits the `AsyncIterator` is to begin with. Certainly the .NET memory allocator is not that bad and also assuming operator reuse is uncommon (i.e., no `retry`/`repeat` involved in the chain.)

quinmars:
I'll try to provide some benchmarks next week. Note, the current implementation does not provide a reuse of the observer. It is burned once it was used.

danielcweber:
I like how it removes the need to for a inheriting producer to pass it's own type as a type parameter to the base class (e.g. `All<TSource> : Producer<bool, All<TSource>._>`). I'm wondering: Is that already possible today, as a preparing step towards pipes? Don't really remember why we kept this.

Yes, the leaking of `IDisposable` could lead to funny problems, I remember Autofac automatically disposing my stuff and it drove me crazy for hours ;-). Also, it leaks `IObserver`. The implications are unclear to me. Generally, I think if external code thinks it's a good idea to randomly cast stuff to shoot itself in the foot, go ahead, should not be our concern....in theory.

When will we save allocations? When operator chains are built and subscribed exactly once. If chains are reused the savings will amortize to zero. I definitely use this create-subscribe-forget-pattern a lot, so I think it's definitely an interesting approach.

bartdesmet:
This particular optimization exists for `IEnumerable<T>` exactly for the reasons described in the discussion, namely the single iteration pattern that's rather common. All C# iterators generated for `IEnumerable<T>` (and later `IAsyncEnumerable<T>`) emit code that fuses both interface implementations with a lazy-cloning approach based on a check for the state field and the current thread ID (to allow interleaved enumerations on the same thread, e.g. `xs.Zip(xs, f)` to get unique enumerations over the shared sequence).

It does suffer though from leaking the transitive closure of interfaces related to enumeration:

```csharp
IEnumerable<int> xs = new[] { 1, 2 }.Where(x => x < 10);
Console.WriteLine(xs is IEnumerator<int>);
Console.WriteLine(xs is IDisposable);
```

will print `true` twice. This isn't a major problem though, because even disposing `xs` doesn't render it unusable. It simply transitions its internal state such that a subsequent attempt to start enumeration using `GetEnumerator` results in cloning of the object to start afresh. The one place where this can be potentially dangerous is when the enumerable instance is an alias for the enumerator instance and disposal through the enumerable will affect the enumerator:

```csharp
IEnumerable<int> xs = new[] { 1, 2 }.Where(x => x < 10);
IEnumerator<int> e = xs.GetEnumerator();
Console.WriteLine(object.ReferenceEquals(xs, e));
IDisposable d = (IDisposable)xs;
Console.WriteLine(e.MoveNext());
d.Dispose(); // oops!
Console.WriteLine(e.MoveNext());
```

For the observable world, applying the same optimization could be valuable and was in fact considered as part of the Rx 2.x performance work many years ago, under the assumption we'd be as safe as the corresponding enumerable world (i.e. disposing the observable sequence doesn't render it unusable for subsequent subscribers, and lazy cloning is performed to achieve this effect).

However, we steered away from it because it also leaks `IObserver<T>` and thus any reference to the `IObservable<T>` instance can be used to inject values into the currently attached observer, or into the first observer (depending on the implementation strategy where the original object could be reused by sequential subscribers, i.e. after disposing or termination, it'd be feasible to restart the observable with a new observer slotted in, thus limiting the cloning to the number of concurrent observers rather than unique subscribers).

This is strictly worse than early termination of an enumerator through abuse of `IDisposable` because arbitrary values can be injected into a query. The further rationale for considering this a potential problem is that Rx exposes first-class types that implement both interfaces, namely subjects, so it is very conceivable for users to use dynamic casts to figure out where an observable happens to be a subject merely by checking for the observer interface (and we did indeed find such use cases in some Windows 8 "Metro" store apps using Rx back then). A similar collision of interfaces does not exist in the enumerable world at the time of inventing iterators back in C# 2.0 (though it could make sense, and Ix does provide a "buffer" that combines both interfaces).

While it'd be great not having to worry about dynamic casts, there's a real risk involved. In fact, we hit a similar issue in CoreFx when specializing implementations of `Expression<T>` for various arities, causing a change in the inheritance hierarchy of `LambdaExpression`, which resulted in customers reporting a compatibility issue (see https://github.com/dotnet/corefx/issues/24481). This was even more contrived than a simple dynamic cast, but it's a good example of how innocent looking changes can cause compat breaks which may keep users from upgrading (or worse, drive them to abandoning the dependency).

My advice would be to err on the safe side. However, if performance benchmarks show a real significant value here, we could consider safe approaches to attempt object reuse (e.g. for sequential subscribers). Note though that there's a fundamental difference with the enumerable world in that there's always an external allocation of the observer that's given to us. In the enumerable world, we are the factory of the enumerator, so we control a bigger realm of allocations. So, there's a natural limit to what can be achieved here.

quinmars:
> When will we save allocations? 

You get basically the first sink for free. If you have three observers you only pay for two.

> All C# iterators generated for `IEnumerable<T>` (and later `IAsyncEnumerable<T>`) emit code that fuses both interface implementations with a lazy-cloning approach based on a check for the state field and the current thread ID (to allow interleaved enumerations on the same thread, e.g. `xs.Zip(xs, f)` to get unique enumerations over the shared sequence).

Interesting, I didn't know that `IEnumerable<T>` does already the same.

> However, if performance benchmarks show a real significant value here, we could consider safe approaches to attempt object reuse (e.g. for sequential subscribers).

The problem is that it is hard to track if an observer will no longer be used. The `Dispose` method works in Rx more like the cancellation token in AsyncIx and not like the `DisposeAsync` method, where you know that no ongoing or sequentiell calls to the enumerator will be performed.

quinmars:
I performed now the first perfomance test. I took the [Shakespeare plays scrabble](https://github.com/akarnokd/akarnokd-misc-dotnet/tree/master/rxnet_4_0_0) test of @akarnokd. And run it with Rx-4.1.2 and with the pipes version.

``` ini

BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.523 (1803/April2018Update/Redstone4)
Intel Core i5-5200U CPU 2.20GHz (Max: 0.50GHz) (Broadwell), 1 CPU, 4 logical and 2 physical cores
Frequency=2143476 Hz, Resolution=466.5319 ns, Timer=TSC
.NET Core SDK=3.0.100-preview-010184
  [Host]     : .NET Core 2.1.7 (CoreCLR 4.6.27129.04, CoreFX 4.6.27129.04), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.7 (CoreCLR 4.6.27129.04, CoreFX 4.6.27129.04), 64bit RyuJIT


```
| Method       | Run |     Mean |    Error |      Gen 0 |     Gen 1 | Allocated |
|------------- | ---:|---------:|---------:|-----------:|----------:|----------:|
|  RxNET 4.1.2 | 1   | 153.1 ms | 2.215 ms | 31000.0000 | 1000.0000 |  47.62 MB |
|  RxNET pipe  | 2   | 151.9 ms | 1.041 ms | 28500.0000 |  250.0000 |  43.49 MB |
|  RxNET 4.1.2 | 3   | 154.0 ms | 1.582 ms | 31250.0000 |  250.0000 |  47.62 MB |
|  RxNET pipe  | 4   | 154.1 ms | 1.203 ms | 28500.0000 |  250.0000 |  43.49 MB |

There is little to zero performance improvement with the pipes version, but it shows at least a memory saving of around 10%. The result is not particularly surprising, because we do not excpect any improvements in the through put of items.

The next test, that I will prepare, will mimic a ReactiveUI view model. With ReactiveUI you typically create a dozen pipelines in the constructor of one view model. Depending on the setup that number can grow rapidly. For example if you have a vm per list item. Even worse the vm is typically constructed on page switches where the CPU is already very busy. On the other hand, the through put is not so important, because it is usally triggered by user input, i.e., comparatively few events.

akarnokd:
I'd also stay on the safe side similar to [Bart's](https://github.com/dotnet/reactive/pull/867#issuecomment-462460075) comment.

quinmars:
I started to write the ReavtiveUI tests, but the first results were very scattered, without a real tendency. So without any evidence the performance argument is moot. While I still like the simplicity of the operator implementations, this probably does not outweigh the conter-arguments discussed above.

PR #882, opened 2019-04-10
Update Behavior/ReplaySubject to more eagerly handle disposed observers

### LINKED ISSUE #875: Unsubscribing from BehaviorSubject while OnNext is executing fails to unsubscribe immediately
Test case:
```
[Test]
public void UnsubscribeFromOnNext()
{
    var subject = new BehaviorSubject<int>(0);

    bool receivedValue = false;
    IDisposable subscription = null;
    subject.Subscribe(_ =>
    {
        if (subscription != null)
        {
            // Assume order of calls, even though it is not guaranteed (?),
            // otherwise the test doesn't make sense.
            Assume.That(receivedValue, Is.False);
            subscription.Dispose();
            subscription = null;
        }
    });
    subscription = subject.Subscribe(_ => receivedValue = true);

    receivedValue = false;
    subject.OnNext(0);
    subject.OnCompleted();

    Assume.That(subscription, Is.Null);
    Assert.IsFalse(receivedValue);
}
```

I get that the test case is a bit artificial and assumes execution order, but in a case where a subscription gets disposed along with a disposable object, this can end up calling code on the disposed object.

---
Some background info:
I initially encountered this issue in UniRx with `Subject`, so I started cross referencing with dotnet Rx. The underlying issue in UniRx is the use of immutable lists of observers, and indeed, `BehaviorSubject` uses `ImmutableList<IObserver<T>>`, which is a dangerous pattern. Perhaps adding a warning about its use might be a good idea :)

I also ran the test on all types deriving from `SubjectBase`, but this was the only failing case.

akarnokd:
This PR updates the `BehaviorSubject` and `ReplaySubject` to eagerly check for disposed observers while multiple observers are being notified. This handles the case when one observer disposes another one that would come next. The PR makes all 4 subject types consistent in this regard and adds the relevant unit tests.

Resolves #875.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Subjects/BehaviorSubject.cs
+++ b/Rx.NET/Source/src/System.Reactive/Subjects/BehaviorSubject.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information. 
 
 using System.Reactive.Disposables;
+using System.Threading;
 
 namespace System.Reactive.Subjects
 {
@@ -17,7 +18,7 @@ namespace System.Reactive.Subjects
 
         private readonly object _gate = new object();
 
-        private ImmutableList<IObserver<T>> _observers;
+        private ImmutableList<Subscription> _observers;
         private bool _isStopped;
         private T _value;
         private Exception _exception;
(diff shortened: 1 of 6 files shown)


danielcweber:
@onovotny Good to go even without a review by Bart?

akarnokd:
Bart said the original behavior was intended to mirror the delegate behavior, thus this should not go in and the async/publish subjects should be restored as well.

danielcweber:
Alright, didn't realize the review was going on in the issue-topic.

akarnokd:
Closing as the original behavior is the intended one.

PR #888, opened 2019-04-22
Initial work to support EnumeratorCancellation.

bartdesmet:
See:

* https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-04-15.md
* https://github.com/dotnet/csharplang/blob/master/meetings/2019/LDM-2019-04-29.md
* https://github.com/dotnet/standard/pull/1160

This PR adds the attribute for downlevel platforms, and leverages it for `IAsyncEnumerable<T>`-based iterators where the `Create(Core)` pattern using an `IAsyncEnumerator<T>`-based iterator has been used. The result will be reduction of allocations.

Goal is to ship with `HAS_ASYNC_ENUMERABLE_CANCELLATION` always set (and hence `#if` checks removed). Right now, the flag has to be set explicitly because we don't yet have a compiler drop that picks up on this attribute.

~~When we prepare for the final release in concert with C# 8.0 and .NET Core 3.0 shipping, we'll also revisit the `#if USE_ASYNC_ITERATOR` checks; we should be able to get rid of a lot of hand-rolled state machines then.~~

### DIFF
--- a/Ix.NET/Source/System.Linq.Async/System/Runtime/CompilerServices/EnumeratorCancellationAttribute.cs
+++ b/Ix.NET/Source/System.Linq.Async/System/Runtime/CompilerServices/EnumeratorCancellationAttribute.cs
@@ -1,0 +1,15 @@
+﻿#if !BCL_HAS_ENUMERATOR_CANCELLATION
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+    public sealed class EnumeratorCancellationAttribute : Attribute
+    {
+    }
+}
+
+#endif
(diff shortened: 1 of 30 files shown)


Ix.NET/Source/System.Linq.Async/System/Runtime/CompilerServices/EnumeratorCancellationAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace System.Runtime.CompilerServices
+{
+    [AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
+    public sealed class EnumeratorCancellationAttribute : Attribute
+    {
+    }
+}
+
+#endif
```
onovotny: I think we need an `#else` with a `TypeForwardedTo` here to unify like the rest of the types.

bartdesmet:
This is currently blocked on the lack of support for custom attributes on local functions, see https://github.com/dotnet/csharplang/issues/1888.

Also note that if we merge this PR, we should consider converting some of these local functions to be `static` to avoid an additional allocation. Other than the "local" nature of local functions (thus not having to come up with a name for another method), we'd pretty much not be using any of the features of local functions (closure, delegate conversion), so if we really care deeply about optimizing these query operators before we get custom attributes on local functions, we could turn them into non-local functions (with `*Core` method names).

PR #891, opened 2019-04-27
Simplified BaseBlocking, FirstBlocking, and LastBlocking

ts2do:
The code was needlessly bloated and over-engineered for its usage.


danielcweber:
Could you please provide some more rationale for your work? Just calling something "needlessly bloated and over-engineered" is a bold claim (I'm not suggesting it's wrong, though).

akarnokd:
The current structure was a port from Java and thus based on "Java feature X equivalent in C#". If there is a more lenient and C# way of accomplishing the job, improvements are always welcome.

ts2do:
My apologies. I agree that my original explanation was unproductive and not specific. I just wanted to make those classes a bit cleaner by making it a lighter wrapper for the event, ManualResetEventSlim being the most basic event for the job. Since the code using FirstBlocking<> and LastBlocking<> already had the subscription in a using block, a lot of the code wasn't accomplishing anything.

PR #892, opened 2019-04-27
Reduced redundant null checks

ts2do:


danielcweber:
I'm afraid I don't get the point of this PR. It says to remove redundancy, yet we end up with ~60 lines of added code?

ts2do:
By redundancy, I meant at runtime overhead. Some places where checking null on both sides of a method call (e.g. one more basic overload of Subscribe calling another overload with similar arguments). Other places were invoking a public AnonymousObserver<T> constructor with null checks on arguments which cannot be null.

I assumed that it would be best to eliminate such instances, but given 1) that a majority of the changes seem to be in initialization methods, 2) that such null checks aren't going to kill performance, and 3) that the CLR could easily optimize most of these duplicate checks away when inlining, I realize that these changes are not necessarily helpful and would understand if you would prefer to close this pull request.

Issue #808, opened 2018-09-03
AsyncRx.NET lib Throttle(dueTime) doesn't work


slango0513:
I am using AysncRx.NET in Orleans2 project to handle streams. This great library has significantly reduced the workload, although there are still things unfinished, among them Throttle is one. "dueTime" parameter of this method does not work, there is no reference inside the method.

PR #896, opened 2019-05-06
AsyncRx.NET: Introduces VirtualTimeAsyncScheduler, TestAsyncScheduler

NickDarvey:
# Background

I've used Rx.NET at an application level for quite a while, I started down this path because I wanted to get familiar with the codebase (read: I have no faffing idea what I'm doing). I saw that the throttle operator wasn't working (#808) and thought that might be a good first contribution. I thought it was just a parameter not being passed through, but I wanted a way of validating it, so I ended up implementing the bits needed to build a test scheduler too.

# Value

This PR introduces a virtual time scheduler base which allows for schedulers like test schedulers (included in this PR) and historic schedulers (not included in this PR).

~~It also fixes the throttle operator and introduces the bits of Microsoft.Reactive.Testing required to test it.~~

~~(I can also submit this as two PRs—introducing the test scheduler as one, fixing the throttle operator as another—if preferred.)~~

Edit: The fix for the throttle operator is in #898.

# Questions
## Rendezvous, TaskAwaitable
This may not be pertinent to the changes I've made, but I haven't managed to understand quite what [the 'rendezvous' bit ](https://github.com/dotnet/reactive/blob/5906af97e1b94779f961fac5e8c9ac6ad5e3dfcc/AsyncRx.NET/System.Reactive.Async.Concurrency/System/Reactive/Concurrency/AsyncScheduler.cs#L13) is for. (I see it ensures an awaitable's continuation is invoked by the scheduler, but I don't understand the consequences.) Nor when it's supposed to be used. (It looks like it's only used in the Linq layer for most but not every Task-returning function.)
Can anyone clarify this part for me?

Edit: Oh, maybe [an Ada kind of rendezvous](http://homepage.divms.uiowa.edu/~jones/syssoft/fall00/notes/17conc.html#The%20Rendezvous%20Construct)? I guess I have some reading to do...

## Recursive scheduling
I see the [`IAsyncScheduler`](https://github.com/dotnet/reactive/blob/f71f2d62fcec2eb44ca6eaced3b58b21e0372076/AsyncRx.NET/System.Reactive.Async.Interfaces/System/Reactive/Concurrency/IAsyncScheduler.cs#L10) deals with schedule functions with a signature like `(cancelation, action) -> disposable` rather than `(scheduler, state, action) -> disposable`.
If I understand, the schedulers in Rx.NET allow for recursive scheduling of work. Is that not a thing in AsyncRx.NET or am I misunderstanding how it does that?

Edit: I think this might actually be related to the previous question 🤔

# Interfaces
I can imagine there is an argument for strictly matching interfaces so that it's exactly Rx.NET except for the Task-returning bits, but I can also imagine AsyncRx.NET being seen as an opportunity to do away with some legacy which was only retained in Rx.NET for backwards compatibility.

In this PR, there are some places where I deviated slightly from the interfaces in Rx.NET. (I'll highlight the exact deviations with comments.) Please guide me as to whether it is too much (I need to strictly match Rx.NET) or too little (I've brought across some legacy bits which I shouldn't have).

# Implementation
A lot of this was copying and pasting from Rx.NET and updating invocations to be Task-returning. I tried to leave the implementation mostly as-is as I assume there were optimization reasons for it being written specific ways.


AsyncRx.NET/AsyncRx.NET.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26730.3
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28803.352
```
NickDarvey: I opened this in Visual Studio 2019. Is it okay to bump this version number or should this be a separate PR?

AsyncRx.NET/Microsoft.Reactive.Async.Testing/ITestableAsyncObservable.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// <summary>
+    /// Observable sequence that records subscription lifetimes and timestamped notification messages sent to observers.
+    /// </summary>
+    /// <typeparam name="T">The type of the elements in the sequence.</typeparam>
+    public interface ITestableAsyncObservable<T> : IAsyncObservable<T>
+    {
+        /// <summary>
+        /// Gets a list of all the subscriptions to the observable sequence, including their lifetimes.
+        /// </summary>
+        IReadOnlyList<Subscription> Subscriptions { get; }
```
NickDarvey: I didn't think we wanted anyone altering the contents of these collections from the outside, but in it Rx.NET it's an `IList<>`

AsyncRx.NET/System.Reactive.Async.Concurrency/System.Reactive.Async.Concurrency.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\System.Reactive.Async.Disposables\System.Reactive.Async.Disposables.csproj" />
```
NickDarvey: I saw that there was already a re-implementation CancellationAsyncDisposable in Concurrency, but as I also needed to use SingleAssignmentDisposable in Concurrency I thought it was time to take a dependency on Disposables.

AsyncRx.NET/System.Reactive.Async.Concurrency/System/Reactive/Concurrency/AsyncSchedulerBase.cs
```diff
@@ -63,19 +64,5 @@ protected virtual async Task<IAsyncDisposable> ScheduleAsyncCore(Func<Cancellati
         protected abstract Task Delay(TimeSpan dueTime, CancellationToken token);
 
         protected static TimeSpan Normalize(TimeSpan timeSpan) => timeSpan < TimeSpan.Zero ? TimeSpan.Zero : timeSpan;
-
-        private sealed class CancellationAsyncDisposable : IAsyncDisposable
```
NickDarvey: See https://github.com/dotnet/reactive/pull/896/files#r281156488

AsyncRx.NET/System.Reactive.Async.Concurrency/System/Reactive/Concurrency/PriorityQueueVirtualTimeAsyncScheduler.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Threading.Tasks;
+
+namespace System.Reactive.Concurrency
+{
+    /// <summary>
+    /// Base class for virtual time schedulers using a priority queue for scheduled items.
+    /// </summary>
+    /// <typeparam name="TAbsolute">Absolute time representation type.</typeparam>
+    /// <typeparam name="TRelative">Relative time representation type.</typeparam>
+    public abstract class PriorityQueueVirtualTimeAsyncScheduler<TAbsolute, TRelative> : VirtualTimeAsyncScheduler<TAbsolute, TRelative>
```
NickDarvey: This is lumped together with the base definition for the VirtualTimeScheduler in Rx.NET but that confused me multiple times when navigating to definitions. I think making it explicitly separate might make it easier for the next developer.

AsyncRx.NET/System.Reactive.Async.Disposables/System/Reactive/Disposables/IAsyncCancelable.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace System.Reactive.Disposables
+{
+    public interface IAsyncCancelable : IAsyncDisposable
```
NickDarvey: I added this interface as a way of exposing `IsDisposed` for `ScheduledItem`. I noticed it's not used in all places in Rx.NET, so I'm not sure if it's something that we should be bringing across.

Also, perhaps it should return a `Task<bool>`?

AsyncRx.NET/tests/Tests.System.Reactive.Async.Concurrency/VirtualTimeAsyncSchedulerTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                return new DateTimeOffset((absolute ?? string.Empty).Length, TimeSpan.Zero);
+            }
+
+            protected override char ToRelative(TimeSpan timeSpan)
+            {
+                return (char)(timeSpan.Ticks % char.MaxValue);
+            }
+        }
+
+        // Is this supposed to be a [Fact] with AsyncRX.NET?
```
NickDarvey: Is this supposed to be a fact even in the async version? (It is a fact in non-async Rx.) If it is, I have implemented something wrong because it's ending up with a different thread ID.

AsyncRx.NET/System.Reactive.Async.Concurrency/System/Reactive/Concurrency/Internal/ScheduledItem.cs
```diff
@@ -0,0 +1,144 @@
...
+        public async Task Invoke()
+        {
+            if (_disposable.IsDisposed) return;
+            var disposable = await _action().ConfigureAwait(false);
+            await _disposable.AssignAsync(disposable).ConfigureAwait(false);
+        }
+
+        public Task DisposeAsync() => _disposable.DisposeAsync();
+
+        public bool IsDisposed { get => _disposable.IsDisposed; }
```
NickDarvey: I see in the [ScheduledItem in Rx.NET ](https://github.com/dotnet/reactive/blob/f71f2d62fcec2eb44ca6eaced3b58b21e0372076/Rx.NET/Source/src/System.Reactive/Concurrency/ScheduledItem.cs#L157) that it's using the set of Disposable static methods to handle dispose-related race conditions however I have opted for using a SingleAssignmentDisposable. I believe those new methods are an optimisation (maybe?) and the behaviour will be the same with this implementation. I didn't bring those methods across with this PR as I think that would mean updating the other disposables to use it, we could do that in future.

AsyncRx.NET/System.Reactive.Async.Concurrency/System/Reactive/Concurrency/Internal/ScheduledItem.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Reactive.Disposables;
+using System.Threading.Tasks;
+
+namespace System.Reactive.Concurrency
+{
+    /// <summary>
+    /// Represents a work item that has been scheduled.
+    /// </summary>
+    /// <typeparam name="TAbsolute">Absolute time representation type.</typeparam>
+    internal sealed class ScheduledItem<TAbsolute> : IComparable<ScheduledItem<TAbsolute>>, IAsyncCancelable
```
NickDarvey: I haven't added the [IScheduledItem](https://github.com/dotnet/reactive/blob/f71f2d62fcec2eb44ca6eaced3b58b21e0372076/Rx.NET/Source/src/System.Reactive/Concurrency/IScheduledItem.cs) abstraction. Is it valuable?

danielcweber:
>I can also submit this as two PRs

That would be great, if there's a bugfix it should go in separately. Discussion may then continue on your feature PR.

bartdesmet:
Looks like a good start on bringing these things into Async Rx; thanks for the effort on this.

I suggest we proceed towards feature parity with classic Rx, and consider the duality with `IAsyncEnumerable<T>` in C# 8.0 after that gets wrapped up. I can pour in some design topics for the work we did here on the "Reactor" service, which also has asynchronous Rx notions (in fact, the work in this repo was a direct consequence of that work, bringing back a core Rx library with async support, without the service portions around it).

bartdesmet:
I'll revisit this PR hopefully in the not-too-distance future, now that there's a second wave of work on this (after `System.Linq.Async` has crystallized, among some other reasons that motivated revamping this work).

danielcweber:
Hi @NickDarvey, thanks for this contribution. Would you be able to rebase this PR onto the current main branch and resolve the conflicts? I was thinking about tackling the Async-Rx tests to some extent and this could use the schedulers you introduced.

PR #898, opened 2019-05-09
AsyncRx.NET: Fixes throttling by passing through dueTime to scheduler

### LINKED ISSUE #808: AsyncRx.NET lib Throttle(dueTime) doesn't work

NickDarvey:
Passes through `dueTime` parameter to the scheduler.

I verified this fixes the behaviour of the throttle operator [running the same test](https://github.com/dotnet/reactive/blob/ed1b84d8354a7778698dfc9da20cf6ed5ba9c93c/AsyncRx.NET/tests/Tests.System.Reactive.Async.Linq/ThrottleTests.cs) as in the Rx.NET codebase, via #896.

Resolves #808.


danielcweber:
Thx!

PR #907, opened 2019-05-25
Synchronize MostRecent when appropriate

ts2do:
I added a check to MostRecent<>.Run which will return a different sink to synchronize reads & writes on _value when the size of TSource exceeds the native word size (checked using `System.Runtime.CompilerServices.Unsafe.SizeOf<T>()`) to prevent potentially reading an intermediate value in the middle of a write operation.


ts2do:
If this doesn't warrant adding Unsafe as a dependency, it could instead either use locking for reading and writing _value (less code) indiscriminately or permit reference types & safe primitives.

akarnokd:
I'm certain we would like to keep Rx.NET a clean C# implementation without unsafe juggling.

PR #911, opened 2019-05-28
Ix providers code generation using T4

bartdesmet:
Use T4 to generate these APIs, just like we do for the asynchronous variant. Support for factories on the `IQueryProvider` interface - which is a non-standard query provider technique borrowed from Rx - is dropped for now (but is rarely used). We can reintroduce it across the board, but need to consider a few technical challenges (e.g. support of `EnumerableRewriter` to target these methods).


onovotny:
we needed the netstandard 2.1 for some of the Ix libraries as a couple methods moved to the BCL and we provide a trimmed facade for those platforms. may not be needed for all of them, but something to be aware of

bartdesmet:
Ah I see; let me look into that.

PR #915, opened 2019-06-05
Ix: Do not dispose the enumerator while enumerating in the ToObservable operator.

quinmars:
As pointed out by @danielcweber the current implementation of `ToObservable` may violate the contract of `IAsyncEnumerator`, because it is possible to call `DisposeAsync` during an on going `MoveNextAsync` call.

https://github.com/dotnet/reactive/pull/423#discussion_r290825529


Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
```diff
@@ -28,51 +29,40 @@ public ToObservableObservable(IAsyncEnumerable<T> source)
...
-
-                    if (hasNext)
+                    catch (Exception error)
                     {
-                        observer.OnNext(e.Current);
-
                         if (!ctd.Token.IsCancellationRequested)
                         {
-                            Core();
+                            observer.OnError(error);
```
bartdesmet: There's a violation of the Rx contract here. An exception due to calling `OnNext` above can be handled here and cause a subsequent call to `OnError` on the same observer. A failing observer shall never be given its own exception.

bartdesmet: I think the simplest change is to keep the original code but surround it with

```csharp
await using (var e = _source.WithCancellation(ctd.Token).ConfigureAwait(false))
{
    while (await e.MoveNextAsync())
    {
        // ...
    }
}
```

and get rid of the recursive call to `Core`, as is already done here.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
```diff
@@ -32,30 +31,35 @@ public IDisposable Subscribe(IObserver<T> observer)
 
                 async void Core()
                 {
-                    try
+                    // REVIEW: fire-and-forget DisposeAsync?
+                    await using (var e = _source.GetAsyncEnumerator(ctd.Token))
```
quinmars: `e.DisposeAsync()` can potentially throw without being catched.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
```diff
@@ -32,30 +31,35 @@ public IDisposable Subscribe(IObserver<T> observer)
...
-                        {
-                            observer.OnError(error);
-                        }
-                        return;
-                    }
 
-                    if (!ctd.Token.IsCancellationRequested)
-                    {
-                        observer.OnCompleted();
+                            observer.OnNext(e.Current);
```
quinmars: `e.Current` can potentially throw without being catched.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
danielcweber: I think is is ok, it shouldn't throw since we you are now adhering to the correct flow, never cancelling in-flight. If it does anyway, it's really exceptional and there's nothing reasonable that can be done anyway I guess.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
danielcweber: This should maybe be forwarded to OnError.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
quinmars: Well, we could forward the exception to the observer, instead of swallowing it.

danielcweber: I didn't mean swallow it, although effectively it is what will happen on the ThreadPool. But when DisposeAsync throws, the observer has either unsubscribed or been notified of stream termination already, right? So you can't forward.

quinmars: The enumerator can be diposed before, completion is pushed. As it was done in my first atempt. Then the catched exception can be forwarded.

danielcweber: FWIW, the sync counterpart doesn't attempt to catch anything: https://github.com/dotnet/reactive/blob/f650dded4d6d6ced6178f864f2eaecdbb57d5b43/Rx.NET/Source/src/System.Reactive/Linq/Observable/ToObservable.cs#L77

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
bartdesmet: Yes, all exceptions from the enumeration should end up being propagated using `OnError`.

It's unfortunate but historical that consumption of enumerable sequences involves two virtuals in the form of `MoveNext` and `Current`, and both of them can throw (rather than `TryGetNext` or a means to state that `Current` cannot throw). Alas, it's the world we live in, and we should handle all cases here.

bartdesmet: In fact, that was an oversight in the initial implementation here, so we can correct it in this PR.

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
bartdesmet: Exceptions for lifecycle operations such as `Get[Async]Enumerator` and `Dispose[Async]` can escape into the aether. Especially the latter, where exceptions during `Dispose` are different from exceptions due to retrieving values, and the two shouldn't be confused.

For the former of `Get[Async]Enumerator` there's often deferred propagation due to the way iterators work, i.e. an iterator enumerating over a sequence will see the call to the sequence's `Get[Async]Enumerator` occur in the iterator's first call to `MoveNext[Async]`, so exceptions caan naturally propagate there, because there's still a consumer and it's really an error in retrieving the data (i.e. not being able to start retrieving the data to begin with).

For `Dispose[Async]` things can be a bit more blurry, but a key observation is that we may be disposing because we reached an error state and called `observer.OnError` with such an error already, or because we have completed successfully and have called `observer.OnCompleted`. In both cases, the observer is in a terminal state and expects no further calls, so we can't bark an exception up its tree anymore either. Thus, exceptions from `Dispose[Async]` can just propagate. If the call to `Dispose[Async]` happens due to disposal of the operator, the exception may propagate to the caller. But if we're in some place removed from the user - e.g. on a threadpool - we have to let go of the exception there. In Rx, schedulers would enter the picture, with support for a `Catch` operator on schedulers, or global excpetion handler events on concrete scheduler types (akin to the thread pool having such handlers).

Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
```diff
@@ -28,51 +28,54 @@ public ToObservableObservable(IAsyncEnumerable<T> source)
...
+                                return;
+                            }
 
-                        if (!ctd.Token.IsCancellationRequested)
-                        {
-                            Core();
+                            T v;
+                            try
+                            {
+                                v= e.Current;
```
bartdesmet: You can combine these cases higher up with the first `try...catch` construct:

```csharp
bool hasNext;
T value = default;
try
{
    hasNext = await e.MoveNextAsync().ConfigureAwait(false);
    if (hasNext)
    {
        value = e.Current;
    }
}
…
```

quinmars: Ok, I will do that.

danielcweber:
Review by @bartdesmet still needed? His earlier review has been adressed by @quinmars if I see this correctly.

quinmars:
Somehow, github could not link the commit to @bartholinie. Does any one know how to make that work?

danielcweber:
"Jens Chr authored and quinmars committed 12 hours ago". Commit info is preserved, isn't it?

danielcweber:
The requested change was about the recursion and has been adresses, right? Would merge after checks pass.

quinmars:
Right the commit info is preserve, but @bartholinie is not listed in the contributors list.

PR #916, opened 2019-06-06
IxAsync.Timeout: propagate timeout cancellation to main source

akarnokd:
The Timeout operator did not cancel the source if it lost the race against the item timeout. Also the timeout timer was not cancelled if the downstream cancelled the whole sequence.


bartdesmet:
LGTM.

The remarks in the code around the case where the timeout delay task wins are along the same lines as I was thinking for `Amb` in the other PR, namely to abandon the losing task and let its exception go unhandled (thus ending up on global handlers on the task pool).

The following invariant would be great to have:

```csharp
source.Amb(Throw<T>(new TimeoutException()).Delay(timeout))
```

behaves identical to

```csharp
source.Timeout(timeout)
```

for all aspects of cancellation and error propagation of the loser.

akarnokd:
> The following invariant would be great to have:

Okay, but `source` has to be a single item only for `Amb` as `Timeout` plays the race for each source item over and over whereas `Amb` does it only for the first signal.

Anyway, adding such invariance test requires both PRs to be merged first.

PR #938, opened 2019-06-16
Fix nullable property in csproj files

bartdesmet:
Future is `<Nullable>` so changing those. See https://github.com/dotnet/roslyn/issues/35432.


onovotny:
Closing as out of date

PR #964, opened 2019-07-16
Convenience Using overloads

eugbaranov:
To cater for the cases when only one of resource/observable factories is asynchronous where other is not.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable.Creation.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable.Creation.cs
@@ -695,6 +695,62 @@ namespace System.Reactive.Linq
             return s_impl.Using(resourceFactoryAsync, observableFactoryAsync);
         }
 
+        /// <summary>
+        /// Constructs an observable sequence that depends on a resource object, whose lifetime is tied to the resulting observable sequence's lifetime. The resource is obtained through asynchronous method.
+        /// The <see cref="CancellationToken"/> passed to the asynchronous methods is tied to the returned disposable subscription, allowing best-effort cancellation at any stage of the resource acquisition or usage.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the elements in the produced sequence.</typeparam>
+        /// <typeparam name="TResource">The type of the resource used during the generation of the resulting sequence. Needs to implement <see cref="IDisposable"/>.</typeparam>
+        /// <param name="resourceFactoryAsync">Asynchronous factory function to obtain a resource object.</param>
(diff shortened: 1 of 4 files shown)


akarnokd:
Is there a reason these conveniences for you should live in dotnet/reactive? Even if so, there should be unit tests verifying the expected behaviors of the new additions.

eugbaranov:
I will add tests.

With respect to why should it live here, I think those help from the usability perspective when refactoring existing codebase.

Imagine if you have `Observable.Using(CreateConnection, CreateObservable)`.

Now, if you want to create connection asynchronously instead you can't just change it to `Observable.Using(CreateConnectionAsync, CreateObservable)` but forced to change other argument too, e.g. `Observable.Using(CreateConnectionAsync, x => Task.FromResult(CreateObservable(x)))` which makes it less readable.

akarnokd:
I'd just create an extension method in my project, or at most, in [my own library](https://github.com/akarnokd/reactive-extensions#operators) if I need many reuses.

eugbaranov:
In this particular case because it can't be chained with fluent syntax having it implemented elsewhere significantly reduces its visibility and it doesn't help in the refactoring case I've described.

Were we to have an API e.g. `IObservableBuilder<T> Observable.Builder<T>()` then all create operators could have been implemented as extension method on top of that, e.g.:

    IObservable<TResult> Using<TResult, TResource>(
        this IObservableBuilder<T> builder,
        Func<CancellationToken, Task<TResource>> resourceFactoryAsync,
        Func<TResource, IObservable<TResult>> observableFactory)

Not sure if such APIs were ever considered.

onovotny:
@bartdesmet any thoughts?

danielcweber:
I appreciate the effort, but I think we should probably avoid explicitly having every combination of sync/async parameters because it'll grow exponentially for complexer signatures.

IMHO, the sync case is already the "convenience" case, since the async overloads already provide all the expressive power there is.

@eugbaranov I agree that `Observable.Using(CreateConnectionAsync, x => Task.FromResult(CreateObservable(x)))` is aesthetically less pleasing, but maybe you'll find `Observable.Using(CreateConnectionAsync, async x => CreateObservable(x))` more acceptable?

eugbaranov:
> @eugbaranov I agree that `Observable.Using(CreateConnectionAsync, x => Task.FromResult(CreateObservable(x)))` is aesthetically less pleasing, but maybe you'll find `Observable.Using(CreateConnectionAsync, async x => CreateObservable(x))` more acceptable?

This still forces somewhat unexpected change to second argument when only changing first but besides that won't work in projects with 'treat warnings as errors' turned on.

akarnokd:
I agree with @danielcweber, I don't think such conveniences should be in Rx.NET proper and can live where such conveniences will be needed.

PR #969, opened 2019-07-18
Handle the case when EventLoopScheduler gets disposed with in-flight items

### LINKED ISSUE #286: EventLoopScheduler: unexpected Exception after Dispose
When calling Dispose on an EventLoopScheduler, that has at least one item in its working queue, it will throw an ObjectDisposedException. The exception is thrown from its worker thread (which gives us no chance to catch it).

Code to reproduce the problem:

``` c#
class Program
{
    public static void Main() {
        var event_stream = new Subject<string>();

        var subscription = Observable.Using(
            resourceFactory:    () => new EventLoopScheduler(),
            observableFactory:  scheduler => event_stream
                .ObserveOn(scheduler)
                .Do(LongRunningAction))
            .Subscribe();

        event_stream.OnNext("event 1");

        Thread.Sleep(TimeSpan.FromSeconds(1));
        subscription.Dispose(); // Scheduler is still busy!

        Console.ReadLine();
    }
    private static void LongRunningAction(string event_text) {
        Thread.Sleep(TimeSpan.FromSeconds(2));
        Console.WriteLine(event_text);
    }
}
```

I've opened a question at [stackoverflow](https://stackoverflow.com/questions/36364615/eventloopscheduler-unexpected-behavior-on-dispose-objectdisposedexception) (containing a work-around) some time ago. The workaround does not seem to catch all possible cases - I still get this (unhandled) exception from time to time:

```
System.ObjectDisposedException: Cannot access a disposed object.
   at System.Reactive.Concurrency.EventLoopScheduler.Schedule[TState](TState state, TimeSpan dueTime, Func`3 action)
   at System.Reactive.Concurrency.LocalScheduler.Schedule[TState](TState state, Func`3 action)
   at System.Reactive.Concurrency.Scheduler.<>c__DisplayClass50`1.<InvokeRec1>b__4e(TState state2)
   at System.Reactive.ScheduledObserver`1.Run(Object state, Action`1 recurse)
   at System.Reactive.Concurrency.Scheduler.<>c__DisplayClass50`1.<InvokeRec1>b__4d(TState state1)
   at System.Reactive.Concurrency.Scheduler.<>c__DisplayClass50`1.<>c__DisplayClass52.<InvokeRec1>b__4f(IScheduler scheduler1, TState state3)
   at System.Reactive.Concurrency.ScheduledItem`2.InvokeCore()
   at System.Reactive.Concurrency.EventLoopScheduler.Run()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

eugbaranov:
Resolves #286

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs
@@ -153,7 +153,7 @@ namespace System.Reactive.Concurrency
             {
                 if (_disposed)
                 {
-                    throw new ObjectDisposedException("");
+                    throw new ObjectDisposedException(nameof(EventLoopScheduler));
                 }
 
                 if (dueTime <= TimeSpan.Zero)
@@ -351,7 +351,15 @@ namespace System.Reactive.Concurrency
                     {
                         if (!item.IsCanceled)
                         {
-                            item.Invoke();
+                            try
+                            {
+                                item.Invoke();
+                            }
+                            catch (ObjectDisposedException ex) when (nameof(EventLoopScheduler).Equals(ex.ObjectName))
+                            {
+                                // Since we are not inside the lock at this point
+                                // the scheduler can be disposed before the item had a chance to run
+                            }
                         }
                     }
                 }
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Concurrency/EventLoopSchedulerTest.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Concurrency/EventLoopSchedulerTest.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Reactive.Concurrency;
 using System.Reactive.Disposables;
+using System.Reactive.Linq;
 using System.Threading;
 using Microsoft.Reactive.Testing;
 using Xunit;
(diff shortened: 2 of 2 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs
```diff
@@ -351,7 +351,15 @@ private void Run()
                     {
                         if (!item.IsCanceled)
                         {
-                            item.Invoke();
+                            try
+                            {
+                                item.Invoke();
+                            }
+                            catch (ObjectDisposedException ex) when (nameof(EventLoopScheduler).Equals(ex.ObjectName))
```
danielcweber: I have difficulties seeing why the Exception thrown in [Line 156](https://github.com/dotnet/reactive/pull/969/files#diff-ed22d4c5fdc8c84b0546c122201041d8R156) should bubble up here. After all, `item.Invoke` will invoke the wrapped action, not the queueing done above. But maybe I'm missing it? Is it inserted back into the queue somewhere? Been a while since I dealt with this code.

However, catching the expression with this specific `when` guard probably doesn't do any harm, so I guess it's fine if it solves the issue.

eugbaranov: It would still try to call into the original scheduler when trying to perform the work - which could have been disposed right after the lock has been released.

PR #1035, opened 2019-09-23
Fix a number of nullable warnings.

bartdesmet:
Fix various C# 8.0 warnings, most notably due to:

* Generated code using `MethodInfo` fields rather than `MethodInfo?`.
* Test code checking for `ArgumentNullException` by having to pass in `null`.
* Redundant `Debug.Assert(x != null)` asserts influencing nullability analysis for `x`.
* Misc. smaller cases.


onovotny:
Don't worry about the failing RxNET builds - known issue that's being dealt with separately.

PR #1067, opened 2019-10-31
Fix Generate (timed) crash upon disposing a long sequence

akarnokd:
Due to how recursive scheduling has to link up `IDisposable`s in the [`UserWorkItem`](https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Concurrency/UserWorkItem.cs#L29) class, a long running timed `Generate` would produce a long chain of `UserWorkItem`s and crash with stack overflow if the whole sequence itself is disposed.

The fix is to have the recursive scheduling in `Generate` not return the next step's `IDisposable` from the function but keep replacing a main reference with the newer and newer `IDisposable`s, and thus no link chain is formed.

The use of the SAD indirection is necessary because when the `Schedule` returns with the `IDisposable`, the payload may have already run and we would overwrite a newer step's `IDisposable`.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/Generate.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/Generate.cs
@@ -195,9 +195,19 @@ namespace System.Reactive.Linq.ObservableImpl
                 private bool _hasResult;
                 private TResult _result;
 
+                private IDisposable _timerDisposable;
+
                 public void Run(IScheduler outerScheduler, TState initialState)
                 {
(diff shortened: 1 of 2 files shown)


khellang:
Thanks for fixing this memory leak. This is what my app looked like with 4.1.0:

image

And this is what it looks like now (with 4.3.0-preview.53):

image

Any clue when this will be part of an official release?

khellang:
I left the app running over night with the newly released 4.3.1 release and this is the result:

image

Thanks again for fixing this and cutting a release so swiftly 😘

PR #1116, opened 2019-12-24
Avoid infinite generics recursion in SynchronizationContextScheduler

jkotas:
The infinitive generics recursion interacts poorly with AOT. The AOT compilers have hard figuring
out where the stop generating the code for generics with infinite recursion. They either fail or
produce large images by giving up once the generics get too complex.

This change reverts a small part of #500 that introduced infinite generic recursion and adds comment.

Fixes https://github.com/dotnet/corert/issues/7920


onovotny:
Thanks! Will merge once the PR build completes

onovotny:
Also will fix ##1113

onovotny:
This is fixed now in 4.2.2 and 4.3.2.

PR #1121, opened 2020-01-03
Fix another case of infinite generics recursion.

PJB3005:
This time in CatchSchedulerLongRunning. See #1116

**I have not built or tested this at all because I literally couldn't figure out how to (I'm on Linux), but I hope it works/is helpful.**


danielcweber:
Thx a lot for spotting this.

PR #1165, opened 2020-03-28
move to verify

SimonCropp:
@clairernovotny this approach does not use pdbs to determine the path. can u verify it work for your deterministic builds?

if it does, and u are happy with this as an approach, i will finish the diffplex logging and u can merge


clairernovotny:
The approach seems fine to me, thank you! There's a unit test failing though because the `.cs` files are missing a license header.

SimonCropp:
ye fixed the license header

Rx.NET/Source/tests/Tests.System.Reactive/Tests/LicenseHeaderTest.cs
```diff
@@ -53,7 +53,7 @@ private int ScanPath(string path, StringBuilder error)
                     || file.Contains("AssemblyInfo.cs")
                     || file.Contains(".Designer.cs")
                     || file.Contains(".Generated.cs")
-                    || file.Contains(".approved.cs")
+                    || file.Contains(".verified.cs")
```
clairernovotny: Cool, LGTM!

SimonCropp:
ok. i will add a extension point to Verify so u can intercept the files on failure.

SimonCropp:
unfortunately in the end i had to add the following to the csproj to get this to work

```
<ContinuousIntegrationBuild>false</ContinuousIntegrationBuild>
<Deterministic>false</Deterministic>
```

The reason being is that Deterministic seems to not just clean pdb paths (which breaks approvaltests), it also cleans CallerFilePath attributes (which breaks Verify).

So with that workaround in mind, it is up to you if you want to use Verify or ApprovalTests

clairernovotny:
What's the difference between ApprovalTests or Verify?

SimonCropp:
https://github.com/SimonCropp/Verify/blob/master/docs/compared-to-approvaltests.md 

disclaimer: i am a maintainer on both

clairernovotny:
Cool, I'll stick with Verify. For now, I can set `Deterministic` to false in the Approvals project. Might be useful to think about an alternative way to get the directory needed though, perhaps with build-time env variables. You can detect deterministic paths by looking for `/_/` as the root and take action accordingly.

PR #1177, opened 2020-04-02
Fix infinite recursive generics in CatchScheduler

### LINKED ISSUE #1133: CoreRT build never finishes.
#### Bug

> Which library version?
4.3.0-preview-* (ran dotnet add package System.Reactive -v "4.3.0-preview*" after adding the preview NuGet feed.)

macOS Catalina

> What is the use case or problem?
.NET CoreRT build.

> What is the expected outcome?
Working successfully.

> What is the actual outcome?
Takes forever and never finishes.

> Part of the CoreRT ILC debug output:

[log.txt](https://github.com/dotnet/reactive/files/4114259/log.txt)

`

hez2010:
#### Bugfix
Fix infinite recursive generics in CatchScheduler

Fixes #1133

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/CatchScheduler.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/CatchScheduler.cs
@@ -108,7 +108,8 @@ namespace System.Reactive.Concurrency
                     _catchScheduler = scheduler;
                     _action = action;
 
-                    Disposable.SetSingle(ref _cancel, scheduler._scheduler.SchedulePeriodic((@this: this, state), period, tuple => tuple.@this?.Tick(tuple.state) ?? default));
+                    // Note that avoiding closure allocation here would introduce infinite generic recursion over the TState argument
+                    Disposable.SetSingle(ref _cancel, scheduler._scheduler.SchedulePeriodic(state, period, state1 => this.Tick(state1).state ?? default));
                 }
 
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/Concurrency/CatchScheduler.cs
```diff
@@ -108,7 +108,7 @@ public PeriodicallyScheduledWorkItem(CatchSchedulerPeriodic scheduler, TState st
                     _catchScheduler = scheduler;
                     _action = action;
 
-                    Disposable.SetSingle(ref _cancel, scheduler._scheduler.SchedulePeriodic((@this: this, state), period, tuple => tuple.@this?.Tick(tuple.state) ?? default));
+                    Disposable.SetSingle(ref _cancel, scheduler._scheduler.SchedulePeriodic(state, period, state1 => this.Tick(state1).state ?? default));
```
clairernovotny, line 111:
```suggestion
                    // Note that avoiding closure allocation here would introduce infinite generic recursion over the TState argument
                    Disposable.SetSingle(ref _cancel, scheduler._scheduler.SchedulePeriodic(state, period, state1 => this.Tick(state1).state ?? default));
```

hez2010: Done.

PR #1299, opened 2020-09-25
Reduce #if checks

bartdesmet:
Remove `#if` checks that are always true. Flip the `#if` checks to encode the exceptional case rather than the common case, thus reducing the number of `#define`s (e.g. `NO_TRACE` and `HAS_REMOTING` are the exceptions to the rule).

NB: For the time being, I'm leaving `HAS_WINRT` alone. I'll review these cases separately, because they're often combined with `CRIPPLED_REFLECTION` which also needs some closer look. If we can deprecate UAP 10.0 at some point, a lot more `#if` checks can go away.

NB: The `PREFER_ASYNC` stuff was used to control `[Obsolete]` attributes on various blocking members. I'll look into the steps towards final removal of these, given they've been marked as such for a long time (will review the history to make a call; alternatively, we can make the warning stronger and start removing a lot of code by layering the blocking ones on top of the non-blocking ones using the pattern we expect users to write, e.g. `.GetAwaiter().GetResult()`).

### DIFF
--- a/Rx.NET/Source/Directory.build.targets
+++ b/Rx.NET/Source/Directory.build.targets
@@ -15,24 +15,28 @@
   <PropertyGroup>  
     <Product>$(AssemblyName) ($(TargetFramework))</Product>
   </PropertyGroup>
+
+  <PropertyGroup>
+    <DefineConstants>$(DefineConstants);HAS_WINRT</DefineConstants>
+  </PropertyGroup>
     
   <PropertyGroup Condition="'$(TargetFramework)' == 'net46'">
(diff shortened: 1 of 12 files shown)


bartdesmet:
I'll leave it in place for now, though I may try to reduce the places it's needed (if any).

clairernovotny:
@bartdesmet I just checked in .NET 5 support -- so you'll need to make sure you're using 16.8p3. That adds the .NET 5 targets, the main one does define `HAS_WINRT` since it's the cross platform one, and the `net5.0-windows10.0.19041` does -- the latter one supports all of the Windows UI models: UWP/WPF/Winforms.

PR #1314, opened 2020-09-28
Fix Observable.FromAsync documentation grammar

joelhoisko:
Fixed grammar mistake in `Observable.ForAsync` documentation. I was reading up about how the `ForAsync` works and I noticed a typo in the docs.

Some parts of the docs use the wording `Convert a X to an Y` while others use `Convert the X to an Y`, not quite sure which one is the wanted behavior here, I'm fine with the `an X to an Y`.


bartdesmet:
Thanks for the fix. Looks good to me.

PR #1326, opened 2020-09-30
Use static lambdas.

bartdesmet:
Check if the CI build does already support C# 9.0 features. If so, conversion to static lambdas (to help identify cases where we have unexpected or avoidable closures) can continue.


bartdesmet:
First pass over a subset of the code, while addressing a few unnecessary closures in the process. Future PRs will continue this work:

1. Mark more existing lambdas as `static` where possible.
2. Try to convert closure-capturing lambdas to `static` where possible and worthwhile.
3. Optionally annotating the non-`static` lambdas with a clear indicator of a closure (`/* closure */`) so these can be easily CTRL-F'd.

bartdesmet:
Possible flaky test to be investigated:

EventLoop_ScheduleTimeAndOrderedInFlightActions

PR #1353, opened 2020-10-02
AsyncRx: Avoid closure allocations in Linq-operators when only the source-observable is captured.

danielcweber:
This avoids quite a few allocations of closures for Linq-operator-creation when the only object that would have been captured is the source-observable. More closure-allocation improvements may follow.


bartdesmet:
LGTM. We can also mark lambdas without a closure as static from now on.

danielcweber:
I'll update this PR tomorrow to enable C# 9

danielcweber:
>I'll update this PR tomorrow to enable C# 9

Did it in a separate PR in #1365 because it'll need separate review w.r.t. all the CI properties. @clairernovotny Would you please have a look?

danielcweber:
Lambdas have been marked static.

cabauman (approved):

PR #1360, opened 2020-10-02
More static lambdas in Rx.

bartdesmet:
Mark some more lambdas as `static`.


bartdesmet:
Flaky test or bug in `ReactiveTests.Tests.ObserveOnReactiveTest.ObserveOn_LongRunning_DisableOptimizations`.

PR #1395, opened 2020-10-15
Avoid even more closures with an additional overload of AsyncObservable.Create

danielcweber:
There's more room for improvement, especially wrt the T4-generated operators.

### DIFF
--- a/AsyncRx.NET/System.Reactive.Async.Linq/System/Reactive/Linq/AsyncObservable.cs
+++ b/AsyncRx.NET/System.Reactive.Async.Linq/System/Reactive/Linq/AsyncObservable.cs
@@ -18,7 +18,7 @@ namespace System.Reactive.Linq
 
                 public AsyncObservableImpl(IAsyncObservable<TSource> source, Func<IAsyncObservable<TSource>, IAsyncObserver<TResult>, ValueTask<IAsyncDisposable>> subscribeAsync)
                 {
-                    _source = source;
+                    _source = source ?? throw new ArgumentNullException(nameof(source));
                     _subscribeAsync = subscribeAsync ?? throw new ArgumentNullException(nameof(subscribeAsync));
                 }
 
@@ -39,8 +39,8 @@ namespace System.Reactive.Linq
 
                 public AsyncObservableImpl(IAsyncObservable<TSource> source, TState state, Func<IAsyncObservable<TSource>, TState, IAsyncObserver<TResult>, ValueTask<IAsyncDisposable>> subscribeAsync)
                 {
-                    _source = source;
                     _state = state;
+                    _source = source ?? throw new ArgumentNullException(nameof(source));
                     _subscribeAsync = subscribeAsync ?? throw new ArgumentNullException(nameof(subscribeAsync));
                 }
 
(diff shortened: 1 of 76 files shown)


bartdesmet (changes requested):

AsyncRx.NET/System.Reactive.Async.Linq/System/Reactive/Linq/AsyncObservable.cs
```diff
@@ -45,6 +67,19 @@ public static IAsyncObservable<T> Create<T>(Func<IAsyncObserver<T>, ValueTask<IA
             return new AsyncObservableImpl<TSource, TResult>(source, subscribeAsync);
         }
 
+        internal static IAsyncObservable<TResult> Create<TSource, TResult, TState>(IAsyncObservable<TSource> source, TState state, TResult dummy, Func<IAsyncObservable<TSource>, TState, IAsyncObserver<TResult>, ValueTask<IAsyncDisposable>> subscribeAsync)
```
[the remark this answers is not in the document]
danielcweber: >I'm not keen on the TResult dummy witness parameter simply for type inference purposes

Me neither, but saw it as superior than cluttering the code with explicit types.
I like the idea of builder structs, but I would try to do it in reverse: Instead of `WithState(state)` (which will have to copy the state around), I'd have a method `Build<TResult>` that would return the same builder instance (without any state), and one could go from there (e.g. `Build<TResult>().From(source, state, (source, state, observer) =>... )`. I think that could work.

>But I think we'd ultimately just have to get rid of the Create with delegate altogether and implement a base class instead,

Actually, I liked the way it's done in AsyncRx vs. how it's done in classic Rx with all its sinks and stuff. I like how the Observable is now generic while the Observers implement all the logic. It does help clarity a lot. Once the dummy parameters are gone, I'm not sure I understand the benefit of having custom `AsyncObservableBase<TResult>` implementations for each operator.

danielcweber:
@bartdesmet First impression of how dummy parameters can be avoided. Synatx is currently `Build<TResult>.From(....)` though the final naming is to be determined.

bartdesmet:
The builder pattern looks good and concise to me. I'm open to naming suggestions. Given it's internal, we have flexibility to rename it any time. My initial thought was CreateAsyncObservable<T>.From(...). Verbose but clear.

danielcweber:
Renamed it.

In a future update to this, I'd like to remove the distinction between source and state because this currently prevents easy updates to most T4-generated operators.

AsyncRx.NET/System.Reactive.Async.Linq/System/Reactive/Linq/AsyncObservable.cs
```diff
@@ -9,23 +9,70 @@ namespace System.Reactive.Linq
...
+            }
+
+            public static IAsyncObservable<TResult> From<TSource>(IAsyncObservable<TSource> source, Func<IAsyncObservable<TSource>, IAsyncObserver<TResult>, ValueTask<IAsyncDisposable>> subscribeAsync)
             {
-                _source = source;
-                _subscribeAsync = subscribeAsync ?? throw new ArgumentNullException(nameof(subscribeAsync));
+                if (source == null)
+                    throw new ArgumentNullException(nameof(source));
+
+                if (subscribeAsync == null)
```
bartdesmet, line 61: Nit: there are two checks for `null` on this (also in the constructor). We can remove one.

AsyncRx.NET/System.Reactive.Async.Linq/System/Reactive/Linq/AsyncObservable.cs
```diff
@@ -9,23 +9,70 @@ namespace System.Reactive.Linq
...
-            protected override ValueTask<IAsyncDisposable> SubscribeAsyncCore(IAsyncObserver<TResult> observer)
+            public static IAsyncObservable<TResult> From<TSource, TState>(IAsyncObservable<TSource> source, TState state, Func<IAsyncObservable<TSource>, TState, IAsyncObserver<TResult>, ValueTask<IAsyncDisposable>> subscribeAsync)
             {
-                if (observer == null)
-                    throw new ArgumentNullException(nameof(observer));
+                if (source == null)
+                    throw new ArgumentNullException(nameof(source));
 
-                return _subscribeAsync(_source, observer);
+                if (subscribeAsync == null)
```
bartdesmet, line 72: Same as above. Can eliminate redundant checks and put all of them next to each other.

bartdesmet (approved):
LGTM. Just a few minor remarks.

bartdesmet:
A logical next step to move AsyncRx.NET forward is to invest in the test framework - as closely mirroring the Rx one as possible to ensure familiarity - and have some good test coverage prior to a lot more surgery. If we're ambitious, we could try to get the virtual time tests for query operators in a centralized place so they can be reused across Rx and AsyncRx. Their behavior should be equivalent except for a few rare places (where we can have minor Rx/AsyncRx-specific test cases remain decentralized). The exact mechanism to achieve this is up in the air; I can think of a few ways.

danielcweber:
Addressed the remarks.

Would you mind labelling this hacktoberfest-accepted? I like the shirts...

danielcweber:
@bartdesmet 
>The exact mechanism to achieve this is up in the air; I can think of a few ways

One thing that comes to my mind is that, since sharing test sources between projects is probably impossible, we might use an approval test framework, so that all the assertions on the received messages that we have in the tests today will instead be *.verified files, and those might be shared. Just brainstorming.

PR #1396, opened 2020-10-15
Only specify usewpf/usewinforms for the platforms where we're testing it

clairernovotny:
Fix the build


bartdesmet:
LGTM. Can merge?

PR #1406, opened 2020-10-30
Replace most uses of Disposable.Utils

danielcweber:
As discussed in #1373, this replaces lots of uses of Disposable.Utils by more structs that reveal the intention (Serial, Multiple etc.).

Current stats:

GetValue - 1 reference (was 1)
GetValueOrDefault - 2 references (was 3)
SetSingle - no references (was 48)
TrySetSingle - 19 references (was 22)
TrySetMultiple - 11 references (was 16)
TrySetSerial - 14 references (was 31)
GetIsDisposed - no references (was 18 )
Dispose - 25 references (was 111)

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/CatchScheduler.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/CatchScheduler.cs
@@ -97,7 +97,7 @@ namespace System.Reactive.Concurrency
         {
             private sealed class PeriodicallyScheduledWorkItem<TState> : IDisposable
             {
-                private IDisposable _cancel;
+                private SingleAssignmentDisposableValue _cancel;
                 private bool _failed;
 
                 private readonly Func<TState, TState> _action;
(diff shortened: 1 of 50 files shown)


bartdesmet:
Let's avoid the breaking change on the public API surface (it's been that way since Rx v1.0 and code may have been written that relies on the null-coalescing behavior), so we can do the `OrDefault` bit over there on the `Disposable` property getter. Internally, when using the `*DisposableValue` wrapper, we can drop the requirement if we properly use `?.Dispose()` in these places. Externally, the non-`Value` public equivalent does the coalescing in its getter. Makes sense?

bartdesmet:
IIRC, we made the breaking change in v2.0 (introducing null-coalescing behavior) because some folks stepped on a `NullReferenceException` when `Disposable` unexpectedly returned `null` (and having the dispose attempts possibly originate on different threads makes it hard to remember it had been called before by some other thread, so assuming idempotent behavior seemed a reasonable thing for them to do).

Though it was a questionable pattern at the time (to get a `Disposable` out of a wrapper and dispose that, rather than the wrapper itself), it seemed okay for that case. (This is all from vague memory; I've done a quick search in my mailbox but the mail conversation I vaguely recall must have been lost in the dust of time, ~8 years.)

Hence my conservative stance of the breaking change potential. We allowed ourselves a bit more in the v2.0 days when things were relatively fresh and the adoption curve was still picking up (but, as you saw in the code, we annotated all places where we made a breaking change to be able to communicate all of these properly in release notes). We can reshuffle the internals now, but retain the public behavior as-is.

bartdesmet:
> I could imagine a method on SerialDisposableValue eg. TrySetFirst.

That sounds reasonable.

danielcweber:
Alright I've rebased and skipped the breaking change commit.

bartdesmet:
LGTM. Feel free to do the work in a few PRs (with some stats to track progress, if that's quick to do). Makes it easier to review as well (I do miss some good tooling in the UI here - we do have it internally on other service "wrappers" around source control - to see the diff between subsequent pushes to see what's new since a previous review on a prior iteration).

danielcweber:
Alright, so I will then remove the [WIP] from this and in a subsequent PR add the proposed new TrySetFirst methods.

danielcweber:
Tagged if you don't mind (I have a streak with those shirts ;-))

bartdesmet:
Stats are looking good. Good to merge.

PR #1407, opened 2020-10-30
Even less Disposable.Utils

danielcweber:
Contains previous unmerged work.

Introduces TrySetFirst methods on (Serial, Multiple)DisposableValue to make mixed usage scenarios possible.
What remains are usages that mostly observe the Try... return value or do custom Interlocked logic. This can possibly be mitigated too but it's not as mechanic as these previous changes.

GetValue - 1 reference (was 1)
GetValueOrDefault - 2 references (was 2)
SetSingle - no references (was none)
TrySetSingle - 6 references (was 19)
TrySetMultiple - 5 references (was 11)
TrySetSerial - 6 references (was 14)
GetIsDisposed - no references (was none )
Dispose - 12 references (was 25)

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs
+++ b/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs
@@ -208,7 +208,7 @@ namespace System.Reactive.Concurrency
 
             private TState _state;
             private TimeSpan _next;
-            private IDisposable _task;
+            private MultipleAssignmentDisposableValue _task;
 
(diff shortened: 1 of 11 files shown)


bartdesmet:
Stats are looking good. Can do an in-depth review after rebasing with the other pending PR, so the delta gets smaller. We can consider if the remaining ones are worth looking into after dealing with all the mechanical changes as you pointed out as well.

danielcweber:
Rebased.

bartdesmet (approved):
LGTM.

PR #1408, opened 2020-10-30
Change signature of Observable.FirstOrDefaultAsync to a IObservable<TSource?> return type.

danielcweber:
API approval file has changed! Not sure if nullability on the surface is already considered breaking. But this is the only correct method signature.

More to come.


bartdesmet:
LGTM; the API approval file change is expected.

Let's add the `IQbservable<T>` changes in the same PR so these stay in sync. Okay to edit the generated file by hand; HomoIcon doesn't understand nullable attributes (yet) anyway.

We can also do `LastOrDefault` and `SingleOrDefault` in this PR. That would complete the "family" or related operators in one go.

danielcweber:
Closing in favour of #1409.

PR #1409, opened 2020-10-30
Change signature of ...OrDefault/DefaultIfEmpty to a IObservable<TSource?> return type.

danielcweber:
There's a warning suppression which I think is fine in this case.


Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.StandardSequenceOperators.cs
```diff
@@ -24,14 +24,16 @@ public virtual IObservable<TResult> Cast<TResult>(IObservable<object> source)
...
-        public virtual IObservable<TSource> DefaultIfEmpty<TSource>(IObservable<TSource> source)
+        public virtual IObservable<TSource?> DefaultIfEmpty<TSource>(IObservable<TSource> source)
         {
-            return new DefaultIfEmpty<TSource>(source, default!);
+            return new DefaultIfEmpty<TSource>(source, default);
         }
 
         public virtual IObservable<TSource> DefaultIfEmpty<TSource>(IObservable<TSource> source, TSource defaultValue)
         {
+#pragma warning disable CS8619 // Nullability of reference types in value doesn't match target type.
```
bartdesmet, line 34: We can solve it the other way around. Have the `DefaultIfEmpty` constructor take in a `TSource` rather than a `TSourece?` and then use `default!` in the overload above. That's less change, and given we don't access `_defaultValue` ourselves, but just flow it, it suffices for the API signatures to have the proper annotations.

bartdesmet: FWIW, `System.Linq` takes the same approach (as do we in other places around the code):

https://github.com/dotnet/runtime/blob/e7204f5d6fcaca5e097ec854b3be6055229fc442/src/libraries/System.Linq/src/System/Linq/DefaultIfEmpty.cs#L12

bartdesmet:
Let's add the `IQbservable<T>` changes in the same PR so these stay in sync. Okay to edit the generated file by hand; HomoIcon doesn't understand nullable attributes (yet) anyway.

bartdesmet:
Keeping track of a TODO list:

* `SingleOrDefault`
* `FirstOrDefault`
* `LastOrDefault`
* `ElementAtOrDefault`
* `DefautlIfEmpty`

Both for `IObservable<T>` and `IQbservable<T>`.

danielcweber:
I think I got them all.

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.StandardSequenceOperators.cs
```diff
@@ -24,9 +24,9 @@ public virtual IObservable<TResult> Cast<TResult>(IObservable<object> source)
 
         #region + DefaultIfEmpty +
 
-        public virtual IObservable<TSource> DefaultIfEmpty<TSource>(IObservable<TSource> source)
+        public virtual IObservable<TSource?> DefaultIfEmpty<TSource>(IObservable<TSource> source)
         {
-            return new DefaultIfEmpty<TSource>(source, default!);
+            return DefaultIfEmpty(source, default!);
```
bartdesmet, line 29: Let's expand this one to:

```csharp
return new DefaultIfEmpty<TSource?>(source, defaultValue!);
```

rather than calling from one virtual to another. If we end up restoring the functionality around `IQueryLanguage` (for debuggers and visualizers), we don't want to dispatch into interceptors twice. I.e. it'd look like you "break into the Rx debugger" twice for `DefaultIfEmpty`.

bartdesmet (commented):
LGTM other than one small thing to tweak in `QueryLanguage`.

If I get some spare cycles over the holiday period, I'll try to do some write-up on the original debugger infrastructure (which we never shipped after releasing Rx 2.0; it fell through the cracks of people moving around and reorgs happening, and the prototype of it never ended up in the OSS domain), which is based on the indirection provided by `IQueryLanguage`. The community maintainers can then decide if it's worth their time and effort to re-create this.

Long story short, the debugger interceptors can be auto-generated in a way similar to HomoIcon. They end up wrapping inputs and outputs of operators with a facility that's akin to `Defer`, `Finally`, and `Do`, allowing them to see the lifecycle (`Subscribe`/`Dispose`) and events flowing through (via `Do`). Selectors such as the one on `SelectMany` get wrapped with a delegate that does some additional stuff to associate the observable returned from the selector with the parent observable. (Similar tricks for subjects returned by `Window` and `GroupBy`.) The end result is that the whole dynamic object graph of operators gets a shadow representation in a "debugger assist" assembly, and it can be used to visualize a running Rx computation (which may dynamically change shape due to the behavior of various operators). And from within that visualization, one can set breakpoints on nodes (operators, which have lifecycle events such as `Subscribe` and `Dispose`) and edges (carrying events, governed by `Do` wrappers inserted).

Rx.NET/Source/src/System.Reactive/Linq/QueryLanguage.StandardSequenceOperators.cs
danielcweber: Done

bartdesmet (approved):
LGTM. Thanks again!

PR #1481, opened 2021-01-28
Remove redundant NuGet package references.

teo-tsirpanis:
`System.Runtime.InteropServices.WindowsRuntime` was added 3 and a half years ago in https://github.com/dotnet/reactive/commit/9534e68caf0d877ba0c010da8e0ba4eeda76fffe with no clear reason why. The package itself hasn't been updated since November 2016. Its presence bloats the package dependency graph and might pose problems for those who need their package dependencies to have SPDX license expressions.

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
+++ b/Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
@@ -35,18 +35,7 @@
     <None Update="Platforms\Windows\Strings_WindowsThreading.resx" CustomToolNamespace="System.Reactive" Generator="ResXFileCodeGenerator" LastGenOutput="Strings_WindowsThreading.Designer.cs" />
   </ItemGroup>
 
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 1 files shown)


Rx.NET/Source/src/System.Reactive/System.Reactive.csproj
```diff
@@ -35,18 +35,7 @@
     <None Update="Platforms\Windows\Strings_WindowsThreading.resx" CustomToolNamespace="System.Reactive" Generator="ResXFileCodeGenerator" LastGenOutput="Strings_WindowsThreading.Designer.cs" />
   </ItemGroup>
 
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="System.Runtime.InteropServices.WindowsRuntime" Version="4.3.0" />
-  </ItemGroup>
-
   <!-- UWP -->
-  <ItemGroup Condition=" '$(TargetFramework)' == 'uap10.0' ">
```
teo-tsirpanis, line 44: The item group's condition is never satisfied. And all these four packages are pretty old.

teo-tsirpanis:
Hmm, seems that the package was actually needed for some WinRT interop. I am closing it...

PR #1545, opened 2021-05-24
Implement ToImmutable*Async extension methods

bash:
Resolves #1540.

I copied most of the XML documentation and tests of the non-immutable equivalents (with adjustments of course).

b667720 contains a bit of a refactoring, but I'm not sure if it's a big improvement.

Working with `AsyncOverloads.tt` was a bit unpleasant, especially since the generated code has diverged from the generated code (nullability and XML documentation) which meant I had to be careful of what to commit and what to discard.
Wouldn't a source generator be a better fit, because it knows about nullability and generic constraints?

PR #1575, opened 2021-07-07
Async overloads source generator

bash:
I took a crack at writing a source generator to replace the `AsyncOverloads.t4` template. This is a follow up to my PR #1545.

## Details
The source generator looks for methods annotated with the `[GenerateAsyncOverload]` attribute.
It generates a public method with the same signature, return type, generic parameter constraints, etc. that calls the annotated method.
The generator even copies the XML documentation to the public method. 🤓

## API Diff
I used [Microsoft.DotNet.GenAPI] (the tool generates a C# file representing the API of an assembly) to ensure that I didn't accidentally break the API surface.

A pre-generated diff is available [in this gist]. The only thing that changed is that the `comparer` parameter in the generated methods is now marked as nullable too which is a good change I think?

<details>
<summary>Instructions for verifying the API diff</summary>

1. Install the [Microsoft.DotNet.GenAPI] package (this requires adding a [custom NuGet feed])
   Here's a [.patch file] that installs Microsoft.DotNet.GenAPI (can be applied using `git apply <patch-file>`).
2. Switch to the `main` branch
3. Generate an API snapshot before the changes: \
    `dotnet build /p:GenerateReferenceAssemblySource=true /p:GenAPITargetPath=System.Linq.Async-before --framework netcoreapp3.1 --no-incremental Ix.NET/Source/System.Linq.Async`
4. Switch to this PR's branch
5. Generate an API snapshot after the changes: \
    `dotnet build /p:GenerateReferenceAssemblySource=true /p:GenAPITargetPath=System.Linq.Async-after --framework netcoreapp3.1 --no-incremental Ix.NET/Source/System.Linq.Async`
6. Compare the changes: \
    `git diff --no-index -- Ix.NET/Source/System.Linq.Async/System.Linq.Async-before Ix.NET/Source/System.Linq.Async/System.Linq.Async-after`

</details>

## Unrelated changes
I found that having a [solution filter] that only opens projects needed for System.Linq.Async to be really helpful while working on System.Linq.Async. I can revert the commit ba8ba26 if you don't want the solution filter checked into the repo.

[in this gist]: https://gist.github.com/bash/0a8f66b0a2595099b97afc5fe31d3c4a#file-api-diff-patch
[Microsoft.DotNet.GenAPI]: https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.GenAPI
[custom NuGet feed]: https://github.com/dotnet/arcade#getting-started
 [.patch file]: https://gist.github.com/bash/0a8f66b0a2595099b97afc5fe31d3c4a#file-add-microsoft-dotnet-genapi-patch
[solution filter]: https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019


bartdesmet:
Thanks for your contribution. This simplifies and modernizes the code significantly and sets a good example for future source generators in this space (e.g. the generation of expression tree based "queryable" variants of operators, which is a bit more involved).

PR #1630, opened 2021-10-30
exposing SingleAssignmentDisposableValue and making it ICancelable

Fijo:
When implementing various things outside of this library like a custom IScheduler or a custom rx operator the _SingleAssignmentDisposableValue_ can be quite useful. Since it is already well documented and it seams to me like it can't be used wrong, I'd like to make it public and make implement _ICancelable_ also.
Would that be ok with you?

Also please let me know if I have to bump versions manually.


idg10:
Are you sure we should make it implement `ICancelable`? As far as I can tell, every endpoint that accepts an implementation of `ICancelable` would end up boxing this value, defeating the point of using a value type.

If there were cases where generic methods imposed a constraint, it might be different, but Rx doesn't appear to do that anywhere. So I'm wondering if the reason this type was defined as not implementing `ICancelable` was to make it less likely that people would accidentally box it by passing it to methods that accept an `ICancelable`.

Fijo:
I agree with your argument. I'll refrain from making the SingleAssignmentDisposableValue implement ICancelable.

idg10 (approved):

PR #1643, opened 2021-11-13
reduced work for the garbage collector

Fijo:
Purpose: Minor garbage collector CPU time reduction.

This is just a minor enhancement as I noticed it's quite easy to make the garbage collectors job a little easier with quite a minor change in the framework. The code should be quite self explanatory too.


cabauman:
Because of how this class/struct is used by GetForwarder (return type is IObserver rather than a generic constrained by IObserver), this would result in boxing.

Fijo:
Thanks for the hint! I didn't know about the unique behaviour regarding boxing when it comes to structs but I read up on it now. As you pointed out it results in an additional boxing operation which negates the benefit hence why I'll retract the proposed change now.

PR #1782, opened 2022-06-23
Fix IAsyncEnumerable SkipLast for count = 0

thibault-reigner:
#### Bugfix

Fixes bug where calling `SkipLast(0)` over a custom sequence (not being a `AsyncIteratorBase` :https://github.com/dotnet/reactive/blob/85f1eb7c53e27cccdbeee3e0b044916168843fcc/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SkipLast.cs#L36)
would result in `Core`calling `Dequeue` over an empty queue, as `queue.Count == count` (both value being equal to 0)
https://github.com/dotnet/reactive/blob/85f1eb7c53e27cccdbeee3e0b044916168843fcc/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SkipLast.cs#L58

As a fix, I rewrote the `Core` method using `await foreach` and matching its synchronous implementation for `IEnumerable` **https://github.com/dotnet/reactive/blob/93386a90d9e7a78c2a0c3aaa16d31e1328f71b72/Ix.NET/Source/System.Interactive/System/Linq/Operators/SkipLast.cs#L33

I have also edited the unit test `SkipLast_Zero_NoAlias` so this one fails with current version of the code.

### DIFF
--- a/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/SkipLast.cs
+++ b/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/SkipLast.cs
@@ -1,6 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT License.
-// See the LICENSE file in the project root for more information. 
+// See the LICENSE file in the project root for more information.
 
 using System;
 using System.Collections.Generic;
@@ -64,12 +64,13 @@ namespace Tests
         }
 
         [Fact]
-        public void SkipLast_Zero_NoAlias()
+        public async Task SkipLast_Zero_NoAlias()
         {
             var xs = Xs();
             var ys = xs.SkipLast(0);
 
             Assert.NotSame(xs, ys);
+            await NoNextAsync(ys.GetAsyncEnumerator());
         }
 
         private async IAsyncEnumerable<int> Xs()
--- a/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/SkipLast.cs
+++ b/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/SkipLast.cs
@@ -70,7 +70,8 @@ namespace Tests
             var ys = xs.SkipLast(0);
 
             Assert.NotSame(xs, ys);
-            await NoNextAsync(ys.GetAsyncEnumerator());
+            var e = ys.GetAsyncEnumerator();
+            await HasNextAsync(e, 1);
         }
 
         private async IAsyncEnumerable<int> Xs()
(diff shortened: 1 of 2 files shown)


thibault-reigner:
> Could this, instead, throw early for invalid values of `count` resp. just return the identity enumerable for `count == 0` ?

If I am not mistaken this is already handled in the `SkipLast` method itself, `Core` only being used if the value of `count` is valid, or there is a need to really return a new enumerable as stated in the comment https://github.com/dotnet/reactive/blob/ff6e92a87d3c514de2f322632de5936f5f0dd565/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/SkipLast.cs#L34-L35 which the test https://github.com/dotnet/reactive/blob/ff6e92a87d3c514de2f322632de5936f5f0dd565/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/SkipLast.cs#L67
also enforces.

danielcweber:
I see...so there is a case in which we have to return `source` unchanged but have to 'hide' the object identity. I wasn't aware of that.

I am thinking about not touching the implementation of SkipLast, but instead having a generic way to wrap `source` and keep its identity. `AsAsyncEnumerable` comes to mind, but that'll only return the object itself.

Is there more cases like this one, where `Core` is called only to hide `source`? If so, could you provide a more generic solution, like e.g. a `Wrap` method? I think that would be the cleanest solution because we wouldn't touch any implementations and correctness of the changes would be immediately obvious.

In any case, good find!

thibault-reigner:
> I see...so there is a case in which we have to return `source` unchanged but have to 'hide' the object identity. I wasn't aware of that.
> 
> I am thinking about not touching the implementation of SkipLast, but instead having a generic way to wrap `source` but keep its behaviour. `AsAsyncEnumerable` comes to mind, but that'll only return the object itself.
> 
> Is there more cases like this one, where `Core` is called only to hide `source`? If so, could you provide a more generic solution, like e.g. a `Wrap` method? I think that would be the cleanest solution because we wouldn't touch any implementations and correctness of the changes would be immediately obvious.
> 
> In any case, good find!



> I see...so there is a case in which we have to return `source` unchanged but have to 'hide' the object identity. I wasn't aware of that.
> 
> I am thinking about not touching the implementation of SkipLast, but instead having a generic way to wrap `source` but keep its behaviour. `AsAsyncEnumerable` comes to mind, but that'll only return the object itself.
> 
> Is there more cases like this one, where `Core` is called only to hide `source`? If so, could you provide a more generic solution, like e.g. a `Wrap` method? I think that would be the cleanest solution because we wouldn't touch any implementations and correctness of the changes would be immediately obvious.
> 
> In any case, good find!

This seems to be a bit out of scope of this PR and quite some work, as it requires to search for such pattern in all extensions methods of `IAsyncEnumerable`.

Also, I noticed that there is an extensive use of `GetAsyncEnumerator() + MoveNextAsync()` instead of `await foreach` in this library, most likely due to the fact it predates the introduction of `IAsyncEnumerable` and C# 8 `await foreach`.
It led to extra complexity, and here bug, with no easy way to compare them with their synchronous variations (for `IEnumerable`).

thibault-reigner:
@danielcweber Requested changes were done in https://github.com/dotnet/reactive/pull/1782/commits/e818952bbf232547e35dd58dfa623d3f703cba43

PR #1787, opened 2022-07-05
Ix: Correct .Scan() to use matching behavior

### LINKED ISSUE #1672: Inconsistent behavior of the Scan operator
The implementation of the `Scan` operator differs between Ix (both sync and async) and Rx with regards to the first element.

The source code contains a comment explaining the behavior for Ix stating it's compatible with Rx (but that's not actually the case):
```
// NB: Implementations of Scan never yield the first element, unlike the behavior of Aggregate on a sequence with one
//     element, which returns the first element (or the seed if given an empty sequence). This is compatible with Rx
//     but one could argue whether it was the right default.
```

This test code demonstrates the issue:
```
var ob = Enumerable.Range(0, 3).ToObservable();
var ae = Enumerable.Range(0, 3).ToAsyncEnumerable();

var obResult = await ob.Scan((acc, i) => i).ToList(); // 0, 1, 2
var aeResult = await ae.Scan((acc, i) => i).ToListAsync(); // 1, 2

var isSame = aeResult.SequenceEqual(obResult); // false
```

Furthermore, the implementation of `Scan` in MoreLINQ yield the first element, adding to the confusion when switching between libraries.

I understand that it's less than ideal to change the behavior of existing operators, but in this case I think it's worth considering due to the current inconsistency.

I would be happy to make a PR if you agree it should be changed.

viceroypenguin:
The scan operator in most languages includes the seed or first element. This commit updates the Scan operator in System.Interactive to match both Rx.NET and other languages.

Fixes #1672

### DIFF
--- a/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Scan.cs
+++ b/Ix.NET/Source/System.Interactive.Async.Tests/System/Linq/Operators/Scan.cs
@@ -28,6 +28,7 @@ namespace Tests
             var xs = new[] { 1, 2, 3 }.ToAsyncEnumerable().Scan(8, (x, y) => x + y);
 
             var e = xs.GetAsyncEnumerator();
+            await HasNextAsync(e, 8);
             await HasNextAsync(e, 9);
             await HasNextAsync(e, 11);
             await HasNextAsync(e, 14);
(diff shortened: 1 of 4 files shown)


akarnokd:
Hi. This is a breaking change on an established operator. Imagine, suddenly everyone would get an extra item even though they were fine with not having that first item.

viceroypenguin:
Yes, it is a breaking change. However, it corrects the behavior to match every other implementation of the scan operator (both Rx as well as F#, Haskell, etc.), so it restores expectation of how the operator will work. Breaking change means release should only happen on Major Version (v7).

viceroypenguin:
Alternatives:
a) `[Obsolete]` the original version with an explanation, but leave it as is, and publish the corrected version under a different name; or
b) Update the documentation to be explicit about the fact that it is an incorrect implementation of the operator, according to common interpretation of scan behavior.

viceroypenguin:
@akarnokd Any thoughts on the three paths forwards?

akarnokd:
Implement the desired operator and behavior in your own library.

Issue #1726, opened 2022-03-10
CancellationToken in ToAsyncEnumerable can cause NullReferenceException


slavashar:
#### Bug

It look like there is a race condition in `ToAsyncEnumerable` which can cause `NullReferenceException` when the cancellation token is requested.

> Which subcomponent library (Ix, Async.Ix)?

Ix

> Which library version?

System.Linq.Async 6.0.1

> What are the platform(s), environment(s) and related component version(s)?

Windows, .NET6

> What is the use case or problem?

Cancellation of `ObservableAsyncEnumerable`.

> What is the expected outcome?

`OperationCanceledException` is thrown.

> What is the actual outcome?

`NullReferenceException` is thrown intermittently.

> What is the stacktrace of the exception(s) if any?

```
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Linq.AsyncEnumerable.ObservableAsyncEnumerable`1.MoveNextCore() in /_/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.Observable.cs:line 88
   at System.Linq.AsyncIteratorBase`1.MoveNextAsync() in /_/Ix.NET/Source/System.Linq.Async/System/Linq/AsyncIterator.cs:line 70
   at System.Linq.AsyncIteratorBase`1.MoveNextAsync() in /_/Ix.NET/Source/System.Linq.Async/System/Linq/AsyncIterator.cs:line 75
   at Program.<<Main>$>g__ConsumeAsync|0_2(Subject`1 subject, CancellationToken ct) in C:\Users\shara\source\repos\ToAsyncEnumerableTest\Program.cs:line 30
```

> Do you have a code snippet or project that reproduces the problem?

```
using System.Reactive.Subjects;

Parallel.For(1, 10, _ => RunAsync().Wait());

async Task RunAsync()
{
    // Set custom context to avoid sychronous execution of enumerable continuation
    SynchronizationContext.SetSynchronizationContext(new CustomSynchronizationContext());

    using var subject = new Subject<int>();
    using var cts = new CancellationTokenSource();

    var task = ConsumeAsync(subject, cts.Token);

    subject.OnNext(1);
    cts.Cancel();

    try
    {
        await task;
    }
    catch (OperationCanceledException)
    {
        // ignore
    }
}

async Task ConsumeAsync(Subject<int> subject, CancellationToken ct)
{
[...]

PR #1855, opened 2023-01-07
Fix a race condition in ToAsyncEnumerable on cancellation

slavashar:
Fixes https://github.com/dotnet/reactive/issues/1726

### DIFF
--- a/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.Observable.cs
+++ b/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToAsyncEnumerable.Observable.cs
@@ -173,7 +173,7 @@ namespace System.Linq
             {
                 var cancelledTcs = default(TaskCompletionSource<bool>);
 
-                Dispose();
+                DisposeSubscription();
 
                 while (true)
                 {


slavashar:
@idg10, can we proceed with the merge? It seems the issue is still there.

idg10:
Hi, sorry for how long this is taking.

I've got a couple of concerns:

* there are no tests around this bug
* I don't understand the thinking around why this is the fix
* is anything going to dispose the `_ctr` that was getting disposed back when it called `Dispose` instead of `DisposeSubscription`?
* presumably the call to `_values.TryDequeue` is still happening, it's just no longer failing because we're no longer nulling out the relevant fields, but is that OK? If feels like this change just masks a problem (which is that code is attempting to do things in this object after it should have been shut down)

Is there a reason for replacing the full `Dispose` with what seems to be a half-baked disposal instead of something like loading `_values` into a local variable and then checking it for null? Or trying to prevent the situation from occurring?

slavashar:
Hi @idg10,

It’s been a few years since I last looked at this, but I wanted to address some of your concerns:
- The issue is related to a race condition, which makes it challenging to create a reliable unit test without the risk of false positives. However, this functionality is already covered in the existing tests.
- Currently, on cancellation, ObservableAsyncEnumerable calls Dispose() on itself, which I found unconventional, especially since it handles some signaling logic afterward. This is problematic as the instance is still in use by the consumer (the enumerator pattern). Instead, we should unsubscribe and signal that it has completed.
- Dispose() should only be called by the consumer (the enumerator pattern) and not internally.
- The cancellation of the enumerable is controlled by _cancellationToken. If it is raised after the check on line 60, I believe we should allow the last item to be produced consistently before canceling the enumerable in the next cycle.
- Checking for null reliably without locking is tricky because three variables (_signal, _values, and _error) might be set or accessed by different threads.

Let me know if you have any further thoughts.

fedeAlterio:
I think there is another issue also. In the OnCancelled method it sets the signal to a canceled task completion source. But it's possible that in parallel another thread is iterating, and it's right after the .Configure await(false) (so it's just about to resume iterating). At this point it substitutes null to the signal and the iteration would not be cancelled. 

Maybe a fix could be using Interlocked.Exchange to set to null, checking if the old value was cancelled, and throwing OperationCancelled Exception if that is the case?

Issue #1628, opened 2021-10-30
Discussion: Changing SchedulerDefaults in a 3rd party lib to support Unity Engine


Fijo:
### Motivation:
I'm currently attempting to create a [lightweight extension](https://github.com/Cosmic-Shores/Rx.Unity) on top of _System.Reactive_ for use with the latest version 2021.2 of the Unity Game engine. The existing adoption of _System.Reactive_ for Unity called [UniRx](https://github.com/neuecc/UniRx) has lacked maintenance in a big way for years hence a growing community interest in the ability to benefit from the improvements made to _System.Reactive_ in recent years.
Sources:
* [Discussion: UniRx vs .NET Rx(with Unity integrate port from UniRx](https://github.com/neuecc/UniRx/issues/324)
* [Discussion: commnutity driven development](https://github.com/neuecc/UniRx/issues/325)
* [Open PRs - old ones first](https://github.com/neuecc/UniRx/pulls?q=is%3Apr+is%3Aopen+sort%3Acreated-asc)

_UniRx_ is a _System.Reactive_ fork and that's why it's maintenance is quite a bit of work. Also it makes it so you can't just consume a library that depends on _System.Reactive_ when using _UniRx_ inside unity. (This alone has lead to some [creative solutions/ points of pain](https://ecsrx.gitbook.io/project/other/microrx) in the past)

To make _System.Reactive_ work well within Unity one key thing has to be changed though and that's unfortunately currently not possible as far I was able to tell from the research I did. It's also something that probably everyone switching over from _UniRx_  will rely on to make a switch without having to add unreasonable 'hacks' all over their project.
I'm talking about being able to change the deafult scheduler behaviour to accommodate for unities unique multithreading considerations.
Thus it's required for TimeBasedOperations to be scheduled on the main unity thread.
Also if a build is for WebGL multithreading support isn't a thing altogether hence why all AsyncConversions need to also be scheduled on the main unity thread.

From my reading it looks to me like in the past it was possible to make these behaviour changes from a third party library by implementing an _IPlatformEnlightenmentProvider_. However as of version 5 of System.Reactive it looks like it's no longer possible to influence the behaviour sufficiently when using that method. If feels like that concept has been partially retired.

As an alternative the easiest way that I'd at least have the option (though it wouldn't be entirely clean yet) would be to change SchedulerDefaults.cs to look like this:
```cs
namespace System.Reactive.Concurrency
{
    internal static class SchedulerDefaults
    {
[...]

PR #1879, opened 2023-03-01
Add Rx.NET 2023 roadmap, and tooling update ADR

### LINKED ISSUE #1880: [New Maintainers] Publish a Roadmap for 2023
As part of the plan described in https://github.com/dotnet/reactive/discussions/1868 we need to define a roadmap for Rx.NET development for 2023.

idg10:
As https://github.com/dotnet/reactive/discussions/1868 states, one of the steps in bringing Rx.NET 'back to life' is to "Publish a Roadmap for 2023 (covering the .NET 8.0 LTS release train) & beyond".

This PR addresses that by setting out a road map. It also provides an Architectural Decision Record (ADR) describing the first proposed step (updating the Rx.NET) in more detail, along with all the context that informs that part of our proposals.

Resolves #1880


Rx.NET/Documentation/Rx-Roadmap-2023.md
```diff
@@ -0,0 +1,193 @@
...
+
+This caused problems for libraries that need to do things with threads, with Rx's schedulers being a prime example. What's the `NewThreadScheduler` supposed to do on a platform that doesn't allow applications to create new threads directly? There are workarounds, such as using the thread pool, but those workarounds are suboptimal if you're running on unrestricted platforms. This leads to a choice between using an unsatisfactory lowest common denominator, or somehow arranging for different behaviour on UWP vs other platforms.
```
jkotas, line 94: This was true in the early versions of UWP tooling. 

The current UWP tooling supports netstandard2.0 that lets you create new threads just fine.

In general, it is not really worth it to pay attention to anything that does not support netstandard2.0.

danielcweber:
>I don't know to what extent any of Rx's former contributors continue to have any interest in this project

I'm planning to adopt Rx.Async in some form once it gets a release and would happily pick up contributing to this repo again (bug fixes, etc.). I still have write access to this repo from my batch of work back in 2018/9.

Rx.NET/Documentation/Rx-Roadmap-2023.md
```diff
@@ -0,0 +1,193 @@
...
+
+A more detailed description of how this first step would be achieved can be found in [ADR 0001-net7.0-era-tooling-update](./adr/0001-net7.0-era-tooling-update.md).
+
+### Rx.Async
+
+We want to make preview builds of Rx.Async available as soon as possible. We don't know of any technical reason to tie this to the release described in the preceding section. In the long run we expect that we would want Rx.NET and Rx.NET.Async releases to be synchronized. (So each new Rx.NET release would simply include packages with the async version.) But since Rx.Async is currently in a preview state, we don't need to unify things yet—it would be better to make `alpha` builds available as soon as possible. So the work to do this could proceed concurrently with the work described in the preceding section.
```
NickDarvey, line 169: > In the long run we expect that we would want Rx.NET and Rx.NET.Async releases to be synchronized. (So each new Rx.NET release would simply include packages with the async version.)

Rx.NET.Async probably deserves it’s own version numbers because, if the implementation of Rx.NET Async remains mostly seperate, then even in the long run it’ll change independently of Rx.NET.

Rx.NET/Documentation/Rx-Roadmap-2023.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Why UWP in 2023?
+
+Over two years have passed since Rx 5.0 was released. Microsoft actively discourages the use of UWP, so why do we still target it? One obvious answer is that there has been no new release of Rx in the intervening time. However, it's not as simple as that.
+
+Although Microsoft pushes people strongly in the direction of WinUI v3, the fact is that if you want to write a .NET Windows Store application today, certain Windows features are only available to you if you target UWP. For example, full access to the camera (including direct access to the raw pixels from captured images) is currently (February 2023) only possible with UWP.
```
Trinitek, line 124: I'm curious, which camera API specifically? CameraCaptureUI?

Rx.NET/Documentation/Rx-Roadmap-2023.md
idg10: It's a good point. When I wrote that, I was thinking about major releases—I was anticipating the major driving factor behind new updates being that we would need to test against new frameworks, drop out-of-support TFMs, update tooling, add support for newly available functionality (e.g. trimming) and so on.

So I'm thinking that the revised position should be something like: "In the long run, we expect that releases of new major versions are likely to be synchronized because they are typically made necessary by older .NET runtimes going out of support. However, minor releases and bug fixes for Rx.NET and Rx.NET.Async will be handled by separate releases, since although they share test code, their implementations are separate."

Rx.NET/Documentation/adr/0001-net7.0-era-tooling-update.md
```diff
@@ -0,0 +1,140 @@
...
+These changes will have the following positive effects:
+
+1. it will be possible to download the repository and, using current tooling (VS 2022 and .NET 7.0 SDK), open, build, and test `System.Reactive.sln` without errors, warnings or other diagnostic messages
+2. automated builds will now be running tests on .NET 6.0 and .NET 7.0 (in addition to .NET Framework)
+3. when we release components based on these changes to NuGet, they will no longer target frameworks that are out of support, and will include `net6.0` targets
+4. if setting `<IsTrimmable>true</IsTrimmable>` turns out to be all that is required for trimmability, this will address that requirement
+
+It will impose the following new requirements for anyone wanting to build the solution:
+
+1. the repository must be cloned into a location where the path length is no longer than 25 characters
```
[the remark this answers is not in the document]
idg10: No. 25 characters. It really is that bad.

The total windows path limit is 260. And the astoundingly long filenames used by this component mean that by the time it's taken its share, and you've included the other directory hierarchy in the Rx repo, you're left with just 25 characters for wherever you want to locate the repo.

Rx.NET/Documentation/Rx-Roadmap-2023.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+## Plans for upcoming release
+
+We need to decide whether we should try to solve all of the problems in one step, or whether we produce any interim releases that get us part way to a solution. Our current position is that we will not attempt to do everything at once. We would like to get automated tests running against .NET 6.0 and .NET 7.0 in place as soon as possible so that we can verify that these platforms are supported. (This necessarily entails fixing the problems that prevent Rx from being built using the latest tool versions.) And we would like to make preview builds of AsyncRx.NET available as quickly as possible.
```
noseratio, line 138: Releasing of AsyncRx.NET (`IAsyncObservable` etc) would be very heartily welcomed by some of us 🙏

E.g., MS Orleans folks have had to come up with their [own version of it](https://twitter.com/reubenbond/status/751963797179371521?s=20).

HowardvanRooijen:
See https://github.com/dotnet/reactive/pull/1882

Rx.NET/Documentation/Rx-Roadmap-2023.md
idg10: I was using `CaptureElement` in XAML and a whole bunch of types from `Windows.Media.Capture`.

Rx.NET/Documentation/Rx-Roadmap-2023.md
idg10: This is an important goal for us, which is why we're working to make packages available very soon. Right now we're gradually working out which parts of the build and deployment mechanisms need attention as a result of having languished for a couple of years, and which are still working just fine, but should be working out what steps will be required for us to start publishing the async packages.

HowardvanRooijen:
See PR https://github.com/dotnet/reactive/pull/1883

Fijo:
# About platform independent rx, test-sharing and unity

I just wanted to leave some information here in regards to IL2CPP compatibility and testing with Unity. In an attempt to reuse as much of the test code from this repository as possible and I've created a pretty unorthodox automation toolchain as part of the [Rx.Unity](https://github.com/Cosmic-Shores/Rx.Unity) repo.

The main reason I am making this post is because I saw you, talking about sharing tests so adoptions for different platforms (like UWP) can keep running them after they've been moved out of the main repository. Thus I thought I'd share my approach to give you some insight, inspiration and in hopes for a good solution that'll be easily maintainable for everyone.

Before I forget thank you so much for the work you've been putting in to bring Rx.Net back on track. I really appreciate it.

While that (Rx.Unity) needs updating again, which I am still committed to getting back to.
I'm also planning to do some more work on there once it's possible to use it along with the official System.Reactive NuGet Packages.

Why did I have to run all the tests?
Mostly to figure out what works with IL2CPP and in Unity in general.

How the build works (only focusing on what's relevant for Rx.net):
1. I clone the Rx.Net sources into a folder.
2. Patching the sources (described below).
3. Custom System.Reactive Build (described below).
4. Sideloading partial classes from tests (described below).
5. Including migrated UniRx-Tests to verify support for people migrating
6. Building and running tests in Unity (described below).

## Patching the sources
Build on [gulp](https://github.com/gulpjs/gulp) I've [scripted out a few tasks](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Dependencies/patcher/gulpfile.js) to transform the tests from the [dotnet/reactive](https://github.com/dotnet/reactive/tree/rxnet-v5.0.0) repo so they run within Unity.

Here are some of the things being done to rx sources themselves:

### Patch System.Reactive.csproj
- TargetFramework => netstandard2.0
- LangVersion => 9
- DefineConstants => NO_NULLABLE_ATTRIBUTES
- SignAssembly => false

### Patch './Rx.NET/Source/src/System.Reactive/TaskObservable.cs'
- Removing the `[AsyncMethodBuilder(typeof(TaskObservableMethodBuilder<>))]`-Attribute (Apparently this was unsupported / caused issues - I'll have to look into that again to find out more.)

### Patch System.Reactive - Entire File replacements
I replace the following files with different ones or my own (Links lead to the replaced file):
- [./Rx.NET/Source/src/System.Reactive/Directory.build.props](https://github.com/dotnet/reactive/blob/rxnet-v5.0.0/Rx.NET/Source/Directory.build.props)
- [./Rx.NET/Source/src/System.Reactive/Properties/InternalsVisibleTo.cs](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Dependencies/Patches/InternalsVisibleTo.cs)
- .[/Rx.NET/Source/src/System.Reactive/Concurrency/SchedulerDefaults.cs](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Dependencies/Patches/SchedulerDefaults.cs) (no longer relevant once [this issue](https://github.com/dotnet/reactive/issues/1628) is addressed)
- [./Rx.NET/Source/src/global.json](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Dependencies/Patches/global.json) (just tooling stuff; workaround character / subject to change)

### Patch XUnit to NUnit
Unitys tooling, as well as the [runtime test runner](https://github.com/Cysharp/RuntimeUnitTestToolkit) I'm using only really works with NUnit. Adding too many additional dependencies is kind of painful (it would propobly work for assertations and I think I'll look if that's an option for when I'm adapting the now MS Test based sources.). Luckily a few replacements were enough to cover everything being used within the tests. Asyncronous Tests the trickiest here.
For more [see xunit2nunitTransformer-Method](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Dependencies/patcher/gulpfile.js#L86).

### Patches on Tests
- `ThreadPoolScheduler.Instance` is replaced with `Rx.Unity.Concurrency.ThreadPoolOnlyScheduler.Instance` in most test sources, as that's generally preferred when using `Rx.Unity`.
- `public class` is replaced with `public partial class` in most test sources, to allow additional setup / teardown logic to be provided by additional files.
- Various Tests are blacklisted by either making those methods private (via search and replace) or by excluding those files. The goal is ofc to get as much as possible to work in Unity but certain things just aren't there yet or would be a lot more difficult to troubleshoot. Alot of them cause the Tests to end up hanging up as well.

## Custom System.Reactive Build
Building is done with `dotnet build` in `./Rx.NET/Source/src/System.Reactive`.

=> Once the whole `SchedulerDefaults.cs` thing is fixed I think the whole `InternalsVisibleTo.cs` required for the tests is the only real thing preventing me from just using the NuGet Package of `System.Reactive`.

I'm only using the `netstandard2.0` build for now. The [latest version of Unity supports](https://docs.unity3d.com/2022.2/Documentation/Manual/dotnetProfileSupport.html) up to `netstandard2.1`, today. They are however working on a [https://forum.unity.com/threads/net-6-support.1059992/](full move to dotnet6) AFAIK.

## Sideloading partial classes from tests
Alot of the test classes are being made partial and they are sideloaded with [a custom SetUp/ TearDown-Methods specified in additional sources](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Assets/Tests/Tests.System.Reactive.Custom/Tests/TaskObservableExtensionsTest.cs).
The main purpose of this is generally to switch the static instances within `SchedulerDefaults` to the [more Unity compatible ones](https://github.com/dotnet/reactive/issues/1628#issue-1040212133) and to change them back on teardown.
In a few instances [additional static methods are also provided](https://github.com/Cosmic-Shores/Rx.Unity/blob/master/Assets/Tests/Tests.System.Reactive.Custom/Tests/Linq/Observable/BufferTest.cs) (which are never actually used) but provide a workaround to force the AOT Linker to generate certain generic classes with the required specific type parameters as in rare cases it fails to make correct predictions. Having all the generic type combination being generated used to be a requirement for IL2CPP to work. Recently Unity have added support for generic type creation at runtime in IL2CPP but it is slow.

## Building and running tests in Unity
Only test code and the source for Rx.Unity is build within Unity itself but not System.Reactive itself. The required files are copied into the appropriate folder by my script.

Unity tests are built and run in a via [github actions](https://github.com/Cosmic-Shores/Rx.Unity/actions/runs/1485007135). To ensure correct behaviour with IL2CPP an actual game is built [including some framework that enables just running tests](https://github.com/Cysharp/RuntimeUnitTestToolkit) in that game. Then those tests are run via console. For Mac, iOS, Android and WebGL this isn't possible (or in the case of Mac I didn't get it to run that - I don't have a Mac so I couldn't troubleshoot it easily). Here an artefact with the build and a limited expiration is generated which can be manually downloaded, installed and run to verify all tests are working.

@idg10

HowardvanRooijen:
@Fijo thanks very much for sharing that. We're very keen to enable better support for Unity, but it's an ecosystem we're unfamiliar with. If you're happy, we'll reach out to get your feedback when we get to those backlog items (most likely as part of v7 / v8)

Thaina:
@HowardvanRooijen Are there any progress about changing default scheduler or anything to support other platform like unity at all?

HowardvanRooijen:
Hi @Thaina 

There are a number of things in flight at the moment, that we're trying to land in time for the .NET 9 launch (and .NET 6.0 going out of support, as well as giving Ix some TLC #2135 #2131 #2102) and general tidy up #2113 #2092 #2134

There was a big discussion around packaging... which would solve some of the problems we're facing #2038 - but I don't think we got to a consensus.

We've been blocked for a couple of months as the .NET Foundation has mandated a move to Managed Identities for code signing auth. This is a great idea, but since Rx was the first project to attempt this, it took a while to work out how to get the config right (https://github.com/dotnet-foundation/projects/issues/387). Ian got that working this week (#2167), with the help of Chris Sfanos from the DNF, so progress is moving ahead again.

Ian has been looking at getting Async Rx into a better state:

- #1900 - specifically using Source Generators to transform the existing Rx test suite - https://github.com/dotnet/reactive/issues/1900#issuecomment-2325911446 
- #2148 -  https://github.com/dotnet/reactive/issues/2148#issuecomment-2373881870

We're hoping to do another yearly update either as part of .NET Conf or released independently at the same time.

Thaina:
Thank you for your response

HowardvanRooijen:
@Thaina - you're welcome! There's lots of stuff going on, to try and get all the bits ready for the next release. Hopefully we'll be able to make this a little clearer and more cohesive over the next month as we head towards the November milestone.

khellang:
Thanks for the hard work you're putting into this, @HowardvanRooijen and team ❤️

HowardvanRooijen:
@khellang it's all @idg10 hard work!

Thaina:
@HowardvanRooijen Are there any progress about changing default scheduler or anything to support other platform like unity at all?

Next month of September seem so long

HowardvanRooijen:
@idg10 is currently working on this and should have an update very soon (within the next week or so)

Thaina:
Thank you very much

Thaina:
> @idg10 is currently working on this and should have an update very soon (within the next week or so)

@HowardvanRooijen Next week seem like forever

HowardvanRooijen:
Are you keeping track of https://github.com/dotnet/reactive/pull/2268 ? 

Ian's current focus is getting the v7 release out with .NET 10 and recording a talk for .NET Conf 2025.

Thaina:
@HowardvanRooijen Last week I have ask @idg10 and he said like it was actually unrelated

https://github.com/dotnet/reactive/discussions/2211?sort=new#discussioncomment-14839559

image

HowardvanRooijen:
I think we're missing some useful context. For the avoidance of doubt, the v7 release is the thing we've been working towards for 2.5 years and that's to de-couple the platform specific dependencies, which should help solve a lot of these cross platform (Uno / Avalonia / Unity) problems. We're a couple of weeks away from this happening (.NET 10 release). Once this is done it will free up a HUGE amount of cognitive bandwidth, and will also clean up a number of open PRs which explored alternative approaches, which are cluttering up the repo.

I would, as Ian suggests create a new feature request / issue, so this isn't buried in a 2.5 year old complex issue,  and please be as detailed as possible, as Unity is not a workload we have any experience with. 

A sample project we can use to repro, with examples of the scheduler changes you propose would be very useful.

Thaina:
@HowardvanRooijen My expectation of this project is that, I don't actually want this project to have anything to do with unity at all. I just want it to be compatible with .NET Standard 2.0 or 2.1 and so it's dll or even the code itself could be imported into unity directly

And then the most important things is to have https://github.com/dotnet/reactive/issues/1628#issuecomment-1046056449. One way to remove all kind of dependency workload from reactive system is to allow us to specify default scheduler by ourselves for each platform we work with. So that I could always implement scheduler for each platform specifically

Thaina:
@HowardvanRooijen Actually are there any problem if these all `internal` would be changed into public

image

Or implement `IPlatformEnlightenmentProvider` like suggested here

https://github.com/dotnet/reactive/issues/1628#issuecomment-974138466

idg10:
I would very much be in support of enabling the broad goal of supporting platforms where use of timers comes with threading constraints that are different from the platforms where Rx.NET works today. (So much for "One .NET...")

In fact I'm pretty sure we also need to do this to get Rx.NET working properly on in-browser WASM. (The main blocker for supporting that is actually the inability to run tests in WASM; I haven't checked what blockers, if any, remain on that, with the latest MSTest framework. But I know that one of the problems that happens in practice is that use of certain timer-based operators immediately hits problems because Rx.NET presumes it can use threads in particular ways, and those assumptions fail in browser WASM. Or at least they did last time I checked.)

However, I'm a bit reluctant to do this by simply making a bunch of internal things public. The limited digital archaeology I've done in the Rx.NET codebase on this particular issue suggests that this `SchedulerDefaults` replaced something else, because some traces of earlier ways of doing things remain. So I don't think that the `SchedulerDefaults` type necessarily represents a fully-resolved solution to this problem.

`IPlatformEnlightenmentProvider` might be closer to the right way, although its design philosophy is not an exact match: the model there (which was inspired by a feature of the same name in VMs) is not that the `IPlatformEnlightenmentProvider` is essential to correct functioning: the basic design assumption is that things will with without platform-specific enlightenments, but that a suitable platform-specific provider can do things in a better way because it knows things about the specific platform it's running on (that's what it means to be 'enlightened' here).

So if you don't have the right `IPlatformEnlightenmentProvider`, things still work, it's just that they work better if you do. That, at least, is the _original_ intent of these things as I understand it.

But here I think we're talking about basic correctness: without a proper understanding of how timers and threading interact, certain operators simply don't work at all on Unity or browser WASM.

I think this would need to be an Rx v8 feature: the big goal with Rx 7 is to fix the long-standing problems that arise from the Great Unification, and I don't want to change too much in one major release. That said, I'm also perfectly happy for a v8 to follow fairly quickly on the tails of v7... But I would want to get a non-preview v7 shipped before we fix this threading issue. (However, I also want to think about it during this current v7 time frame, because v7 is already quite disruptive, and if there are changes that will need to be made to accommodate this new feature, maybe we want to think about paving the way while we're breaking things... So I wonder if it might make sense to do some sort of experimental preview of this feature in v7.)

Thaina:
I would like to comment on one thing

> But here I think we're talking about basic correctness: without a proper understanding of how timers and threading interact, certain operators simply don't work at all on Unity or browser WASM.

No one could know in advance how any platform in future will design how timer and threading interact with each other. So I don't think we can

Instead I would like library like RX (all of RX include dotnet) should just be as abstract as possible. It should not even know if that platform has timer or threading. Only serve us to fabricate signal. So we can reuse RX in any platform in the future too

> So if you don't have the right IPlatformEnlightenmentProvider, things still work, it's just that they work better if you do. That, at least, is the original intent of these things as I understand it.

It's better to have us fix library to make it work when anything broken in some platform, instead of trying to have it work in all platform but if it not work we cannot make it work at all like it had happen in unity

So I think RX and similar library should not have any internal members at all

PR #1882, opened 2023-03-03
Upgrade to supported TFMs and SDK versions

idg10:
Also remove dependency on System.Runtime.InteropServices.WindowsRuntime in netstandard2.0 - this wouldn't work right on net6.0 or later because they use C#/WinRT for interop.

This enables use of current tooling such as Visual Studio 2022


azure-pipelines.rx.yml
```diff
@@ -33,10 +33,9 @@ stages:
 
     steps:
     - task: UseDotNet@2
-      displayName: Use .NET Core 5.0.x SDK
+      displayName: Use .NET Core 7.0.x SDK
```
HowardvanRooijen, line 36: GH isn't letting me comment on line 27, but I believe this needs to be changed `pool: vs2019-preview`

HowardvanRooijen (approved):
👌

PR #1883, opened 2023-03-06
Replace xUnit with mstest

idg10:
The TFM upgrades in #1882 were a necessary step in being able to build Rx on current tooling, but not the only one. This change addresses the fact that [xUnit is no longer able to run tests on UWP](https://github.com/xunit/devices.xunit/issues/171). MSTest supports this in Visual Studio 2022, so we are moving back to MSTest. (Rx used to used MSTest before, so this is a relatively straightforward change.)


idg10:
Note that a handful of tests stopped working because xUnit (somewhat controversially) set `SynchronizationContext.Current` to be non-null, and some tests were effectively relying on that. We need to investigate for each of these whether we should have versions of the tests that run with and without a synchronization context, since the behaviour is evidently dependent on that.

These are all currently labelled with a comment with "`idg10`" in the text in this branch.

HowardvanRooijen (approved):

PR #1896, opened 2023-03-15
Merge AsyncRx.NET assemblies

### LINKED ISSUE #1891: Reduce the number of AsyncRx.NET NuGet packages
AsyncRx.NET currently has several projects corresponding to areas of functionality which, in Rx.NET, were all merged into the single `System.Reactive` package back in Rx v4.0.

Although we think that 'great unification' may have gone too far, the problems we see there are all to do with the decision to include UI-framework-specific code in that unification. We don't see any problems with the decision to merge concurrency, disposables, interfaces, LINQ, and subjects into a single library.

Given that AsyncRx.NET doesn't currently have any UI-framework-specific features, there seems to be no good reason not to merge it all down to a single package, to be consistent both with where Rx is now, and where we are aiming for Rx to be in V7.0 once we've moved the UI-framework-specific pieces back out again.

idg10:
Resolves #1891

System.Reactive.Async.Concurrency, System.Reactive.Async.Core, System.Reactive.Async.Disposables, System.Reactive.Async.Interfaces, System.Reactive.Async.Linq and System.Reactive.Async.Bcl all merged into System.Reactive.Async

System.Reactive.Shared has been removed because almost everything it did was already in System.Reactive. So we now take a dependency on System.Reactive, and one extension method puts back the one thing that was missing.


AsyncRx.NET/System.Reactive.Async/System.Reactive.Async.csproj
```diff
@@ -2,17 +2,130 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
+
```
HowardvanRooijen, line 5: Do you want to handle standard configuration of csproj files (i.e. Endjin.RecommendedPractices.NuGet settings) in this PR or a different one to cover "productionisation"?

HowardvanRooijen (approved):

AsyncRx.NET/System.Reactive.Async/System.Reactive.Async.csproj
idg10: I'll leave that until later. It would be best to align with Rx.NET and, if possible, Ix.NET.

PR #1910, opened 2023-04-06
Allow opt-out from TaskScheduler.UnobservedExceptions

### LINKED ISSUE #1256: Observable.FromAsync produces unobserved task exceptions
#### Context
Recently I discovered an uncertain behavior in `Observable.FromAsync` method which causes unobserved exceptions to be thrown. 

When there is an active subscription and task throws an exception everything works as expected - exception is catched by RX and forwarded to `onError` handler. The problem happens when subscription is cancelled right before the task failure. Under the hood (inside `SlowTaskObservable`) RX cancels task continuation, `Exception` property is not accessed by anyone and when GC is triggered - task unobserved exception is thrown.

**My main concerns here are next:**
* This behavior is inconsistent from all other non-task cases. For example if exception is thrown from `Select` (or anywhere else) and if subscription is cancelled at this point - exception is not forwarded anywhere and is "silently" ignored.
* RX is responsible for the task creation and by default always accessed `Exception` property. But when subscription is cancelled behavior changes completely.
* It basically means there is no 100% reliable way to get rid of `TaskScheduler.UnobservedTaskException`. Any `Observable.FromAsync` usage can cause an unobserved exception. This part is critical in my view.

**How I expect it to work:** if RX creates a task and takes responsibility for error handling, it should do it fully from the beginning to the end. Behavior has to be consistent, no subscription means we are no longer interested in errors and we shouldn't get them anywhere.

idg10:
**NOTE:** we are more likely to go with #1914 instead now. A problem with the approach in this PR is that it creates additional opportunities for overload resolution ambiguity, so although this is a binary-compatible change, it might cause build errors when people upgrade.

Resolves #1256 

Various Rx operators and extension methods can wrap a `Task` as an `IObservable<T>`. Historically, if you unsubscribe before the task is finished and the task subsequently fails, the resulting exception pops out through `TaskScheduler.UnobservedException`.

There is a point of view that this is inconsistent with how we cope with exceptions in things like `Select` projection callbacks that occur after unsubscription - those are just swallowed.

This PR provides the ability to ask for these kinds of orphaned exceptions to be ignored when they occur in tasks, just like they are in other scenarios.

The default behaviour does not change because that would break applications depending on the `UnobservedException` event. This adds new overloads enabling opting out.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -189,3 +189,8 @@ nuget.exe
 
 # JetBrains Rider adds these
 .idea/
+
+# Local NCrunch settings
+*.v3.ncrunchproject
+*.v3.ncrunchsolution
+/Rx.NET/Source/.NCrunch_*/StoredText
--- a/Rx.NET/Source/src/System.Reactive/Linq/IQueryLanguage.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/IQueryLanguage.cs
@@ -193,27 +193,27 @@ namespace System.Reactive.Linq
         IObservable<TSource> Start<TSource>(Func<TSource> function);
         IObservable<TSource> Start<TSource>(Func<TSource> function, IScheduler scheduler);
 
-        IObservable<TSource> StartAsync<TSource>(Func<Task<TSource>> functionAsync);
-        IObservable<TSource> StartAsync<TSource>(Func<CancellationToken, Task<TSource>> functionAsync);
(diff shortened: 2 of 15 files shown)


quinmars:
It's a pity that the default setting is exactly what most people don't want, but I understand your sentiment about backwards compatibility.

idg10:
Went with #1914 instead

PR #1941, opened 2023-05-19
System.Linq.Async takes a dependency on Microsoft.Bcl.AsyncInterfaces on target frameworks where it should not

### LINKED ISSUE #1834: Remove dependency on Microsoft.Bcl.AsyncInterfaces for net6.0
#### Feature request

Considering the points above, please describe the feature or behavior you would like dotnet/reactive included:

> Which subcomponent library (Ix, Async.Ix)?

lx

> Which next library version (i.e., patch, minor or major)?

patch

Microsoft.Bcl.AsyncInterfaces

> How commonly is this feature needed (one project, several projects, company-wide, global)?

global

> Please describe the feature.

I noticed the dependency on `Microsoft.Bcl.AsyncInterfaces` for `net6.0` on NuGet. I found that odd since, from `Microsoft.Bcl.AsyncInterfaces`'s README (below):
> This package is not required starting with .NET Standard 2.1 and .NET Core 3.0.

So it makes sense to only explicitly include the dependency for the older frameworks.

Looking at the PR history, it looks like this was just auto version bump-ed to the current state.

I found an open PR to address this: #1719

>**NOTE**: I call issue a FR since it technically doesn't break anything to leave the dependency.

danielcweber:
#### Bugfix

A change in [a2410b2](https://github.com/dotnet/reactive/commit/a2410b2267abe193191f3894d243771ae4b126fd#diff-040f99bf5abf4bafb4e02f92a7b83c27b42785a4b692995a2c629cf9b5711b69L4) added net6.0, bumped to net48 and dropped netcoreapp3.1 for System.Linq.Async. As a consequence, the condition on which a reference to Microsoft.Bcl.AsyncInterfaces is taken proves no longer useful. This results in the package reference being present on net6.0 (and beyond) when it should not be.

This commit rewrites the condition.

Fixes #1834.

### DIFF
--- a/Ix.NET/Source/System.Linq.Async/System.Linq.Async.csproj
+++ b/Ix.NET/Source/System.Linq.Async/System.Linq.Async.csproj
@@ -26,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Condition="'$(TargetFramework)' != 'netcoreapp3.1' and '$(TargetFramework)' != 'netstandard2.1' " Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
+    <PackageReference Condition="'$(TargetFramework)' == 'net48' or '$(TargetFramework)' == 'netstandard2.0' " Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
     <ReferenceAssemblyProjectReference Include="..\refs\System.Linq.Async.Ref\System.Linq.Async.Ref.csproj" />
     <ProjectReference Include="..\System.Linq.Async.SourceGenerator\System.Linq.Async.SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" Private="false" />
   </ItemGroup>


Ix.NET/Source/System.Linq.Async/System.Linq.Async.csproj
```diff
@@ -26,7 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Condition="'$(TargetFramework)' != 'netcoreapp3.1' and '$(TargetFramework)' != 'netstandard2.1' " Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
+    <PackageReference Condition="'$(TargetFramework)' == 'net48' or '$(TargetFramework)' == 'netstandard2.0' " Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
```
saul, line 29: If we only want to include this package reference on non-.NET Core targets, the idiomatic condition would be:

```suggestion
    <PackageReference Condition=" '$(TargetFrameworkIdentifier)' != '.NETCoreApp' " Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
```

Note that net5.0, net6.0, etc fall under `.NETCoreApp` too.

See some examples of this condition in [the dotnet/runtime repo](https://github.com/search?q=repo%3Adotnet%2Fruntime%20TargetFrameworkIdentifier%20path%3A*.csproj&type=code).

danielcweber:
Just realized #1719 also fixes the issue. Closing in favor of #1719.

danielcweber:
Fixed by #1719.

PR #2039, opened 2023-11-14
Run tests on .NET 8.0

idg10:
Added `net8.0` and `net8.0-windows10.0.19041` targets to test project.

Modified `Directory.build.targets` to recognize .NET 8.0 targets.

One test needed to be lightly modified because the way in which it was trying to test what it wanted to test doesn't quite work in .NET 8.0.


mwadams (approved):
This removes test support for dotnet7.0?

HowardvanRooijen (approved):

idg10:
@mwadams wrote:
>This removes test support for dotnet7.0?

It's certainly not supposed to! I had a quick look through the PR to double check, and as far as I can tell this just adds .NET 8.0 in the test projects.

I was planning to remove .NET 7.0 from the tests once .NET 7.0 goes out of support, but not before.

PR #2084, opened 2024-02-12
Move UI and platform types out of System.Reactive

### LINKED ISSUE #1745: Adding System.Reactive to WinUI3/MAUI app increases package size by 44MB because it includes WPF and WinForms references
#### Bug

> Which library version?
System.Reactive 5.0.0

WinUI3, MAUI

> What is the use case or problem?
Using System.Reactive in WinUI3 app

> What is the expected outcome?
Only required references for the platform are included in app package

> What is the actual outcome?
The app package increases from 44MB to 88MB with the following additional dlls in the package:
```
"Accessibility.dll"
"D3DCompiler_47_cor3.dll"
"DirectWriteForwarder.dll"
"Microsoft.VisualBasic.Forms.dll"
"Microsoft.Win32.Registry.AccessControl.dll"
"Microsoft.Win32.SystemEvents.dll"
"PenImc_cor3.dll"
"PresentationCore.dll"
"PresentationFramework.Aero.dll"
"PresentationFramework.Aero2.dll"
"PresentationFramework.AeroLite.dll"
"PresentationFramework.Classic.dll"
"PresentationFramework.dll"
"PresentationFramework.Luna.dll"
"PresentationFramework.Royale.dll"
"PresentationFramework-SystemCore.dll"
"PresentationFramework-SystemData.dll"
"PresentationFramework-SystemDrawing.dll"
"PresentationFramework-SystemXml.dll"
"PresentationFramework-SystemXmlLinq.dll"
"PresentationNative_cor3.dll"
"PresentationUI.dll"
"ReachFramework.dll"
"System.CodeDom.dll"
"System.Configuration.ConfigurationManager.dll"
"System.Design.dll"
"System.Diagnostics.EventLog.dll"
"System.Diagnostics.PerformanceCounter.dll"
"System.DirectoryServices.dll"
"System.Drawing.Common.dll"
"System.Drawing.Design.dll"
"System.IO.Packaging.dll"
"System.Printing.dll"
"System.Reactive.dll"
"System.Resources.Extensions.dll"
"System.Security.Cryptography.Pkcs.dll"
"System.Security.Cryptography.ProtectedData.dll"
"System.Security.Cryptography.Xml.dll"
"System.Security.Permissions.dll"
"System.Threading.AccessControl.dll"
"System.Windows.Controls.Ribbon.dll"
"System.Windows.Extensions.dll"
"System.Windows.Forms.Design.dll"
"System.Windows.Forms.Design.Editors.dll"
"System.Windows.Forms.dll"
"System.Windows.Forms.Primitives.dll"
"System.Windows.Input.Manipulations.dll"
"System.Windows.Presentation.dll"
"System.Xaml.dll"
"UIAutomationClient.dll"
"UIAutomationClientSideProviders.dll"
"UIAutomationProvider.dll"
"UIAutomationTypes.dll"
"vcruntime140_cor3.dll"
"WindowsFormsIntegration.dll"
"wpfgfx_cor3.dll"

These seem to be localization folders:
"cs"
"de"
"es"
"fr"
"it"
"ja"
"ko"
"pl"
"pt-BR"
"ru"
"tr"
"zh-Hans"
"zh-Hant"

Inside each localization folder:
"System.Windows.Input.Manipulations.resources.dll"
"System.Xaml.resources.dll"
"UIAutomationClient.resources.dll"
"UIAutomationClientSideProviders.resources.dll"
"UIAutomationProvider.resources.dll"
"UIAutomationTypes.resources.dll"
"WindowsBase.resources.dll"
"WindowsFormsIntegration.resources.dll"
"Microsoft.VisualBasic.Forms.resources.dll"
"PresentationCore.resources.dll"
"PresentationFramework.resources.dll"
"PresentationUI.resources.dll"
"ReachFramework.resources.dll"
"System.Windows.Controls.Ribbon.resources.dll"
"System.Windows.Forms.Design.resources.dll"
"System.Windows.Forms.Primitives.resources.dll"
"System.Windows.Forms.resources.dll"
```

> What is the stacktrace of the exception(s) if any?
N/A

> Do you have a code snippet or project that reproduces the problem?

1. Create a File > New MAUI app
2. Update OnCounterClicked
```
    private void OnCounterClicked(object sender, EventArgs e)
    {
        var source = Observable.Timer(DateTimeOffset.MinValue, TimeSpan.FromSeconds(1)).Timestamp();
        source.Subscribe(x =>
        Application.Current.Dispatcher.Dispatch(() =>
        {
            CounterLabel.Text = x.Timestamp.ToString();
        }));
        SemanticScreenReader.Announce(CounterLabel.Text);
    }
```
3. Add System.Reactive nuget or the following in csproj:
```
	<ItemGroup>
	  <PackageReference Include="System.Reactive" Version="5.0.0" />
	</ItemGroup>
```
4. Build the app in command prompt
```
msbuild MauiReactive.sln -restore -p:Configuration=Release
```
5. Package the app in command prompt
```
"C:\Program Files (x86)\Windows Kits\10\App Certification Kit\MakeAppx" pack /v /h SHA256 /d "MauiReactive\bin\Release\net6.0-windows10.0.19041\win10-x64" /p MauiReactive.msix
```

### LINKED ISSUE #1847: Split platform specific schedulers into separate packages

idg10:
Deprecate all UI-framework-specific and platform-specific types in System.Reactive. Add various System.Reactive.Integration.* packages with the replacement types.

Resolves #1745
Resolves #1847


glopesdev:
@idg10 this looks great! Is there a reason why we can't drop the "Integration" bit and make these just System.Reactive.WindowsForms, System.Reactive.Wpf, etc? I was searching nuget.org but couldn't find any existing listed package (maybe they are deprecated and unlisted?).

HowardvanRooijen:
If you've been keeping track of this and the mammoth ADR that preceded it, the first BIG problem was working out what to do. The second BIG problem is working out what to call it 🤣

glopesdev:
@idg10 For some reason I cannot find any of the other compatibility facades except for the WindowsRuntime one. Do you have links for them?

I'm not sure if creating more packages will do anything to alleviate the confusion, in my experience this tends to always increase with number of packages.

I will need to digest this a bit, but "Integration" seems far too abstract.

If absolutely necessary to introduce a new root name I would prefer "PlatformServices" mainly because it has a recognisable meaning dating back to the original Rx architecture.

glopesdev:
@idg10 I think now I understand what is going on and why I couldn't find any of the existing packages, and it might actually work to our advantage if we are willing to clean everything up properly.

Here is the list of the currently existing "integration" packages:
- [`System.Reactive.Windows.Forms`](https://www.nuget.org/packages/System.Reactive.Windows.Forms)
- [`System.Reactive.Windows.Threading`](https://www.nuget.org/packages/System.Reactive.Windows.Threading)
- [`System.Reactive.WindowsRuntime`](https://www.nuget.org/packages/System.Reactive.WindowsRuntime)

My proposal would be to name the packages the following:
- `System.Reactive.WindowsForms`
- `System.Reactive.Wpf`
- `System.Reactive.Uwp`
- `System.Reactive.Maui`
- `System.Reactive.WindowsRuntime`

From this list, my understanding is we have overlap with just one package (`System.Reactive.WindowsRuntime`). I don't think overlapping the two functionalities of forwarding the backwards compatibility facades and containing the new integration bits are that much of a problem, especially because we will be doing already this hybridization in the core `System.Reactive` package. If the end goal in several years is to do a full cleanup of the namespaces, I think this is worth it.

The aforementioned naming scheme I think would be the preferred one, since I think there is an implicit understanding that creating a sub-namespace with a platform or framework name means an integration with that platform or framework (this is used extensively within several existing Microsoft and third-party NuGet package names).

JakenVeina:
`System.Reactive.Extensions.Wpf` and `System.Reactive.Extensions.Uwp`?

anaisbetts:
Instead of "Integration" just use the word "For"

`System.Reactive.For.Uwp`

mikemilleresq:
Have you considered Targets? very similar to the For suggestion.

dezfowler:
I think I'd like to see this follow the recent MS convention of ancillary stuff like frameworks and UI not having the `System.` prefix.

So have it exist in a separate tree altogether; but what that tree is I don't know as I feel like all the reasonable prefix choices are probably burnt already.

So, for example, you have...

`System.Reactive`

and then something like...

`Reactive.Extensions.Wpf`

Or...

`ReactiveUI.Extensions.Wpf`

kronic:
System.Reactive.Extensions.Uwp
System.Reactive.Extensions.Blazor
System.Reactive.Extensions.WPF

idg10:
I really like @anaisbetts suggestion of `System.Reactive.For...` I think it's a big improvement on my `Integration.*` names.

@dezfowler suggested not using `System.Reactive` at all. This has its attractions but I see one particular problem with it: it's a lot less clear that packages such as `Reactive.Extensions.Wpf` are part of the same world as `System.Reactive`. If we were going to rename everything, including the core Rx.NET package, then perhaps it would make more sense. But we've always wanted to keep `System.Reactive` as the main package. (And although we initially thought that this wasn't practical, the discovery of a workaround changed that.)

I don't want to completely dismiss the idea (using something other than a `System` package name) because I think it has its upsides. But I think it is worth highlighting the downsides. If most people like the idea even after any problems have been fully examined, then we shouldn't ignore that.

It's worth noting that because Rx.NET is a .NET Foundation project, we are able to use certain reserved prefixes in our package names. Normally you're not allowed to publish packages with a name starting `System`, but we get an exemption, and this is part of what enables people to be confident that they are using the right package.

There are already `Reactive.*` packages out there (e.g. ReactiveUI's [Reactive.Wasm](https://www.nuget.org/packages/Reactive.Wasm) so I don't think this is likely to be made a protected package prefix like `System` and `Microsoft` are

Supply chain attacks have become very important over recent years, so I really don't want to remove a layer of protection against those that Rx.NET consumers currently get.

So for me, these are the two main arguments against `Reactive.Extensions.Wpf` etc. 1) It makes it harder to understand that this is the officially supported package provided by the same team that provides `System.Reactive`. 2) it offers less defence against supply chain attacks.

glopesdev:
> I really like @anaisbetts suggestion of `System.Reactive.For...` I think it's a big improvement on my `Integration.*` names.

@idg10 I feel this suggestion is not acceptable for the following reasons:
- `For` is a reserved keyword in VB and only differs in casing from the reserved keyword `for` in C#. If we introduce `System.Reactive.For` namespace, this means that simply `using System.Reactive` can make this identifier visible all over existing and future codebases.
- In English, the word "for" as used here is a preposition linking two nouns which is only meaningful when read in its entirety. However, namespaces in C# can be pulled apart leaving the preposition meaningless. Again taking the common example of `using System.Reactive` this means we can then write an expression such as `For.WinForms.ControlScheduler.Schedule` which to me becomes extremely ambiguous in terms of meaning to the naive reader.

Finally, I don't think I have ever seen packages or namespaces designed as a composition of verbs and nouns meant to be read as a sentence, and most definitely not in the `System` namespace. If I was a consumer of such a package I think it would make me triple-check whether this is really the official package.

> True, but we would have both of these:
> 
> * `System.Reactive.WindowsForms`
> * `System.Reactive.Windows.Forms`

I agree these two forms should not coexist, and to me the solution here is simply to deprecate all legacy integration packages. This way it would make it clear to users what is the way forward.

I understand the argument for backwards compatibility, but what we are discussing here goes beyond that, and is starting to create undue burden on the project maintainers and the rest of the community. Who is really using these packages and how do we assess their impact? Download counts is not enough unfortunately because these could be automated CI workflows or other deployment artifacts so I doubt they will give us a clear idea of the real impact of the change.

These packages were designed in the .NET framework era which has itself been left behind by Microsoft itself. The .NET core team had no qualms to remove core features of .NET such as application domains, `CompileToMethod` and so many other great features in the name of cleaning up the ecosystem. Why do we, as an open-source community with drastically more limited resources, need to hold ourselves to a stricter standard and bend over backwards in the name of packages which only target UAP or .NET framework?

I keep having the strong feeling we are coming up with proposals that risk compromising the future of a great library by introducing further and further confusion, and I think continuing to update and support these old facade packages just keeps sending the wrong message to users, especially if they are advertised using names which are the most natural to look for.

If we do have to give up on the natural names, then to me the best proposal I have seen so far is to go with `System.Reactive.Extensions.*`, which at least stays true to the original spirit of Rx being "Reactive Extensions to the .NET framework", where now we can have not just one but multiple "frameworks".

glopesdev:
> I believe the idea is to use the `for` convention just for the package name, not for the internal namespaces (for many of the reasons you highlighted).

@HowardvanRooijen thanks for the clarification, I agree the packages should be functional in that case. However, `System.Reactive.For.Uwp` still sounds uncanny to me. I guess the reason is that even if `For` is not an actual namespace name, it still makes up part of the assembly qualified name, with an enforced dot-separated identifier which conveys at least some notion similar to a namespace identifier.

Also the point remains that the biggest source of confusion is that the "legacy" packages are still being maintained. They should be actively deprecated, and their updates discontinued as soon as possible. If we keep updating them and `System.Reactive.Windows.Forms` still shows up on NuGet side-by-side with `System.Reactive.For.WinForms`, I somehow don't think it will be very clear to people which one they should pick up, and even less clear if they use the command-line.

I don't have many other arguments and apparently it doesn't bother anybody else, so I'll leave it at that, but for the record this doesn't look good to me.

dezfowler:
That [same comment](https://github.com/dotnet/reactive/blob/0e0272ad282dbacbf40f90a3fa0d451f876c600a/Rx.NET/Documentation/ReleaseHistory/Rx.v5-and-before.md?plain=1#L41) I linked above says that the package name should match...

> the dominant namespace in each package

So if the intent is to not have a `.For.` namespace then we'd be breaking that convention.

As I said, though, I would really like to see all of these integration packages using `Microsoft.Reactive` instead of `System.Reactive`... especially where they're targeting a Microsoft technology which uses `Microsoft.*` package naming itself.

glopesdev:
> As I said, though, I would really like to see all of these integration packages using `Microsoft.Reactive` instead of `System.Reactive`... especially where they're targeting a Microsoft technology which uses `Microsoft.*` package naming itself.

This is a great suggestion and consistent with existing package names coming from this repo, such as [`Microsoft.Reactive.Testing`](https://www.nuget.org/packages/Microsoft.Reactive.Testing).

@idg10 I like this naming split for two main reasons:

1. It completely separates the infrastructure for Rx from the platforming integrations; and encourages vendors to make their own integrations (i.e. same as Microsoft sponsoring integrations with WinForms or UWP, other vendors could provide integrations with Avalonia, Unity3D, etc).
2. There is a precedent for the Rx team to publish under this namespace, e.g. `Microsoft.Reactive.Testing`.

@dezfowler @idg10 If we go with this split would something like the below work?

- `Microsoft.Reactive.Windows.Forms`
- `Microsoft.Reactive.Uwp`
- `Microsoft.Reactive.Wpf`
- `Microsoft.Reactive.WindowsRuntime`

I don't feel confident to claim what are good "standard" names for WinUI platforms anymore so I need some help here (is WPF different from WindowsRuntime, and what is the deal with UWP anyway?).

dezfowler:
> > As I said, though, I would really like to see all of these integration packages using `Microsoft.Reactive` instead of `System.Reactive`... especially where they're targeting a Microsoft technology which uses `Microsoft.*` package naming itself.
> 
> This is a great suggestion and consistent with existing package names coming from this repo, such as [`Microsoft.Reactive.Testing`](https://www.nuget.org/packages/Microsoft.Reactive.Testing).
> 
> @idg10 I like this naming split for two main reasons:
> 
> 1. It completely separates the infrastructure for Rx from the platforming integrations; and encourages vendors to make their own integrations (i.e. same as Microsoft sponsoring integrations with WinForms or UWP, other vendors could provide integrations with Avalonia, Unity3D, etc).
> 2. There is a precedent for the Rx team to publish under this namespace, e.g. `Microsoft.Reactive.Testing`.
> 
> @dezfowler @idg10 If we go with this split would something like the below work?
> 
> * `Microsoft.Reactive.WindowsForms`
> * `Microsoft.Reactive.Uwp`
> * `Microsoft.Reactive.Wpf`
> * `Microsoft.Reactive.WindowsRuntime`
> 
> I don't feel confident to claim what are good "standard" names for WinUI platforms anymore so I need some help here. Is WPF different from WindowsRuntime? And what is the deal with UWP anyway, do we even want to publish a totally new package for a deprecated technology, or just let the previous integration packages become naturally obsolete?

@glopesdev @idg10

Would probably go with the shorter versions for `WinForms` and `WinRT` but, yes. 

I was going to suggest something like add an insulating `.UI.` in there too (e.g. `Microsoft.Reactive.UI.WinRT`) in the event of there being common UI pieces that need somewhere to live (i.e. a `Microsoft.Reactive.UI` package) but... 

The more I think about this stuff the more I wonder whether these pieces should just be retired completely from this repo in favour of [ReactiveUI](https://github.com/reactiveui/ReactiveUI/tree/main) which has a ton of UI technologies already covered and already has an [Avalonia-provided package](https://www.nuget.org/packages/Avalonia.ReactiveUI) too.

glopesdev:
Reading back with fresh eyes all these comments after several months, a few things came to me (picking pieces which resonated).

> where now we can have not just one but multiple "frameworks".

Made me think that `System.Reactive.Frameworks`, `System.Reactive.UI`, or `System.Reactive.Schedulers` might work well as names if we wanted to keep everything rooted.

> The more I think about this stuff the more I wonder whether these pieces should just be retired completely from this repo

This also struck with me as possibly the most reasonable first step of simply retiring all functionality altogether.

> But we've always wanted to keep System.Reactive as the main package.

Given the new insights in a different thread (which I cannot even find anymore given the now many separate issues/discussions/PRs around this topic!), is this still the plan?

I was thinking that if we don't keep the same package name, we probably need to abandon the namespace `System.Reactive` entirely.

Why do I say this? Well, one of the big arguments for why we cannot just do a "clean break" seems to be that we don't want to stop supporting the evolution of legacy software which is both "legacy" enough to not upgrade to newer packages, but somehow at the same time "modern" enough to want to keep with the latest packages updates.

By this definition, this must imply that someone is recompiling these "legacy" apps or libraries regularly (otherwise none of this would matter). If they do recompile and build from source regularly, and happen to add modifications or even just updates to their dependencies (which we said by definition they do), then they could in that very same process unexpectedly add (directly or indirectly) a dependency to the new `System.Shiny.Reactive` package (fictitious placeholder name for the putative new clean slate package).

If by mistake or deliberately they were to pull this new dependency, and the new package still used `System.Reactive` as the namespace, the legacy maintainers would be diving straight into a hell ride, where now all references to types would be clashing and a fully qualified (with assembly name) resolution mechanism would be required just to build the code. It basically would detonate any interoperability regardless of package names.

So just for a sense of perspective it seems to me we have really only two choices:

1. Do a clean break with the past, cleanup all packaging properly, release a major version and deprecate all residual packages.

2. Abandon both the `System.Reactive` package *and* the `System.Reactive` namespace completely and absolutely.

P.S.: note that even in this last case of renaming everything you are still "abandoning" developers, specifically the ones who have been keeping their code bases up to date in the hopes their code becomes possible to integrate. So you would keep the "legacy" developers happy but essentially detonate the entire code base of active developers.

idg10:
Superseded by #2268

PR #2102, opened 2024-05-02
Proposal for moving async LINQ to BCL

idg10:
This is a response to @davidfowl's suggestion that we [migrate the `System.Linq.Async` code into the .NET runtime library](https://github.com/dotnet/reactive/discussions/1868#discussioncomment-4905880):

 > Throwing out an idea: it would be good to cede the IX operators back into the BCL given the first class support for IAsyncEnumerable<T>. There’s been many discussions about why we’re not adding async LINQ to the BCL and it’s because this package already does that. While I have no problems in general with an IX dependency, it feels like something that should be built in now given the in the box (including compiler support) for IAE.

### DIFF
--- a/Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
+++ b/Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
@@ -28,11 +28,27 @@ If that last fact has changed—if the BCL team _is_ now open to moving this cod
 
 ### Compatibility
 
+There are two options. The new built-in library could maintain compatibility with the published library, and we could release a new version of the `System.Linq.Async` package with type forwarders for newer runtimes. Or it could make no such attempt
+
+#### Pros of non-compatibility
+
+[Stephen Toub commented](https://github.com/dotnet/reactive/pull/2102#issuecomment-2090886494) on an earlier draft of this document, saying:
+
+>  were a bunch of choices made in the APIs exposed here that we would not want to bring in, e.g. an AggregateAsync method is fine, but not AggregateAwaitAsync or AggregateAwaitWithCancellationAsync... that's simply not how we'd choose to expose such functionality. We would also be able to design the new methods taking into account features that didn't previously exist, e.g. generic math
+
+Also, there are significant dangers in attempting to make functionality available through a public NuGet package when that functionality is also built into some but not all versions of .NET. If you ever had to deal with the [problems with `System.Net.Http` 4.1.1-4.3.0](https://github.com/dotnet/runtime/issues/20777) you will know how painful it is when packages get this wrong.
+
+#### Pros of compatibility
+
 The existing `System.Linq.Async` library is widely used: it has had 126 million downloads as of May 2024. Although it enjoys no official support from Microsoft, there are some factors that may cause people to perceive it otherwise—its name starts with `System`, and the package gets a blue tick followed by "by .NET Foundation" which strongly suggests that it's part of .NET, even though it isn't. (People continue to believe that Rx.NET is still a Microsoft thing, well over a decade after that stopped being true.) The Microsoft Developer YouTube channel positioned this library as "how to use LINQ with IAsyncEnumerable" in https://www.youtube.com/watch?v=Ktl8K2b1-WU and didn't make it at all clear that this is just a community-supported open source project.
 
-We should therefore consider what it would mean for existing users of this library if the code were to migrate into the .NET runtime libraries. Given its widespread usage and the inaccurate but widespread perception that the existing `System.Linq.Async` is a Microsoft thing, our view is that we want to avoid breaking changes. Projects already using `System.Linq.Async` should be able to upgrade to whichever version of .NET incorporates this functionality without needing to change anything in their code. Projects that have acquired a dependency on this package transitively, without the developers being aware that they've done so, shouldn't get any nasty surprises as a result of this change.
+Also, this part of David Fowler's original message is relevant:
 
-Ideally, we would like the .NET runtime libraries to offer a `System.Linq.Async` assembly with a public API fully compatible with the existing library. If this were to go into, say, .NET 10.0, we would publish a new version of the `System.Linq.Async` NuGet package with a `net10.0` target consisting entirely of type forwarders pointing into the .NET runtime library implementation.
+> we’re not adding async LINQ to the BCL and it’s because this package already does that
+
+We should therefore consider what it would mean for existing users of this library if the code were to migrate into the .NET runtime libraries. Given its widespread usage and the inaccurate but widespread perception that the existing `System.Linq.Async` is a Microsoft thing, there is a strong case for avoiding breaking changes. Ideally, projects already using `System.Linq.Async` should be able to upgrade to whichever version of .NET incorporates this functionality without needing to change anything in their code. Projects that have acquired a dependency on this package transitively without the developers being aware that they've done so shouldn't get any nasty surprises as a result of this change.
+
+The .NET runtime libraries could offer a `System.Linq.Async` assembly with a public API fully compatible with the existing library. If this were to go into, say, .NET 10.0, we would publish a new version of the `System.Linq.Async` NuGet package with a `net10.0` target consisting entirely of type forwarders pointing into the .NET runtime library implementation.
 
 That way, code using the existing package would continue to work and would automatically migrate to the .NET runtime implementation in applications that target versions of .NET that have `System.Linq.Async` built in. Developers using such versions of .NET could remove the reference to the `System.Linq.Async` package if they wanted, but would not be required to. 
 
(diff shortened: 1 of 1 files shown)


Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
```diff
@@ -0,0 +1,135 @@
...
+> Throwing out an idea: it would be good to cede the IX operators back into the BCL given the first class support for IAsyncEnumerable<T>. There’s been many discussions about why we’re not adding async LINQ to the BCL and it’s because this package already does that. While I have no problems in general with an IX dependency, it feels like something that should be built in now given the in the box (including compiler support) for IAE.
+
+We would like this to happen. Our initial focus was on re-aligning the source code with current tools, and then addressing the problems that can occur with self-contained and AOT deployment when using Rx on a Windows-specific target. But we would now like to move this suggestion forward, if the .NET team is still willing.
+
+## Context
+
+The `System.Linq.Async` code is in the Rx.NET repository (this repository) as an accident of history. Specifically:
+
+* This repository is also home to Rx's mirror image, `System.Interactive`, aka **Ix**
+* Ix originally introduced its own asynchronous version of `IEnumerable<T>` many years before .NET runtime libraries added `IAsyncEnumerable<T>`
```
HowardvanRooijen, line 14: Shouldn't that be `IAsyncEnumerable<T>` rather than `IEnumerable<T>`?

Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
```diff
@@ -0,0 +1,135 @@
...
+* Ix already had a comprehensive implementation of the standard LINQ operators for its original asynchronous enumerable
+* When [IAsyncEnumerable<T>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.iasyncenumerable-1?view=net-8.0) was added in .NET Core 3.0, it made sense for Ix to start using that instead of its original definition
```
HowardvanRooijen, line 16: Should we link to the .NET Show episode where Bart demonstrates the new support? https://www.youtube.com/watch?v=Ktl8K2b1-WU

Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
```diff
@@ -0,0 +1,135 @@
...
+* At that point, Ix had a complete implementation of LINQ for `IAsyncEnumerable<T>`, and the .NET runtime libraries did not, so it make sense for Ix to publish these as the [`System.Linq.Async`](https://www.nuget.org/packages/System.Linq.Async) package we have today
```
HowardvanRooijen, line 17: When I was trying to explain what Ix was - there was a bit of confusion about the existing packages - what is `System.Interactive` vs System.Linq.Async`.
Is System.Interactive purely about the bespoke, legacy implementation of IAsync? Does the world still need this, or should it just be deprecated?

Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Use of T4 templates
+
+The existing code has various `.tt` files to generate multiple forms of the same code. The `Sum`, `Min`, `Max` and `Average` operators all use this to provide implementations for each of the built-in numeric types. Of course, as of .NET 7.0 this sort of thing isn't strictly necessary because we can use generic math instead. Unfortunately, the need to maintain binary compatibility means we can't do that in practice, because existing code will be looking for the specialized methods. Also we generate 2-, 3-, and 4- argument forms of some internal combination operators use to improve the efficiency of certain `Select` and `Where` scenarios.
```
HowardvanRooijen, line 57: Could modernizing the API to use the new numeric types be an option as part of the migration?

Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
mwadams: I read that as containing an implied parenthetical clause: "introduced its own asynchronous version of `IEnumerable<T>` (not confusingly at the time this was called `IAsyncEnumerable<T>`, now happily confused with the .NET runtime version of same)"

Ix.NET/Documentation/adr/0001-async-linq-to-bcl.md
mwadams: It would break the existing users, and that's explicitly not something we want to do, because people behave as if this is part of the BCL even if it isn't (so we've got very strict guarantees).

HowardvanRooijen: But if there is a time to cause such pain, this is it.

stephentoub:
We can certainly have the conversation again (in the dotnet/runtime repo would be a better place) about what it would look like to add IAsyncEnumerable extensions for the LINQ surface area into the core .NET libraries. But I'm fairly confident the compatibility goals outlined here would not be achieved. First, if we were actually adding this all in, I expect it would be with additions to the System.Linq.dll library and it would only be for .NET Core, no separate package, which means it wouldn't be available for .NET Framework or previous core releases. Second, there were a bunch of choices made in the APIs exposed here that we would not want to bring in, e.g. an AggregateAsync method is fine, but not AggregateAwaitAsync or AggregateAwaitWithCancellationAsync... that's simply not how we'd choose to expose such functionality. We would also be able to design the new methods taking into account features that didn't previously exist, e.g. generic math. And all that means that both a) there would be breaking changes (both source and binary) and b) the existing types could not just be type-forwarded to the new ones. We would have to be ok saying "there are these new APIs and if you're currently using System.Linq.Async.dll you'll have breaking changes you'll need to adapt to when you upgrade to the new version of .NET".

idg10:
One reason I had thought that breaking changes would be a problem is that David Fowler's original message said:

>we’re not adding async LINQ to the BCL and it’s because this package already does that

This implies that `System.Linq.Async` has for some years been the BCL team's answer to the question: how do I do async LINQ to Objects? So although it's not an officially Microsoft-recommended, supported package, it is in practice a thing Microsoft has been telling people to use.

Of course that doesn't mean that an implementation in the runtime libraries would be obliged to be compatible with the existing `System.Linq.Async`. It would just mean that there might be a non-trivial number of aggrieved developers unhappy about the situation. Also, this worried me slightly:

> if you're multi-targeting both core and framework, your life just got more complicated

It gave me flashbacks to the System.Net.Http 4.1.1-4.3.0 problems. I'm keen not to create another situation like that!

If it's a given that any in-the-box async LINQ would definitely not be either source- or binary-compatible (and it sounds like that is the case), it has me wondering whether I should deprecate the existing package, and release a new one with a different name and namespaces (something not in `System`) to a) clarify that this isn't the official implementation it might superficially seem to be and b) clear the path for an official implementation at some point in the future.

Here's one example of the sort of problem that could occur if we remain in the `System` namespace hierarchy and a future .NET provides async LINQ out of the box. Anyone who writes `using System.Linq;` (or `global using global::System.Linq;`; so any .NET >= 8.0 project that accepts the default `ImplicitUsings` setting) who has a reference to our `System.Linq.Async` might write this sort of code:

```cs
public async Task<bool> UseIae(IAsyncEnumerable<int> iae)
{
    return await iae.AllAsync(x => x % 2 == 0);
}
```

If they upgrade to a new .NET with built-in async LINQ, they would find that this is now ambiguous, because there are two different libraries both defining LINQ extension methods for `IAsyncEnumerable<T>`.

(If there were absolutely no overlap between the method names offered by the current `System.Linq.Async` and a new built-in async LINQ—e.g., if the built-in implementation stuck with the standard names like `All` instead of `AllAsync`—then this wouldn't occur. But even so, the idea of two competing async LINQ implementations offering extension methods for the same type, imported via the same namespace, sounds like a bad idea.)

So it seems like there may really be two separate questions here:

1. **Should the Ix.NET project change the package and namespace names, and deprecated the existing `System.Linq.Async` package?** (Probably.)
2. **Can we offer anything to the BCL team to help make built-in async LINQ happen?** (Debatable.)

We could do 1 even if we think there's a high chance the BCL team will never provide an official async LINQ. (And arguably we should, although there are pros and cons. If an official async LINQ never happens, this renaming would just have created some fruitless churn for everyone using the library.)

I'm aware 2 was never going to be as simple as moving our existing code into the .NET runtime repo. This would be a 'free as in puppy' contribution—in practice this would be a new, non-trivial eternal support commitment for the .NET team, and it may well be that after a thorough code review, they decide they don't much like our code and want to rewrite the whole thing. This code was all written in an era when much less attention was paid to GC costs and other performance refinements that people have come to expect from the .NET runtime today. We (the current maintainers of Rx.NET) don't have the budget to modernise `System.Linq.Async`, so I'm aware this isn't so much a gift from the Rx maintainers to the .NET team as it is a request for you to take this support overhead off our hands!

It was only because David Fowler suggested that this might be something the .NET team should do that we have been looking at this.

davidfowl:
> It was only because David Fowler suggested that this might be something the .NET team should do that we have been looking at this.

I still think it's the right path forward.

idg10:
This is happening - https://github.com/dotnet/reactive/issues/2199 - so there isn't much value in merging this proposal.

PR #2135, opened 2024-07-01
Update Ix.NET for .NET 8.0 sdk

idg10:
Ix.NET had not been updated since the .NET 8.0 SDK shipped, and it turns out that we have:

*  build failures because `MSBuild.Sdk.Extras` has not been updated since .NET 5.0, and
* vast numbers of compiler diagnostics because each SDK release gets pickier

This PR removes the use of `MSBuild.Sdk.Extras`. Nobody is maintaining it any more, so it's no longer viable to depend on it.

This also makes numerous small changes to comply with the majority of SDK diagnostics.

It leaves a small number in because they seem to indicate some genuine potential problems around cancellation. (They coincide with some "REVIEW" comments in the code indicating that possible problems here were already known about.)


HowardvanRooijen (approved):
LGTM. It's amazing how some of the new language features really clean up some of the boilerplate code.

Ix.NET/Source/refs/System.Interactive.Providers/System.Interactive.Providers.csproj
```diff
@@ -1,14 +1,22 @@
-﻿<Project Sdk="MSBuild.Sdk.Extras">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <ProduceOnlyReferenceAssembly>true</ProduceOnlyReferenceAssembly>
```
HowardvanRooijen, line 4: Do we document the purpose of this property anywhere?  I'm unfamiliar with it, so thinking it might be worth a comment.

Ix.NET/Source/System.Interactive.Tests/System.Interactive.Tests.csproj
```diff
@@ -1,8 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net48;net6.0;netcoreapp3.1</TargetFrameworks>
-    <NoWarn>$(NoWarn);CS0618;CS8603;CS8625</NoWarn>
+    <TargetFrameworks>net48;net8.0;net6.0;netcoreapp3.1</TargetFrameworks>
```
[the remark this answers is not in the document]
idg10: I had hoped that this read-me file in the solution's `test` folder would answer that:
image

but I hadn't counted on people looking directly at the `csproj` source and not noticing the explanatory file. So I've added comments to all the relevant `.csproj` files so people know the readme is there.

Ix.NET/Source/refs/System.Interactive.Providers/System.Interactive.Providers.csproj
idg10: It's a standard .NET SDK property that is easily found by searching the web. I don't want to open the can of worms of attempting to explain how reference assemblies work in each of the relevant `.csproj` files. Anyone needing to tinker with that will need to understand reference assemblies in some depth, and if they don't already have that knowledge, searching the web for this information is exactly what they need to do as a first step.

This PR does add an ADR explaining in detail the specific and unusual things Ix.NET does with reference assemblies, but there's already a comment in that file pointing to that ADR.

HowardvanRooijen (approved):

PR #2153, opened 2024-07-18
Enable application-defined Synchronize gate in AsyncRx.NET

### LINKED ISSUE #2148: Feature Request: Make `AsyncGate.LockAsync` cancellable
The underlying `SemaphoreSlim.WaitAsync` already supports cancellation.  Can a parameter be exposed to pass a token down to this line?

https://github.com/dotnet/reactive/blob/5f831de0bc70bc660d21c3b9e04e581269691a2a/AsyncRx.NET/System.Reactive.Async/Threading/AsyncGate.cs#L35

idg10:
The AsyncRx.NET `Synchronize` overloads that accept an application-supplied 'gate' (enabling multiple synchronized observables and observers to share a single lock, and also enabling application to acquire that same lock itself) can't use `object` like Rx.NET does, because the monitor  built into every .NET object supports only blocking lock acquisition. Avoiding thread blocking is a big reason for using AsyncRx.NET, so the normal .NET monitor doesn't really work here.

So AsyncRx.NET defines an `AsyncGate` type to enable this kind of shared lock but with an asynchronous lock acquisition method.

Our AsyncGate does not support cancellation, and for now at least we don't want to add it. (For one thing, it opens the can of worms of whether we want to attempt to support cancellation across the board in AsyncRx.NET. But also, more or less everyone who tries to add cancellation support to this sort of primitive ends up creating subtle bugs.)

However, we want it to be possible for applications that want cancellable awaits to have them in exchange for a bit of work. So this defines an IAsyncGate interface modifies the `Synchronize` operators to use that instead of requiring the `AsyncGate` type. This makes it possible to use application-supplied gate implementations.

Resolves #2148

### DIFF
--- a/AsyncRx.NET/System.Reactive.Async/Threading/IAsyncGate.cs
+++ b/AsyncRx.NET/System.Reactive.Async/Threading/IAsyncGate.cs
@@ -1,0 +1,73 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT License.
+// See the LICENSE file in the project root for more information. 
+
+using System.Reactive.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Reactive.Threading
+{
+    /// <summary>
+    /// Synchronization primitive that provides <see cref="System.Threading.Monitor"/>-style
+    /// exclusive access semantics, but with an asynchronous API.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This enables <see cref="AsyncObservable.Synchronize{TSource}(IAsyncObservable{TSource}, IAsyncGate)"/>
+    /// and <see cref="AsyncObserver.Synchronize{TSource}(IAsyncObserver{TSource}, IAsyncGate)"/>
+    /// to be used to synchronize access to an observer with a custom synchronization primitive.
+    /// </para>
+    /// <para>
+    /// These methods model the equivalents for <see cref="IObservable{T}"/> and <see cref="IObserver{T}"/>
+    /// in <c>System.Reactive</c>. Those offer overloads accepting a 'gate' parameter, and if you pass
+    /// the same object to multiple calls to these methods, they will all synchronize their operation
+    /// through that same gate object. The <c>gate</c> parameter in those methods is of type
+    /// <see cref="System.Object"/>, which works because all .NET objects have an associated monitor.
+    /// (It's created on demand when you first use <c>lock</c> or something equivalent.)
+    /// </para>
+    /// <para>
+    /// That approach is problematic in an async world, because this built-in monitor blocks the
+    /// calling thread when contention occurs. The basic idea of AsyncRx.NET is to avoid such
+    /// blocking. It can't always be avoided, and in cases where we can be certain that lock
+    /// acquisition times will be short, the conventional .NET monitor is still a good choice.
+    /// But since these <c>Synchronize</c> operators allow the caller to pass a gate which the
+    /// application code itself might lock, we have no control over how long the lock might be
+    /// held. So it would be inappropriate to use a monitor here.
+    /// </para>
+    /// <para>
+    /// Since the .NET runtime does not currently offer any asynchronous direct equivalent to
+    /// monitor, this interface defines the required API. The <see cref="AsyncGate"/> class
+    /// provide a basic implementation. If applications require additional features, (e.g.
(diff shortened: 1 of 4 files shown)


AsyncRx.NET/System.Reactive.Async/Threading/IAsyncGate.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </para>
+        /// </remarks>
+        public ValueTask<AsyncGateReleaser> LockAsync();
+
+        /// <summary>
+        /// Releases the lock. Applications typically won't call this directly, and will use
+        /// the <see cref="AsyncGateReleaser"/> returned by <see cref="LockAsync"/> instead.
+        /// </summary>
+        /// <remarks>
+        /// This method needs to be publicly accessible so that a single <see cref="AsyncGateReleaser"/>
```
gmkado, line 68: Would it make more sense to get rid of this method, make `AsyncGateReleaser` internal, and make `LockAsync` return `ValueTask<IDisposable>`, and let other implementers decide how to release the lock in the dispose?

And actually, I don't really understand the advantage of having a separate `AsyncGateReleaser` class vs returning `Disposable.Create(Release)` in `AsyncGate.LockAsync`

gmkado: Or is it to keep the semantics of a lock separate from the semantics of `IDisposable`?

idg10: The reason for not returning `IDisposable` is that that would require an allocation on the GC heap every time you called `LockAsync`.

The current implementation (i.e., before this PR) returns `struct`, and although that does implement `IDisposable`, C# is able to use its `Dispose` method from `using` statements or declarations directly without needing to cast it to `IDisposable`. (Casting a `struct` to an interface type causes it to be boxed.)

So when you write this sort of thing:

```cs
using (await gate.LockAsync())
{
    // Do work
}
```

it becomes something like this:

```cs
// Note: variable is of `struct` type, so no heap allocation required
AsyncGate.Releaser d = await gate.LockAsync();
try
{
     // Do work...
}
finally
{
    d.Dispose();
}
```

But if you make `LockAsync` return `ValueTask<IDisposable>` C# has to do something more like this:

```cs
// Note: variable is of an interface type, so it has to refer to an object on the heap
IDisposable d = await gate.LockAsync();
try
{
     // Do work...
}
finally
{
    d.Dispose();
}
```

With the existing design, in cases where there was no contention for the lock, the whole 'acquire, release' sequence could be completely allocation-free. We're using `ValueTask<T>` so a non-blocking `await` is allocation-free, and by having it return a `struct`, no allocation is required for the `using` logic either.

I wanted to preserve that characteristic. Most people aren't going to bring their own gate implementation, and I didn't want to increase the cost for the most common case in which people just use `AsyncGate`.

I did consider making the interface generic: `IAsyncGate<TReleaser> where TReleaser : struct, IDisposable`. This would enable each implementation to bring its own releaser implementation. But really the only point of the releaser is to enable `using`, so I didn't think there was really any use in making every implementation write its own version.

idg10: @glopesdev what would this `ValueDisposable` take as its input?

This `AsyncGateReleaser` takes the `IAsyncGate`, which defines the `Leave` method that it calls. So it is specialized for this scenario.

A more general-purpose `ValueDisposable` would need some more general purpose resource release mechanism, which would presumably be `IDisposable`. But the whole point of `AsyncGateReleaser` is that gives you an `IDisposable` (so you can use `using`) in a scenario where you don't already have one.

If you've already got an `IDisposable`, you don't need to wrap it. You can already use `using`.

So I think this pattern is only useful in cases where you have something other than `IDisposable` to start with. In this particular case, a pair of `Enter` and `Leave` methods is a common kind of API for locks, so it makes sense. (And it doesn't make sense for `AsyncGate` to implement `IDisposable` because `Leave` very much isn't the same thing as `Dispose`.) We could imagine an `ILeaveable` which captured this "I'm done with this for now, but might use it again" (as opposed to the inherently terminal `IDispose`) then I could see how a generic type of this kind could work. But there isn't such an interface, and it feels scope-creepish for Rx to define it.

But perhaps I'm missing how this would work in the absence of such an interface.

idg10:
An update on this: I'm still working on unifying test suites for Rx and AsyncRx. (Currently AsyncRx has no tests.) My intention is to get the test suite into place before moving forward with the rest of AsyncRx.

PR #2172, opened 2024-10-11
Fix ToObservable operator can throw unhandled exception

### LINKED ISSUE #1677: The ToObservable operator can throw unhandled exception
Hi, and happy new year! I think that I have found a bug in the [`ToObservable`](https://github.com/dotnet/reactive/blob/main/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs) extension method for async-enumerable sequences. In case an exception is thrown by the `GetAsyncEnumerator` invocation, the exception is not propagated by the `OnError` mechanism of the produced observable, and instead it is rethrown as un unhandled exception that crashes the process. Here is a minimal demonstration of this behavior:

```C#
using System;
using System.Linq;
using System.Threading;
using System.Collections.Generic;

public class Program
{
    public static void Main()
    {
        var observable = new BadAsyncEnumerable().ToObservable();
        observable.Subscribe(item => { }, error => Console.WriteLine(error.Message));
        Thread.Sleep(1000);
    }

    class BadAsyncEnumerable : IAsyncEnumerable<int>
    {
        public IAsyncEnumerator<int> GetAsyncEnumerator(
            CancellationToken cancellationToken = default)
        {
            throw new Exception("Oops!");
        }
    }
}
```

**Expected behavior:** The error message is printed in the console.

**Actual behavior:** Unhandled exception.

[Try it on Fiddle](https://dotnetfiddle.net/TxPl28).

fedeAlterio:
Fixes #1677 
The issue was that exceptions thrown by GetAsyncEnumerator() were not being propagated to observer.OnError(), resulting in unhandled exceptions. This has been addressed by wrapping the GetAsyncEnumerator() call in a try-catch block,

### DIFF
--- a/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/ToObservable.cs
+++ b/Ix.NET/Source/System.Linq.Async.Tests/System/Linq/Operators/ToObservable.cs
@@ -329,6 +329,33 @@ namespace Tests
             Assert.False(fail);
         }
 
+        [Fact]
+        public void ToObservable_ShouldForwardExceptionOnGetEnumeratorAsync()
+        {
+            var exception = new Exception("Exception message");
+            Exception? recievedException = null;
+            var enumerable = AsyncEnumerable.Create<int>(_ => throw exception);
+            using var evt = new ManualResetEvent(false);
+
+            var observable = enumerable.ToObservable();
+            observable.Subscribe(new MyObserver<int>(_ =>
+                                                     {
+                                                         evt.Set();
+                                                     },
+                                                     e =>
+                                                     {
+                                                         recievedException = e;
+                                                         evt.Set();
+                                                     }, () =>
+                                                     {
+                                                         evt.Set();
+                                                     }));
+
+            evt.WaitOne();
+            Assert.NotNull(recievedException);
+            Assert.Equal(exception.Message, recievedException!.Message);
+        }
+
         private sealed class MyObserver<T> : IObserver<T>
         {
             private readonly Action<T> _onNext;
--- a/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
+++ b/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs
@@ -38,39 +38,57 @@ namespace System.Linq
 
                 async void Core()
                 {
-                    await using var e = _source.GetAsyncEnumerator(ctd.Token);
-                    do
+                    IAsyncEnumerator<T> e;
+
+                    try
                     {
-                        bool hasNext;
-                        var value = default(T)!;
-
-                        try
-                        {
-                            hasNext = await e.MoveNextAsync().ConfigureAwait(false);
-                            if (hasNext)
-                            {
-                                value = e.Current;
-                            }
-                        }
-                        catch (Exception ex)
-                        {
-                            if (!ctd.Token.IsCancellationRequested)
-                            {
-                                observer.OnError(ex);
-                            }
-
-                            return;
-                        }
-
-                        if (!hasNext)
-                        {
-                            observer.OnCompleted();
-                            return;
-                        }
-
-                        observer.OnNext(value);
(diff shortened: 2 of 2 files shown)


fedeAlterio:
@dotnet-policy-service agree

fedeAlterio:
Hi @idg10,

Apologies for the ping! Could you take a look at the fix when you have a chance? This bug is quite severe as it's causing the process to completely shut down.  We’ve implemented internally a custom alternative as a .ToObservableSafe(), but it’s challenging to enforce this version consistently due to the presence of .ToObservable within System.Linq that seems really authoritative.
Thank you very much for your help!

idg10:
Sorry for the delay - many of us at endjin have been unwell, and since nobody pays us to maintain Rx (and we only maintain `System.Linq.Async` because for historical reasons, it lives in the same repo) it tends to drop to the back of the queue under these circumstances.

Looking at this, I think there are 4 obvious options when `GetAsyncEnumerator` throws:

1. terminate the process abruptly (not good, but apparently the current behaviour)
2. throw the exception out of the call to `Subscribe`
3. forward the exception to the subscriber's `OnError`
4. quietly swallow the exceptions.

Although 2 is arguably logically like the right thing to do (because this is a case of failing even to get started) in practice this would involve sync-over-async, so we can't do that. For AsyncRx, `SubscribeAsync` could actually do this, but then you'd have a problem in which the two libraries ended up handling the same error quite differently.

So it does seem like 3 is the only reasonable option left if we want to report the exception. I see you've also implemented 4, allowing people to say they'd like exceptions to be swallowed. That's the one part of this I'm not sure about. I'm glad you've not made it the default, but I'm not sure whether we really want this at all. It seems to enable a bad practice.

What was your rationale for allowing `ignoreExceptionsAfterUnsubscribe: true`?

fedeAlterio:
Thank you for your answer.
I was trying to be consistent with Observable.FromAsync() behavior, but maybe i misunderstood it: https://github.com/dotnet/reactive/blob/7259cabf5456a21462caea8cb81da8e5d5405db2/Rx.NET/Source/src/System.Reactive/Concurrency/TaskObservationOptions.cs#L27C4-L50C28

Maybe we could just handle DisposeAsync exceptions like other exceptions? So invoking observer.OnError if the cancellation token is not canceled yet?

fedeAlterio:
I agree that the check for unsubscription is useless if the exception is thrown synchronously. Thank you!


About that `ignoreExceptionsAfterUnsubscribe `flag, I think there are 2 paths:
- We can remove it completely, swallowing any exception in case the unsubscription already happened (This would be similar to the current implementation)
- We can add it


If we choose the latter, in the case the subscription is already disposed:
- If the flag is `false`maybe we should rethrow all the exceptions (`throw` instead `return` in every `catch` clause). The Core method should also be a ValueTask so that exceptions will not cause the process to crash, but will be routed to `TaskScheduler.UnobservedTaskException`
- if it is `true`, we should just return (similarly at what happens now)

Consider also that following that approach, if `false` is the default, we would be inserting a breaking change (`TaskScheduler.UnobservedTaskException` will now be called)

Also note with the current implementation, the process crashes also if any exception is thrown by DisposeAsync()

idg10:
We probably won't be merging this now, because of https://github.com/dotnet/runtime/issues/79782 - Async LINQ is going to be built into .NET from v10, at which point the Async LINQ library in this repo will be deprecated.

I don't believe they are including `ToObservable` in that implementation, because they are just providing core LINQ features, and not Rx support. So we will actually need to find a new home for this extension. I think it might actually make sense for this to be in the main Rx.NET implementation, now that `IAsyncEnumerable` is built into .NET. Although that might not accommodate .NET FX users. This needs some thought, but I don't think the form it's in now is the right one.

fedeAlterio:
In my opinion it would be great to put this conversion method in .NET. 
IObservable interface is already there and
Something similar already exists for [dataFlow blocks](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.dataflow.dataflowblock.asobservable?view=net-9.0. ).

PR #2180, opened 2024-11-24
Adding ObservableDisposable

nilsauf:
Hej,

this PR adds the class ObservableDisposable, which allows to observe when the disposable will be disposed.

This comes in handy when you want to end streams if the parent object is disposed.
I use it in serveral of my projects and want to share it, albeit the implementation might not be perfect. (I am also here to learn 😄)

Example:

```
private ObservableDisposable observableDisposable = new();

IObservbale<T1> oldObservable = ... ;
IObservbale<T2> newObservable = oldObservable
	.TakeUntil(this.observableDisposable)
	.Select(t1 => */ read data /*)
	.Publish()
	.RefCount(


public void Dispose() => this.observableDisposable.Dispose()
```

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
+++ b/Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
@@ -1,0 +1,38 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT License.
+// See the LICENSE file in the project root for more information. 
+
+namespace System.Reactive.Disposables;
+using System;
+using System.Reactive.Subjects;
+using System.Threading;
+
+/// <summary>
+/// Represents a disposable resource that can be observed
+/// </summary>
+public sealed class ObservableDisposable : ICancelable, IObservable<Unit>
+{
+    private Subject<Unit> _disposedNotification = new();
+    private bool _disposed;
+
+    /// <inheritdoc/>
+    public bool IsDisposed => Volatile.Read(ref _disposed);
+
+    /// <inheritdoc/>
+    public IDisposable Subscribe(IObserver<Unit> observer)
+        => _disposedNotification.Subscribe(observer);
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        if (Volatile.Read(ref _disposed) == false)
+        {
+            _disposedNotification.OnNext(Unit.Default);
+            _disposedNotification.OnCompleted();
+            _disposedNotification.Dispose();
+            Volatile.Write(ref _disposed, true);
+            _disposedNotification = null!;
+        }
+        GC.SuppressFinalize(this);
+    }
+}
--- a/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
+++ b/Rx.NET/Source/tests/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
@@ -903,5 +903,35 @@ namespace ReactiveTests.Tests
             Assert.True(disp3);
             Assert.True(d.IsDisposed);
         }
+
+        [TestMethod]
+        public void ObservableDisposable_Observed()
+        {
+            var disp1 = false;
+
+            var observableDisposable = new ObservableDisposable();
+
+            var sub = observableDisposable
+                .Subscribe(_ => { Assert.False(disp1); disp1 = true; });
+
+            observableDisposable.Dispose();
+
+            Assert.True(disp1);
+            Assert.True(observableDisposable.IsDisposed);
+        }
+
+        [TestMethod]
+        public void ObservableDisposable_ObservationCancelled()
+        {
+            var disp1 = false;
+
+            var observableDisposable = new ObservableDisposable();
+
+            var sub = observableDisposable
+                .Subscribe(_ => { Assert.False(disp1); disp1 = true; });
+
+            sub.Dispose();
+
+        }
     }
 }


nilsauf:
@dotnet-policy-service agree

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
```diff
@@ -0,0 +1,38 @@
...
+    public bool IsDisposed => Volatile.Read(ref _disposed);
+
+    /// <inheritdoc/>
+    public IDisposable Subscribe(IObserver<Unit> observer)
+        => _disposedNotification.Subscribe(observer);
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        if (Volatile.Read(ref _disposed) == false)
```
idg10, line 28: I'm not sure if it's your intention for this code to work in the face of concurrent calls to `Dispose`. The use of `Volatile.Read` does suggest that this is the goal, although maybe this is just intending to handle correctly strictly sequential calls to `Dispose` that happen on different threads.

If this is meant to handle concurrent calls, I think there's a bug here. If two threads call `Dispose` simultaneously, it's possible for each of them to complete this `Volatile.Read` before the other reaches the `Volatile.Write`. So both threads will end up calling `OnNext` and `OnCompleted`. We will have violated the `IObserver<T>` contract:

* A call to `OnCompleted` followed by further calls (at least one more `OnCompleted`, and possibly a call to `OnNext` after a call to `OnCompleted`)
* Potentiall starting a new calls into an `IObserver<T>` when another call was already in progress and had not returned

There are two ways to deal with this:

1. declare that this is the caller's problem
2. implement something more robust
3. rely on the fact that Rx will treat the `Subject<T>` as a suspect external source, and it will actually block illegal call sequences

We can rule out 3. Rx helps out applications that make such mistakes, up to a point, but we certainly shouldn't be relying on these safety guards in Rx's own code. In any case, while it would detect and block the `IObservable<T>` calls that occur after the first `OnCompleted` I don't think it would help at all with concurrent calls.

Option 1 actually has its charms (and may be what you intended, in which case the XML doc comments should definitely clarify that instances of this class are not safe for concurrent use; someone looking at the source code might see `Volatile` and assume that it is designed for multithreaded use). It's basically our official position for any `IObserver<T>` we hand out: concurrent calls are not supported, and you get undefined behaviour if you make them. In that case, this method could be simplified: you'd remove the use of `Volatile` entirely and make no attempt to handle concurrent use.

However, although that's reasonable for `IObservable<T>` (because there are well defined rules restricting its use) arguably it isn't for `IDisposable`. It's not unreasonable to want an `IDisposable` that tolerates concurrent calls to `Dispose`.

So that would leave 2. But although this could probably be fixed by using some sort of interlocked exchange instead of simply a volatile read, but in practice, I think if this were all changed to fit the `IProducer<T>` pattern, and if you used one of the `Sink<...>` types normally used as part of such an implementation, we'd need to look at how concurrent calls to `Dispose` interact with the existing disposal handling in there. This already uses `Interlocked.Exchange` to prevent double disposal, so it's possible you could take advantage of that.

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            _disposedNotification.OnNext(Unit.Default);
+            _disposedNotification.OnCompleted();
+            _disposedNotification.Dispose();
+            Volatile.Write(ref _disposed, true);
+            _disposedNotification = null!;
+        }
+        GC.SuppressFinalize(this);
```
idg10, line 36: Since this is a sealed class without a finalizer, I don't see how instances of this type can ever end up on the finalization queue, so this won't do anything will it?

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Reactive.Subjects;
+using System.Threading;
+
+/// <summary>
+/// Represents a disposable resource that can be observed
+/// </summary>
+public sealed class ObservableDisposable : ICancelable, IObservable<Unit>
+{
+    private Subject<Unit> _disposedNotification = new();
+    private bool _disposed;
```
idg10, line 16: As far as I can tell, you don't really need this, because when the object is in a disposed state, the `_disposedNotification` field is `null`. Removing this field, and just relying on the null/not-null state of `_disposedNotification` would simplify things because it removes an intermediate state where `Dispose` has changed this flag but hasn't yet got around to nulling out the `_disposedNotification` field.

We use a similar technique in the `Sink<...>` base class that almost all of our operators use in their implementation: we use a single field to a) hold the downstream observer and b) know whether we've been disposed. (We use a sentinel observer rather than `null` to signify disposal, but the basic principle is the same.)

However, as noted elsewhere, perhaps this should be using `Sink` and `Producer`, at which point you might be able to defer entirely to its disposal handling.

Rx.NET/Source/tests/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
```diff
@@ -903,5 +903,35 @@ public void StableCompositeDisposable_Nary2()
...
+        {
+            var disp1 = false;
+
+            var observableDisposable = new ObservableDisposable();
+
+            var sub = observableDisposable
+                .Subscribe(_ => { Assert.False(disp1); disp1 = true; });
+
+            sub.Dispose();
+
```
idg10, line 934: Shouldn't this also dispose the `observableDiposable`?

You've checked that the act of unsubscribing doesn't cause our subscriber to be notified. But I think we also want to check that unsubscribing means we no longer get notified _when the thing that normally causes notification happens_.

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <inheritdoc/>
+    public IDisposable Subscribe(IObserver<Unit> observer)
+        => _disposedNotification.Subscribe(observer);
+
+    /// <inheritdoc/>
+    public void Dispose()
+    {
+        if (Volatile.Read(ref _disposed) == false)
+        {
+            _disposedNotification.OnNext(Unit.Default);
```
idg10, line 30: My assumption was that this source would simply complete upon disposal. I wasn't expecting it to produce a value too.

Did you consider both options? I can see either design choice as being valid, and it's not obvious to me which is 'best'. If you thought of both, what made you choose this one.

In any case, the XML doc comments should say what the behaviour is.

Rx.NET/Source/tests/Tests.System.Reactive/Tests/Disposables/DisposableTests.cs
```diff
@@ -903,5 +903,35 @@ public void StableCompositeDisposable_Nary2()
...
         }
+
+        [TestMethod]
+        public void ObservableDisposable_Observed()
+        {
+            var disp1 = false;
+
+            var observableDisposable = new ObservableDisposable();
+
+            var sub = observableDisposable
```
idg10, line 914: This should test that we also get the completion notification, as well as the `OnNext`.

For completeness, we could also verify that we don't get an `OnError`.

idg10 (commented):
Thanks for this offering.

You said you were "here to learn." Brace yourself! 😊

I have one basic issue with this that I'd want to be addressed before we would merge this, which is that this currently looks more like code from an application consuming Rx than it does like Rx internals. I explain in the comments in this review why we normally implement observables in a particular way that looks different from how you'd normally do it as an Rx consumer.

I think there are also questions over exactly how this functionality should look in Rx's public API that need to be resolved. The current approach might be incompatible with addressing the preceding point.

I also have a broader question: how often have you needed this? Any time we add new surface area to Rx, that's more support overhead, more potential bugs, more surface area for security flaws. To be fair, this one doesn't look likely to be a big problem on any of those fronts, so I'm really just asking: how useful is this? I don't _think_ it's something we've had a request for before. (That's not a "no thanks" by the way. But if you've needed this a lot, that's definitely useful to know.)

danielcweber:
@nilsauf Just in case the utility of this PR is found to be not broad enough for an addition to the API, you could rework this into a `TakeUntil(CancellationToken ct)`-operator that'll complete the upstream source whenever the `CancellationToken` signals its cancellation. In combination with the already existing `CancellationDisposable`, this would do what you need (I guess) and its utility might be considered a bit broader (also just a feeling).

nilsauf:
@idg10 Thank you! As a consumer of this package for some years now, I appreciate this valuable insights. This is so interesting 😊 

Well, this code looks like it is consuming Rx because it always was 😉 I will refactor it with your suggestions.

I mostly use it with EventPattern Observables (from external "old-fashioned" events) and in front of `Publish().RefCount()`. Since the EventPattern Observables will never send a `OnComplete`,  I want to complete it myself.

I hope the following snippet will clarify my use case, I think it is very clean this way.

```
public class ExampleClass : IDisposable
{
    private readonly IObservable<T2> exampleStream;
    private readonly ObservableDisposable observableDisposable = new();

    public ExampleClass(IObservable<T1> eventStream)
    {
        exampleStream = eventStream
            .TakeUntil(this.observableDisposable)
	    .Select(t1 => */ read data /*)
	    .Publish()
	    .RefCount();
    }

    public IObservable<T2> Observe() => this.exampleStream;

    public void Dispose() => this.observableDisposable.Dispose()
}
```

I will address your other points in their own comment threads.

@danielcweber Your suggestion to my use case is also very nice. I would love to add this operator, if it is considered in favor of ObservableDisposable.

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using System.Reactive.Subjects;
+using System.Threading;
+
+/// <summary>
+/// Represents a disposable resource that can be observed
+/// </summary>
+public sealed class ObservableDisposable : ICancelable, IObservable<Unit>
+{
+    private Subject<Unit> _disposedNotification = new();
```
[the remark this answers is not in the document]
nilsauf: I understand the conflict the class being an obserable AND a disposable.

- If it will be a class or a `Disposable.Observable()` factory method, it will be be the first creation method of an observable outside of the `Observable` class (and QueryLanguage, if i see it right).
- If it will be a factory method in the `Observable` class, it will be the first disposable outside of the Dispoable namespace.

Would it be an option to go both ways to create such object? 

I also used Subject here to not have to worry about the list of observers. I had a go at implementing `ObservableDisposable` as Producer and while this is great, I now have a list of sinks to notify and worry about. But maybe there is a better solution?

Since this Observable would only send one OnNext and one OnComplete notification, I think the suboptiomal performance is acceptable.

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
nilsauf: Yes, `Interlocked.Exchange` seems to be the better solution.

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
nilsauf: Well I shouls discard this call then... 🤦

Rx.NET/Source/src/System.Reactive/Disposables/ObservableDisposable.cs
nilsauf: Since I use it mostly with the `TakeUntil` Operator, I want to produce a value. I added OnComplete, because I think it is good behavior to complete a created stream. 

I also like to support both scenarios.

idg10:
If this is the usage model, then I think @danielcweber's suggestion is an improvement over the `ObservableDisposable` concept.

The reason I prefer it is that your usage model is an adaptation of an existing `IObservable<T>`: you have an `IObservable<T>` going in and an `IObservable<T>` coming out. This sort of thing is normally modelled as an operator in Rx, so this looks like it might be the right way to handle this particular scenario.

I apologise for the ramble that follows, but if we're introducing something new, I think it's really important to consider all the options. So I wanted to think about a few other ways we could approach this, so we can be confident that a) there isn't already a better way to solve this problem and b) that we've got the best solution.

I briefly wondered if it should be called `TakeUntilCancelled`, but `TakeUntil` already offers a few overloads that accept different kinds of temporal condition, so adding one more seems reasonable. Currently, `TakeUntil` can use a `DateTimeOffset`, or a predicate that inspects each object emerging from the source, or some other `IObservable<T>`. So the idea of adding an overload that accepts a `CancellationToken` doesn't seem like a huge leap.

Two other options have occurred to me.

One is that we could model this as a Subject, e.g. we could introduce a `CancellableSubject<T>`. With the initial design I was slightly uneasy about a Disposable implementing `IObservable<T>`. A Subject seems like a better way to approach it, because your use case is basically forwarding values from some underlying source, but with some additional behaviour around how that is done. My main worry with this, though, is that since `SubjectBase<T>` already implements `IDisposable`, the usage model might not be obvious - people might expect to be able to cancel the subject with `Dispose`, but that actually causes all further calls to OnXxx to throw an `ObjectDisposedException`.

The second is that we could define `AsObservable` extension methods for `CancellationToken`. Just like we have `AsObservable` extensions for `Task`, we could define them for `CancellationToken`. In a way, this feels like a more fundamental mechanism. But I'm not sure exactly what the design would look like. To be useful for the `TakeUntil` scenario you need to be able to produce a single value upon cancellation (because `TakeUntil` ignores a source that completes without producing a value). So you might want to be able to write `ct.AsObservable(Unit.Default)`. But perhaps it should be able to take any observable, so maybe you'd want to be able to write `ct.AsObservable(Observable.Return(Unit.Default))`. I don't think I like the idea of a zero-args `ct.AsObservable()` that just produces a single `Unit`, because it doesn't seem obvious to me that that's what it would do.

I have one nagging doubt about the whole concept for this though (regardless exactly how we would present it in the API).

Seeing your usage model, this reminds me of things Rx already does to ensure that we don't emit notifications after unsubscription. When `Dispose` is called on a subscription, we 'quench' all further notifications by putting the subscription into a state where it will not call any more methods on the subscriber's `IObserver<T>`. (We do this before tearing down everything inside the subscription, and it prevents spurious further notifications emerging while we shut things down.)

So I'm getting a feeling of "don't we already have this mechanism?" I feels like it might be one of those situations where there's a slightly different way of solving the problem that avoids needing this new mechanism entirely.

The thing you're looking to achieve here is a consumer-side thing: essentially you're saying "I no longer wish to hear from this source". But you're modelling it as an `IObservable<T>`. I tend to think about observables as being a _potential_ stream of values, and a subscription as being one particular realisation of that potential. So it feels slightly weird to have an observable change its state just because a consumer is now done with it. That feels like it's a subscription concern, not an observable concern.

E.g., think about `Observable.Timer`. When you create one of those, you don't actually start a timer. You build a description of a timer, and it's only when you subscribe that an actual timer comes into existence. If you subscribe a bunch of times you get a bunch of timers. And each subscriber gets to say individually "I'm done now, thanks." If you want to share you can use `Publish` but the standard semantics are that each subscription has its own lifetime.

So having an observable that can be put into a state where all further new subscriptions instantly get unsubscribed makes me slightly uneasy. Now admittedly there's a precedent: this is precisely what `AsyncSubject<T>` does, which would be a point in favour of modelling this idea as a `CancellableSubject` I suppose. And actually, thinking about it, the `TakeUntil` overload that takes a `DateTimeOffset` already has this characteristic: once the specified time has passed, all subsequent subscriptions to that source will get no elements. So there is a precedent there.

So right now, I'm still leaning towards `TakeUntil` as the answer.

nilsauf:
Yes, I also prefer the new `TakeUntil` operator now. 

Actually `CancellationDisposable` is doing the same as my original proposal, just not as IObservable. And I think the new operator could be used in much more scenarios.

Thank you both for your ideas and insight into this library!

I will close this PR and open an issue about the new `TakeUntil` operator, to discuss details about it.

idg10:
FYI, this is available to test on the Rx.NET nuget feed. If you add this `NuGet.config` to your solution folder:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
    <add key="Rx.NET" value="https://pkgs.dev.azure.com/dotnet/Rx.NET/_packaging/RxNet/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

and then add a reference to **this specific version** (there are versions on that feed that appear to have a later version number, but which don't have this feature):

``xml
  <ItemGroup>
    <PackageReference Include="System.Reactive" Version="6.1.0-preview.6" />
  </ItemGroup>
```

then the following works:

```cs
using System.Reactive.Linq;

CancellationTokenSource cs = new();

Observable
    .Interval(TimeSpan.FromSeconds(0.5))
    .TakeUntil(cs.Token)
    .Subscribe(Console.WriteLine);

Console.WriteLine("Subscribed. Press any key to unsubscribe.");

Console.ReadKey();
cs.Cancel();


Console.WriteLine("Unsubscribed. Press any key to exit.");
Console.ReadKey();
```

Issue #2181, opened 2024-12-10
Feature: TakeUntil Operator with CancellationToken


nilsauf:
As discussed in #2180 I would like to add a new `TakeUntil` operator that takes a `CancellationToken` as parameter.

I think the best approach is to implement it as close as possible to the other `TakeUntil` operators, especially the one that takes a `DateTimeOffset`.

It should forward all notifications from the source to new observers and register the `OnComplete` Method on the prodived `CancellationToken`.
If the `CancellationToken` is already cancled, it should never subscibe to the source and call the `OnComplete` Method directly.

I would be happy to implement it.

idg10:
There is now a preview of Rx 6.1 up on NuGet including this feature: https://www.nuget.org/packages/System.Reactive/6.1.0-preview.9

idg10:
This is in the final version of Rx 6.1, which is now available: https://www.nuget.org/packages/System.Reactive/6.1.0

PR #2182, opened 2024-12-13
Implementing TakeUntil with CancellationToken

nilsauf:
This PR implements a new TakeUntil operator with CancellationToken, as proposed in #2181

### DIFF
--- a/Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeUntilCancellationToken.cs
+++ b/Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeUntilCancellationToken.cs
@@ -1,0 +1,76 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT License.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Reactive.Linq.ObservableImpl;
+using System;
+using System.Reactive.Disposables;
+using System.Threading;
+
+/// <summary>
+/// Relays items to the downstream until the CancellationToken is cancled.
+/// </summary>
+/// <typeparam name="TSource">The element type of the sequence</typeparam>
+internal sealed class TakeUntilCancellationToken<TSource> :
+    Producer<TSource, TakeUntilCancellationToken<TSource>._>
+{
+    private readonly IObservable<TSource> _source;
+    private readonly CancellationToken _token;
+
+    public TakeUntilCancellationToken(IObservable<TSource> source, CancellationToken token)
+    {
+        _source = source;
+        _token = token;
+    }
+
+    protected override _ CreateSink(IObserver<TSource> observer) => new(observer);
+
+    protected override void Run(_ sink) => sink.Run(this);
+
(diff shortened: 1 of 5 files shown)


idg10:
My apologies for the delay in getting to this. I ended up being very busy with customer work for the first part of this year, and have only just got back to working through the backlog of work. (I don't do Rx maintenance full time)

So there are two main issues here:

1. you've now got a merge conflict, so I can't merge this PR
2. you decided to replace the old-style block `namespace` syntax with the new one in the files you modified

That 2nd point is the main reason you now have a merge conflict. Because you changed `namespace` style, Git thinks you changed almost every line of the files you modified.

While I do want to move over to the new style at some point, I don't want a mix of styles in the repo, so when we do move over to the new style, I'll do it with a single PR that does the whole repo.

I realise that since 9 months has passed since you submitted this, you might no longer have any time available to work on this, but if you can, could you retain the current `namespace` style (block-style, not the new style), which I think will have the side effect of getting rid of the merge conflict.

We'll be doing a new 6.1 release soon because I recently merged another change that adds an API, and I'd like to get this change into that as well if possible. If you don't have time to do anything with this, please let me know whether you'd like:

1. me to copy your changes onto a new branch and merge them
2. these change not to be incorporated

Rx.NET/Source/src/System.Reactive/Linq/Observable/TakeUntilCancellationToken.cs
```diff
@@ -0,0 +1,76 @@
...
+// The .NET Foundation licenses this file to you under the MIT License.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Reactive.Linq.ObservableImpl;
+using System;
+using System.Reactive.Disposables;
+using System.Threading;
+
+/// <summary>
+/// Relays items to the downstream until the CancellationToken is cancled.
```
idg10, line 11: Type "cancled" -> "cancelled".

nilsauf:
Thanks for looking into it!

Unfortunately I am on vacation the next 10 days and I have no access to a pc. 
So if you want to do the 6.1 release after that, I can work in your fixes. Otherwise it would be nice if you can pu them in. 

To the namespace issue: Reading my last comment, I think the .editorconfig of the solution forced the files copied namespace on saving. But maybe it was my personal visual studio settings? I can't verify that right now.

idg10:
>  I think the .editorconfig of the solution forced the files copied namespace on saving. But maybe it was my personal visual studio settings? 

I think it's more likely to be that second one. We didn't have any settings in the `.editorconfg` regarding namespace formatting, because that newer `namespace` syntax was introduced after we. I just added it.

Here's what I'm going to do. I'm going to adjust this PR to merge into a different branch, and then make the changes I think it needs, so we can get this in.

idg10:
I realised we were also missing the `IQbservable` version of this. (The tooling for generating the `IQbservable` versions needs some updates, so this process has more manual work than it's supposed to. I've done this in the new PR.) Also, we needed to update the API approval test's verified API. (Those tests ensure that we never make accidental changes to the public-facing API.)

This is all in place now in #2222 on top of your original work.

PR #2213, opened 2025-08-24
AsyncObseravableBase: Fix in case of exceptions thrown synchronously

fedeAlterio:
`IAsyncObserver<T> OnNextAsync, OnCompletedAsync and OnErrorAsync` can all throw exceptions synchronously,
This is not handled correctly in AsyncObseravableBase, and the execution flow will follow a different paths of the asynchronous one.

A concrete caused by this is that if a synchoronous exception is thrown inside OnCompleteAsync,  `AutoDetachAsyncObserver` will not Dispose the subscription automatically


fedeAlterio:
@idg10 There isn’t an open issue for these problems, but the code comments describe them in detail.
Would you prefer that I create a separate issue to document and track them outside of this PR?

Issue #2187, opened 2025-02-03
Multiple await on source returning the same exception reports multiple stack traces


adamjones2:
I'm finding something very strange: when I repeatedly subscribe to a failed sequence the exception that's returned gets the call stack of each subscription appended to it for future callers. This is using Rx v6.0.1. Repro:

```csharp
var src = Observable.Throw<int>(new Exception("test")).Replay(1);
src.Connect();

async Task GetValue()
{
    try { await src.FirstAsync(); }
    catch (Exception ex) { Console.WriteLine(ex); }
}

for (int i = 1; i <= 3; i++)
{
    Console.WriteLine($"\r\n===> {i}");
    await GetValue();
}
```

Output:
```
===> 1
System.Exception: test
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
   at System.Reactive.ExceptionHelpers.Throw(Exception exception)
   at System.Reactive.Subjects.AsyncSubject`1.GetResult()
   at Program.<>c__DisplayClass0_0.<<<Main>$>g__GetValue|0>d.MoveNext() in C:\Source\Test\Program.cs:line 8

===> 2
System.Exception: test
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
   at System.Reactive.ExceptionHelpers.Throw(Exception exception)
   at System.Reactive.Subjects.AsyncSubject`1.GetResult()
   at Program.<>c__DisplayClass0_0.<<<Main>$>g__GetValue|0>d.MoveNext() in C:\Source\Test\Program.cs:line 8
--- End of stack trace from previous location ---
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
   at System.Reactive.ExceptionHelpers.Throw(Exception exception)
   at System.Reactive.Subjects.AsyncSubject`1.GetResult()
   at Program.<>c__DisplayClass0_0.<<<Main>$>g__GetValue|0>d.MoveNext() in C:\Source\Test\Program.cs:line 8

===> 3
System.Exception: test
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception)
   at System.Reactive.ExceptionHelpers.Throw(Exception exception)
   at System.Reactive.Subjects.AsyncSubject`1.GetResult()
[...]

PR #2237, opened 2025-09-26
Add ResetExceptionDispatchState operator and document exception reuse limitations

idg10:
Issue #2187 illustrates a problem caused by a non-obvious feature of how Rx promotes `OnError` notifications to real exceptions in mechanisms such as `await`.

This PR provides two mitgations:

1. it documents the limitations, so that developers can have some way of knowing that there might be a problem
2. it provides a new `CaptureExceptionDispatchState` operator that can provide a straightforward way to avoid the problem


Rx.NET/Source/tests/Tests.System.Reactive/Tests/Linq/Observable/MinTest.cs
```diff
@@ -2354,7 +2354,7 @@ public void MinOfT_Selector_Regular()
                 OnCompleted<string>(240)
             );
 
-            var res = scheduler.Start(() => xs.Min(x => new string(x.ToCharArray().Reverse().ToArray())));
+            var res = scheduler.Start(() => xs.Min(x => new string(x.ToCharArray().AsEnumerable().Reverse().ToArray())));
```
carmeleve, line 2357: And here?

PR #2251, opened 2025-10-09
Fix condition in documentation example

barbaralocsi:
Negation is missing in `ReadFileLines` example.


HowardvanRooijen:
Thanks! I spotted your TryRx repo! That's great. Have you looked at Polyglot Notebooks - they're great for exploring / experimenting: https://code.visualstudio.com/docs/languages/polyglot

PR #2258, opened 2025-10-19
Fix typos in docs

barbaralocsi:


idg10 (approved):
Thanks!

idg10:
Hi @barbaralocsi and thanks for this contribution. GitHub is blocking the merge because it doesn't think you've agreed to the Contributor License Agreement (CLA). We can't accept PRs from anyone who hasn't agreed to the CLA, because that's the only way we can ensure that we remain in compliance with copyright and licensing requirements.

GitHub should have given you a link to fill in the CLA. If not, or if you have in fact done this, please let me know and I'll try to work out how to unblock this.

**Update**: actually, I just found the other PR you submitted, and for for that one, the CLA approval bot seemed to think there is no problem. So I'm a bit confused, but in any case, if you could let me know whether a) GitHub is suggestion any action required to you for the PR and b) whether you believe you have agreed to the CLA.

barbaralocsi:
Hi @idg10,

I don't see any required actions. But I am not sure that I am looking at the right place. Or would it be obvious if it's there?

I did agree a CLA when I contributed to microsoft/playwright a few years ago, but I am not sure if that's the same CLA. On https://opensource.microsoft.com/cla/ it says the .NET Foundation CLA is separate. I don't remember accepting other CLAs.

What I know for sure is different between my 2 PRs (dotnet/reactive#2251, dotnet/reactive#2258) is this:
- dotnet/reactive#2251:
  - I edited it directly in github, so the commit is signed and marked as verified.
- dotnet/reactive#2258:
  - I initially commited and pushed with the wrong comitter name and email address. I later fixed it and force pushed on the branch. Maybe the check run with the old commiter details? (Also the commit is not signed)

idg10:
OK...so I've reached out to the .NET Foundation's support team to try and work out what's happening. But it may well be the email address thing. If you don't want to wait for the .NET Foundation support people to get to this, I guess you could try creating a new branch with the same change but the right email, and create another PR.

But I'm happy to wait and see if the .NET Foundation people can fix it if you prefer.

barbaralocsi:
Hi @idg10,
I tried again with https://github.com/dotnet/reactive/pull/2263. That triggered the CLA bot correctly and I could accept it.
Now I see both PRs are unblocked, feel free to close one of them.

barbaralocsi:
Another interesting thing I just realised is that I accepted the Microsoft CLA back in **2022.10.11**.
I did my first PR in this repo on **2025.10.10**, so just before the 3 years passed since accepting the CLA.
While my second PR here was on **2025.10.19**, which is more than 3 years after initially accepting the CLA.

idg10:
So, I just approved your 2nd version of this PR, and then came here to close this one, since they are the same...and now the CLA bot is no longer complaining!

So maybe the .NET foundation folks did something. But in any case, this change is now in via #2263 - thanks, and sorry it took more effort than it should have!

PR #2308, opened 2026-03-17
Prevent trimmer warnings on current SDK

idg10:
This updates the tool that generates code supporting `IQueryable<T>` (the HomoIcon tool) so that the generated code is better aligned with the capabilities of the .NET SDK's IL trimming functionality.

The most obvious change is that this fixes some build warnings reporting trimmability analysis problems. But it has also enabled us to remove the suppressions for various diagnostics that had been added as a stopgap to get basic trimming support working in Rx 6. This may also enable the trimming static analysis to better understand the relevant methods.

This also finally drags the HomoIcon tool into the modernish era: it's now a .NET 10 tool instead of .NET FX, and it is also able to generate nullability annotations. (For several years, nullability had been handled by error-prone manual adjustments to the files this tool generates.)

One upshot of updating this tool is that it has revealed inconsistencies between the `IQbservable<T>` and `IObservable<T>` APIs, most notably the fact that there were a load of overloads available for `CombineLatest` and `Zip` on `IQbservable<T>` that did not have `IObservable<T>` equivalents! (Specifically, the forms that return tuples with more than 8 values.) The main `IObservable<T>` operators have now been extended to align with what the query-space versions have apparently offered for years.


HowardvanRooijen (commented):
Have you created an integration test for running behaviour successfully on a trimmed assembly?

idg10:
@HowardvanRooijen :`
>Have you created an integration test for running behaviour successfully on a trimmed assembly?

I considered this but it was rather beyond the scope of this particular PR: the primary goal was to preserve existing functionality but to remove build warnings.

To test trimming properly I think we'd need to broaden the integration testing. We'd want to include Blazor since that was the original motivation for adding trimming annotations in the first place. Since wasm testing still appears to be a pain, that's not straightforward and might not even be possible yet. (I've not yet seen anything to indicate that the new test platform did anything to improve testability on WASM. I would hope that it did, and if so, migrating to that would be a prerequisite.)

So I've created #2310 to track this.

HowardvanRooijen (approved):

PR #2341, opened 2026-09-21
Add Microsoft.Reactive.Testing.Async library

idg10:
A necessary prerequisite for developing a proper test suite for AsyncRx.NET is to have an async equivalent to the `Microsoft.Reactive.Testing` library we use for Rx.NET. This adds that library.


AsyncRx.NET/Microsoft.Reactive.Testing.Async/AsyncRecorded.cs
```diff
@@ -0,0 +1,54 @@
...
+    public long Start => Time.Start;
+
+    /// <summary>Virtual time at which delivery completed, or <see cref="OperationTime.Infinite"/>.</summary>
+    public long End => Time.End;
+
+    /// <summary>The recorded value.</summary>
+    public T Value { get; }
+
+    public bool Equals(AsyncRecorded<T> other) =>
+        Time == other.Time && Value!.Equals(other.Value);
```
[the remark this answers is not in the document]
idg10: Fixed

AsyncRx.NET/Microsoft.Reactive.Testing.Async/TestAsyncScheduler.cs
```diff
@@ -0,0 +1,540 @@
...
+
+        work.StartedAt = Clock;
+
+        try
+        {
+            var task = work.Action(work.Token);
+
+            if (task.IsCompleted)
+            {
+                ObserveSynchronousCompletion(task, work);
```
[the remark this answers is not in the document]
idg10: Fixed by adding code to detect when we enter an execution context. (Done by setting up an `AsyncLocal` with a callback. This gives us a chance to discover when work is _about_ to run on our context—this means we can check the thread before it has even started to do whatever it was going to do, eliminating the race condition.

AsyncRx.NET/Microsoft.Reactive.Testing.Async/AsyncReactiveAssert.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// This is the compact form of the assertion, which asserts that each notification's delivery
+    /// started and completed at the same tick. If prolonged completion is intended, use the
+    /// extended (start, end) form:
+    /// <see cref="AssertEqual{T}(IEnumerable{AsyncRecorded{Notification{T}}}, AsyncRecorded{Notification{T}}[])"/>.
+    /// </remarks>
+    public static void AssertEqual<T>(
+        this IEnumerable<AsyncRecorded<Notification<T>>> actual,
+        params Recorded<Notification<T>>[] expected)
+    {
+        if (actual == null)
```
HowardvanRooijen: Should we be using 

```csharp
ArgumentNullException.ThrowIfNull(actual);
```

throughout the new library?

idg10: No, because that does not exist in `netstandard2.0`. And the reason this library continues to support `netstandard2.0` is that it is my intention to support the use of AsyncRx.NET on .NET Framework (which means we need to be able to run tests on .NET FX). `netstandard2.0` also happens to enable use on classic UWP, although I do not intend to provide any official support for that... (And targeting `net48` doesn't help AFAIK, because I don't think that offers that helper either.)

HowardvanRooijen: Apologies, I was reviewing on my phone and saw the `<TargetFramework>net10.0</TargetFramework>` on `Tests.Microsoft.Reactive.Testing.Async.csproj` and thought this was a modern library.

AsyncRx.NET/Microsoft.Reactive.Testing.Async/NotificationExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Invokes the observer method corresponding to the notification.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This provides the asynchronous version of <see cref="Notification{T}.Accept(IObserver{T})"/>.
+    /// Since <see cref="Notification{T}"/> is part of System.Reactive, it doesn't know anything
+    /// about asynchronous observers, so we add this capability as an extension method.
+    /// </para>
+    /// <para>
```
mwadams: Is this a future TODO or a "I mean to do resolve this" TODO? (e.g. by opening an issue)

mwadams (approved):
Approved with a comment about TODO capture.

AsyncRx.NET/Microsoft.Reactive.Testing.Async/NotificationExtensions.cs
idg10: We'll catch this in a future sweep for all of the (many) TODO comments throughout the code.

AsyncRx.NET/Microsoft.Reactive.Testing.Async/Microsoft.Reactive.Testing.Async.csproj
```diff
@@ -0,0 +1,31 @@
...
+    CA1512: Use ArgumentOutOfRangeException throw helpers (ThrowIfNegative, ThrowIfGreaterThan, etc.) - not available on netstandard2.0
+    -->
+    <NoWarn>$(NoWarn);CA1510;CA1512</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Reuses the sync testing vocabulary (Recorded, Notification, Subscription, ReactiveTest)
+         so scenarios stay portable between the sync and async platforms. Version matches the
+         System.Reactive package that System.Reactive.Async references. -->
+    <PackageReference Include="Microsoft.Reactive.Testing" Version="6.0.0" />
```
HowardvanRooijen: Shouldn't that be v7.0.0?

idg10: Ah, probably, but I don't think it really matters for two reasons:

1. we haven't changed that package in age; the only reason 7.0.0 exists is because we release new versions of all the Rx packages any time we release a new version of any one of them
2. the long term plan is to move the AsyncRx.NET code inside the Rx.NET folder, at which point that will become a project reference, not a package reference.

I would have updated it if I hadn't already merged the PR...but since it will go away of its own accord eventually, I'm leaving it for now.

